PR #3, opened 2020-06-20
Add remote submission of jobs

ghjm:
This adds a new `work submit` command to the control socket that is similar to `work start` except it takes a nodeID parameter.  The local node will accept stdin from the socket as normal, but instead of running the command, it will open the control socket (which must be named `control`) on the specified node, and use `work start` to run the job.

The local node will then poll for stdout and write it out locally as it becomes available.  Clients subscribed via `work results` will see results in near-real-time.

If the remote node is unreachable at the time the job is submitted (even if it is unknown), the local node will keep retrying until it becomes available.
If the remote node becomes unreachable while a job is running (say, an intermediate node goes down), local subscribers to `work results` will not be closed.  They will just sit there until the remote node comes back up, at which time all the missing stdout will be streamed.

Still to do on this:
* Handle remote `work cancel` and `work release`.
* Specify a TTL so that if the submission can't be sent by a given time, it will be cancelled.
* Decide if every Receptor node should always run a control service named `control`
* Handle TLS parameters for connecting to the remote `control` service

### DIFF
--- a/pkg/workceptor/workceptor.go
+++ b/pkg/workceptor/workceptor.go
@@ -647,7 +647,6 @@ func (w *Workceptor) GetResults(unitID string, startPos int64) (chan []byte, err
 		buf := make([]byte, 1024)
 		for {
 			if stdout == nil {
-				time.Sleep(250 * time.Millisecond)
 				stdout, err = os.Open(stdoutFilename)
 				if err != nil {
 					continue
@@ -684,6 +683,9 @@ func (w *Workceptor) GetResults(unitID string, startPos int64) (chan []byte, err
 					debug.Printf("Stdout complete - closing channel\n")
 					return
 				}
+				// We got to the end of the file but stdout isn't complete yet, so we need to wait
+				// until the running process writes more output.
+				time.Sleep(250 * time.Millisecond)
 				continue
 			} else if err != nil {
 				debug.Printf("Error reading stdout: %s\n", err)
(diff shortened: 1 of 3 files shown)


ghjm:
Sorry @fosterseth, I had to go ahead and implement resumable transfers for this to work at all.

pkg/workceptor/workceptor.go
```diff
@@ -396,12 +667,23 @@ func (w *Workceptor) GetResults(unitID string) (chan []byte, error) {
...
+				}
+				if IsComplete(unit.status.State) && stdoutSize >= unit.status.StdoutSize {
 					close(resultChan)
-					break
+					debug.Printf("Stdout complete - closing channel\n")
+					return
 				}
-				// Don't spin loop the CPU checking and re-checking for EOF
-				time.Sleep(100 * time.Millisecond)
 				continue
```
fosterseth, line 689: should we be able to just return here, since we reached EOF?

ghjm: No, because the file is being actively written to by the other process.  Suppose some task has 1500 bytes of stdout that are eventually going to be written.  At the time we do the check it has written 1000 bytes, and ten seconds in the future, will write another 500.  We get an EOF after reading the first 1000 bytes - we did, in fact, get to the end of the file as it currently exists on disk.  But we can't exit yet because we haven't delivered the last 500 bytes.

Terminating the loop first requires that the process is in a completed state (`WorkStateSuccess` or `WorkStateFailed` - this is what `IsComplete` checks).  In this state we know that the value of `unit.status.StdoutSize` is no longer changing, so once our own count of bytes written reaches that size, we know we have reached the true end of stdout and can close resultsChan.

fosterseth (approved):

PR #23, opened 2020-06-24
Replace x/net websocket with gorilla websocket

fosterseth:
golang.org/x/net is missing features or is deprecated, so replacing with a more active library, github.com/gorilla/websocket

issue https://github.com/ghjm/sockceptor/issues/5

### DIFF
--- a/pkg/backends/websockets.go
+++ b/pkg/backends/websockets.go
@@ -3,20 +3,15 @@ package backends
 import (
 	"crypto/tls"
 	"fmt"
+	"github.com/gorilla/websocket"
 	"github.com/project-receptor/receptor/pkg/cmdline"
 	"github.com/project-receptor/receptor/pkg/debug"
-	"github.com/project-receptor/receptor/pkg/framer"
 	"github.com/project-receptor/receptor/pkg/netceptor"
-	"golang.org/x/net/websocket"
 	"net/http"
 	"net/url"
 	"strings"
 	"time"
 )
-
-//TODO: Switch to better supported websockets library
-//TODO: TLS Server
-//TODO: HTTP/HTTPS proxy support
 
 // WebsocketDialer implements Backend for outbound Websocket
 type WebsocketDialer struct {
@@ -33,9 +28,13 @@ func NewWebsocketDialer(address string, tlscfg *tls.Config, extraHeader string, 
 	if err != nil {
 		return nil, err
 	}
+	httpScheme := "http"
+	if addrURL.Scheme == "wss" {
+		httpScheme = "https"
+	}
 	wd := WebsocketDialer{
 		address:     address,
-		origin:      fmt.Sprintf("http://%s", addrURL.Host),
+		origin:      fmt.Sprintf("%s://%s", httpScheme, addrURL.Host),
 		redial:      redial,
 		tlscfg:      tlscfg,
 		extraHeader: extraHeader,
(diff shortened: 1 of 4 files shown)


pkg/backends/websockets.go
```diff
@@ -29,13 +29,9 @@ type WebsocketDialer struct {
...
 // NewWebsocketDialer instantiates a new WebsocketDialer backend
 func NewWebsocketDialer(address string, tlscfg *tls.Config, extraHeader string, redial bool) (*WebsocketDialer, error) {
-	addrURL, err := url.Parse(address)
-	if err != nil {
-		return nil, err
-	}
 	wd := WebsocketDialer{
 		address:     address,
-		origin:      fmt.Sprintf("http://%s", addrURL.Host),
+		origin:      fmt.Sprintf("http://%s", address),
```
ghjm: I think the `address` parameter supplied to this function is expected to be a URL.  So you probably still have to parse out the host part to make an origin.  Maybe the scheme also needs to be http or https based on whether `address`'s URL scheme is `ws` or `wss`.

pkg/backends/websockets.go
```diff
@@ -47,21 +43,19 @@ func NewWebsocketDialer(address string, tlscfg *tls.Config, extraHeader string,
...
-			if err != nil {
-				errf(fmt.Errorf("error creating websocket config"), true)
-				return
-			}
+			dialer := websocket.DefaultDialer
 			if b.tlscfg != nil {
-				cfg.TlsConfig = b.tlscfg
+				dialer.TLSClientConfig  = b.tlscfg
 			}
+			header := make(http.Header, 0)
```
ghjm, line 54: This declaration is going to be shadowed by the `:=` a few lines below, inside the `if {}` block.  This is a go-ism - every curly brace block implies a lexical scope, so the second `:=` declares a new variable that only exists within the `if {}`.  I think you get the behavior you want by just deleting the second declaration (line 60), which will make the `header.Add` apply to the outer variable.

pkg/backends/websockets.go
```diff
@@ -47,21 +43,19 @@ func NewWebsocketDialer(address string, tlscfg *tls.Config, extraHeader string,
...
 			if b.extraHeader != "" {
 				extraHeaderParts := strings.SplitN(b.extraHeader, ":", 2)
 				header := make(http.Header, 0)
 				header.Add(http.CanonicalHeaderKey(extraHeaderParts[0]), extraHeaderParts[1])
-				cfg.Header = header
 			}
-			conn, err := websocket.DialConfig(cfg)
+			header.Add(http.CanonicalHeaderKey("origin"), b.origin)
+			u := url.URL{Scheme: "ws", Host: b.address, Path: "/"}
```
ghjm: Callers need to be able to specify `ws://` or `wss://` - we can't assume one or the other here.  That's why `address` needs to be a full URL.

pkg/backends/websockets.go
```diff
@@ -136,28 +132,25 @@ func newWebsocketSession(conn *websocket.Conn) *WebsocketSession {
 // Send sends data over the session
 func (ns *WebsocketSession) Send(data []byte) error {
 	buf := ns.framer.SendData(data)
-	n, err := ns.conn.Write(buf)
+	err := ns.conn.WriteMessage(websocket.BinaryMessage, buf)
```
ghjm: Does `WriteMessage()` take a whole message, which will then be presented in its entirety on the other side via `ReadMessage()`?  If so then we don't actually need `framer` at all here.

ghjm (changes requested):

pkg/backends/websockets.go
fosterseth: fixed

pkg/backends/websockets.go
fosterseth: yes it takes whole message, so I removed framer.

ghjm:
As part of this PR, we should remove the TODOs at the top of `pkg/backends/websockets.go`

ghjm:
I was thinking that in the case where the user does `receptor --node id=foo --ws-peer address=wss://bar.com` it will fail because when it gets to `NewWebsocketDialer`, tlscfg will be nil, so it will be trying to run an unencrypted connection to a `wss://` endpoint.  It would be more user friendly to use default TLS settings in this case.

We have a client TLS configuration named `default` that is always created, and is intended for situations like this.  What I'm thinking is we check in `WebSocketDialerCfg.Run` if `cfg.TLS==""` and the scheme is `wss://`, then when we call `GetClientTLSConfig`, we pass in `default` as the config name instead of `cfg.TLS`.

pkg/backends/websockets.go
```diff
@@ -3,19 +3,17 @@ package backends
...
 	"net/http"
 	"net/url"
 	"strings"
 	"time"
 )
 
-//TODO: Switch to better supported websockets library
-//TODO: TLS Server
-//TODO: HTTP/HTTPS proxy support
+//TODO:
```
ghjm: No reason to leave this here

fosterseth: fixed

ghjm (approved):

PR #26, opened 2020-06-26
Send nonblocking signals to shutdown netceptor instances

squidboylan:


squidboylan:
hmm, this may actually not fix shutdown, but might just prevent the hanging

ghjm:
Yeah, I'd like to figure out who's not listening for this and why.  But I agree with the idea of protecting the main thread.  Maybe it would be better to do this by making the shutdown channels buffered with size 1?  That will still avoid blocking on the main thread, but if the goroutine eventually finishes what it's doing and listens to the done channel, it will still get the shutdown request.

squidboylan:
> Yeah, I'd like to figure out who's not listening for this and why. But I agree with the idea of protecting the main thread. Maybe it would be better to do this by making the shutdown channels buffered with size 1? That will still avoid blocking on the main thread, but if the goroutine eventually finishes what it's doing and listens to the done channel, it will still get the shutdown request.

Ok i think i figured out what's going on, the signal in `runProtocol` shuts down that tcp session, but doesnt close the tcp port because we dont have a way to tell it to in https://github.com/project-receptor/receptor/blob/devel/pkg/backends/tcp.go#L84 so i think we need a way to send a signal to close the tcp port rather than just closing a session, however im not really sure how to architect that

ghjm:
Ok, can you open an issue with the details and I'll take a look at it?

Also I think the shutdown channels are not being properly initialized.

squidboylan:
Closing this as this PR is likely a deadend, https://github.com/project-receptor/receptor/issues/28 will track future work to fix this

PR #33, opened 2020-06-30
Move CI to Docker

ghjm:
This changes our CI to use Docker actions instead of running things in the default Ubuntu runner.  In addition to running the pre-commit steps, this now also verifies that the software can build with EL8's golang.  In the future this will make it easier for us to run things like `go test` as CI steps.


ghjm:
Switching this to draft because I want to find a way to not rebuild the container image on every run before merging this

fosterseth (approved):

PR #34, opened 2020-07-01
Move CI tasks to Makefile target so you can easily run them locally

ghjm:
This moves the CI tests to a Makefile target for ease of running the same test suite locally.  I meant to do this in https://github.com/project-receptor/receptor/pull/33 but forgot.


squidboylan (approved):
LGTM

PR #36, opened 2020-07-01
Expose knownConnectionCosts in Netceptor.Status()

squidboylan:


pkg/netceptor/netceptor.go
```diff
@@ -258,10 +259,11 @@ func (s *Netceptor) Status() Status {
...
 	return Status{
-		NodeID:         s.nodeID,
-		Connections:    conns,
-		RoutingTable:   routes,
-		Advertisements: serviceAds,
+		NodeID:               s.nodeID,
+		Connections:          conns,
+		RoutingTable:         routes,
+		Advertisements:       serviceAds,
+		KnownConnectionCosts: s.knownConnectionCosts,
```
ghjm: You should be holding a read lock on `s.knownNodeLock` before accessing this.

squidboylan: yeah im realizing this is probably all sorts of wrong, i'll fix it shortly. I also need to deep copy that map of maps, this is just copying a pointer i think

squidboylan:
when i run this it sometimes crashes due to reading and writing at the same time i think, i wonder if something is running `printRoutingTable` without holding the lock

ghjm (approved):

ghjm:
`printRoutingTable` only gets called from one place, and it does hold the lock there.  I don't see anywhere else where it's being used without a lock, or being written to with only a read lock.

PR #37, opened 2020-07-01
Add automerge github action

ghjm:
This allows us to set a label of "mergeit" which will auto-merge the PR once approvals and checks are all in place.

I am using rebase rather than squash because (a) I dislike merge commits, (b) I like preserving history, and (c) I don't like the auto generated commit messages that would result from auto-squashing.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -1,0 +1,32 @@
+name: automerge
+on:
+  pull_request:
+    types:
+      - labeled
+      - unlabeled
+      - synchronize
+      - opened
+      - edited
+      - ready_for_review
+      - reopened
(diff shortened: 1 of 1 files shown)


squidboylan (commented):
Seems fine to me, im not entirely sure why we need this rather than just pushing the merge button, unless this reruns tests and only merges if they pass again. What am I missing?

squidboylan (approved):
NVM im realizing this is so you can approve things before tests finish and it'll automatically merge after tests pass

ghjm:
Yes, that's the main reason I want this.  It gives you a way of saying "I am actually ready to merge this" when you are waiting for test results and/or reviews.

PR #42, opened 2020-07-06
Add golang functional tests

squidboylan:


Makefile
```diff
@@ -15,6 +15,7 @@ ci:
 	echo "Running Go builds" && \
 	go build cmd/*.go && \
 	go build example/*.go && \
+	go test ./... -count=1 && \
```
ghjm: Nicer Makefile that lets you run the tests separately if you want:
```
pre-commit:
        echo "Running pre-commit" && \
        pre-commit run --all-files

build-all:
        echo "Running Go builds" && \
        go build cmd/*.go && \
        go build example/*.go

test:
        go test ./... -count=1

ci: pre-commit build-all test
        echo "All done"

.PHONY: lint format fmt ci pre-commit build-all test
```

squidboylan: yeah i was thinking about this, it's also kinda nice to have tests run even if go fmt and go lint fail so i'll probably rework this a bit

tests/functional/mesh/mesh.go
```diff
@@ -0,0 +1,291 @@
...
+			}
+		}
+		Nodes[MeshDefinition.Nodes[k].Name] = node
+	}
+	for k := range MeshDefinition.Nodes {
+		node := Nodes[MeshDefinition.Nodes[k].Name]
+		for conn, cost := range MeshDefinition.Nodes[k].Connections {
+			// Update this to choose which listener to dial into
+			if MeshDefinition.Nodes[conn].Listen[0].Protocol == "tcp" {
+				node.TCPDial(MeshDefinition.Nodes[conn].Listen[0].Addr, cost)
```
ghjm: Error return should be handled or explicitly ignored

tests/functional/mesh/mesh.go
```diff
@@ -0,0 +1,291 @@
...
+			} else if MeshDefinition.Nodes[conn].Listen[0].Protocol == "udp" {
+				node.UDPDial(MeshDefinition.Nodes[conn].Listen[0].Addr, cost)
```
ghjm: Same here

tests/functional/mesh/mesh_test.go
```diff
@@ -0,0 +1,140 @@
...
+				t.Error(err)
+			}
+			yamlDat, err := ioutil.ReadFile(filename)
+			if err != nil {
+				t.Error(err)
+			}
+
+			data := YamlData{}
+
+			yaml.Unmarshal(yamlDat, &data)
```
ghjm: Another unchecked error return

tests/functional/mesh/mesh_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+import (
+	"gopkg.in/yaml.v2"
+	"io/ioutil"
+	"testing"
+	"time"
+)
+
+// Test that a mesh starts and that connections are what we expect and that
+// each node's view of the mesh converges
```
ghjm, line 11: I don't think this comment is accurate ... we seem to be only testing that the mesh starts up successfully, _not_ that the nodes' routing tables converge.

Edit: Oh, looks like WaitForReady actually does verify convergence of connections and known costs.

tests/functional/mesh/mesh.go
```diff
@@ -0,0 +1,291 @@
...
+	return NewMeshFromYaml(&MeshDefinition)
+}
+
+// NewMeshFromYaml takes a yaml mesh description and returns a mesh of nodes
+// listening and dialing as defined in the yaml
+// TODO: Add support for websockets
+func NewMeshFromYaml(MeshDefinition *YamlData) (*Mesh, error) {
+	Nodes := make(map[string]Node)
+
+	// We must start listening on all our nodes before we start dialing so
```
ghjm, line 134: Dialing with `redial=true` should connect even if the listener doesn't start till after the dialer ... though I agree it's desirable to avoid the retry delay for speed of startup.  Maybe a future test could intentionally start the dialer first and verify that it connects after the listener comes up.

tests/functional/mesh/mesh.go
```diff
@@ -0,0 +1,291 @@
...
+			if nodeID == status.NodeID {
+				continue
+			}
+			for k, v := range node.Connections {
+				if k == status.NodeID {
+					expectedConnections[nodeID] = v
+				}
+			}
+		}
+		if reflect.DeepEqual(actualConnections, expectedConnections) {
```
ghjm, line 234: Nice!

tests/functional/mesh/mesh.go
```diff
@@ -0,0 +1,291 @@
...
+	if routesReady == false {
+		return errors.New("Timed out while waiting for routes to settle")
+	}
+
+	return nil
+}
+
+// Status returns a list of statuses from the contained netceptors
+func (m *Mesh) Status() []netceptor.Status {
+	out := []netceptor.Status{}
```
ghjm:
```suggestion
	var out []netceptor.Status
```
avoids an unnecessary allocation

tests/functional/utils/utils.go
```diff
@@ -0,0 +1,38 @@
+package utils
+
+import (
+	"net"
+)
+
+// RandomTCPPort returns an unused TCP port
+// This has a race condition where something else could claim the TCP port
+// before we do
+func RandomTCPPort() (int, error) {
```
ghjm: We could pretty easily add the ability for the back-ends to listen on an ephemeral port and then provide some way to get the port number back, which would eliminate this race condition.

tests/functional/mesh/mesh.go
squidboylan: right, the real issue is we dont know the addresses to dial into because we generate those randomly

tests/functional/utils/utils.go
squidboylan: Yeah this would be my preferred solution

ghjm (approved):

PR #43, opened 2020-07-06
Add support for retrieving local address when running on an ephemeral port

ghjm:
If you don't pass a port number as part of the address, the `net.Listener` and `net.ListenUDP` will listen on an ephemeral port.  This PR provides the ability for a caller to retrieve the resulting address, including port number.

The `net/http` library does not, as far as I can tell, offer the ability to listen on an ephemeral port.  If no port is specified, it listens on port 80.


squidboylan (approved):
Awesome! Looks good to me, I can rework my tests PR to use this and if there are issues we can revist it

PR #45, opened 2020-07-07
Test that every node has a route to every other node

squidboylan:


tests/functional/mesh/mesh.go
```diff
@@ -132,6 +133,42 @@ func (n *Node) WebsocketDial(address string, cost float64) error {
...
+	startTime := time.Now()
+	replyChan := make(chan net.Addr)
+	go func() {
+		buf := make([]byte, 8)
+		_, addr, err := pc.ReadFrom(buf)
+		if err == nil {
+			replyChan <- addr
+		}
+	}()
+	_, err = pc.WriteTo([]byte{}, netceptor.NewAddr(node, "ping"))
```
ghjm: We might want to call WriteTo in a loop here, because PacketConn packets aren't guaranteed to arrive

tests/functional/mesh/mesh.go
```diff
@@ -313,13 +366,22 @@ func (m *Mesh) WaitForReady(timeout float64) error {
 		return errors.New("Timed out while waiting for connections")
 	}
 
-	routesReady := m.CheckKnownConnectionCosts()
-	for ; timeout > 0 && !routesReady; routesReady = m.CheckKnownConnectionCosts() {
+	costsConsistent := m.CheckKnownConnectionCosts()
+	for ; timeout > 0 && !costsConsistent; costsConsistent = m.CheckKnownConnectionCosts() {
+		time.Sleep(100 * time.Millisecond)
+		timeout -= 100
```
ghjm, line 392: https://golang.org/pkg/context/#WithTimeout might be useful here

ghjm (approved):

PR #47, opened 2020-07-07
Use a context to manage shutting down our backends

squidboylan:


pkg/backends/tcp.go
```diff
@@ -82,9 +91,10 @@ func (b *TCPListener) Addr() net.Addr {
 }
 
 // Start runs the given session function over the WebsocketListener backend
-func (b *TCPListener) Start(bsf netceptor.BackendSessFunc, errf netceptor.ErrorFunc) {
+func (b *TCPListener) Start(shutdownContext context.Context, wg sync.WaitGroup, bsf netceptor.BackendSessFunc, errf netceptor.ErrorFunc) {
 	var err error
-	b.li, err = net.Listen("tcp", b.address)
+	var listenerCfg net.ListenConfig
+	b.li, err = listenerCfg.Listen(shutdownContext, "tcp", b.address)
```
squidboylan, line 97: I'm not sure we actually need this, the docs on what this context actually does are lacking https://golang.org/pkg/net/#ListenConfig

squidboylan:
I'm not sure if more needs to be done to kill services listening on the receptor network or if they will stop as the connection to the rest of the mesh closes

ghjm:
I haven't looked at this in much detail yet, but my initial reaction is that I'm not keen on complexifying the implementation of a backend like this.  I feel like the waitgroup and context shouldn't be something a backend implementor needs to know about.  The framework should handle everything it's possible to abstract out of the individual backends.

I don't like the way the framework calls the backend's Start which then calls the framework's session function.  This is left over from a previous version of this interface.  I think this poor design of is making this task harder than it needs to be.

squidboylan:
I forgot to update this PR but we talked and decided this was not the route forward

Issue #10, opened 2020-06-20
Add logging and clean up debug output


ghjm:
Right now there are a lot of calls to `debug.Printf` throughout the application.  These should be changed to use the Go logger and some thought should be put in to how much, if any, output should be produced when running _not_ in debug mode.

PR #48, opened 2020-07-08
Add logger

fosterseth:
related #10 

`debug.Printf(...`

becomes

`logger.Debug(...`

Currently there are the following levels, `Error`, `Warning`, `Info`, and `Debug`

We can control the verbosity level in the yaml or the command line with

`--log-level Debug` This will output all statements at the `Debug` level and below (`Info`, `Warning`, `Error`)

`Trace` is not a log level, and must be enabled with the `--trace` param

example output:
```
DEBUG 2020/07/09 14:40:58 Running Websocket peer connection ws://localhost:2222
INFO 2020/07/09 14:40:59 Initialization complete
DEBUG 2020/07/09 14:41:08 Sending initial connection message
INFO 2020/07/09 14:41:08 Connection established with foo
DEBUG 2020/07/09 14:41:08 Stopping initial updates

```

Note: Errors / messaged related to command line parsing do not go through the logger, and still use `fmt`


ghjm:
Let's make the default verbosity be 0.  We always use --debug ourselves, but that's because we're developers.  End users shouldn't see debug messages unless they ask for them.

Should we use `logger.Warning()` or above for things that are real errors?  There's a lot of stuff in the current code that calls `debug.Printf` but is really one or another kind of actual error, not a debug message.

Also, if you add `barevalue:"yes"` to the struct tags of verbosityCfg.Level, then you'll be able to use `--verbosity 1` instead of having to spell out `level=`.  And it should probably also have `required:"yes"`.

squidboylan (approved):
LGTM once tests are passing

ghjm:
@fosterseth Looks good.  You've got linter errors around the formatting of comments on exported functions.  The reason the linter cares about this comment format is to make sure they are parseable by godoc.

On the `--verbosity` vs `-v` question, cmdline should probably gain the ability to have short options.  Also, it does support bash completion, so if you're using that you can just type `--v<tab>` and not have to type out `--verbosity`.

What would you think of calling it `--log-level` instead of `--verbosity`?

fosterseth:
@ghjm switching from `--verbosity` to `--log-level` as I feel it's more clear

I also wrote some helper functions to better encapsulate the `logLevel` and `showTrace` variables.

ghjm (approved):

ghjm:
This looks good, although maybe we should consider putting the command like config object in logger.go instead of receptor.go?

PR #51, opened 2020-07-13
Rework backend scheme

ghjm:
This PR makes the following changes:

- Instead of `RunBackend` calling `backend.Start()` which then calls back a `BackendSessionFunc`, backends register with a simplified `AddBackend` call, allowing more logic to be abstracted in the framework rather than requiring implementation in each backend.
- Utility functions are added for the common use cases of a Dial/redial and Listen/Accept backend types.
- Netceptor is restructured to use Go contexts instead of the shutdown channels slice, allowing for cleaner and more reliable shutdown, as well as being more idiomatic to Go.
- The logger is updated to treat log levels case insensitively, so users can run `--log-level debug` (lower case D) without error.

See https://github.com/project-receptor/receptor/issues/28

### DIFF
--- a/example/net.go
+++ b/example/net.go
@@ -16,13 +16,9 @@ import (
 */
 
 func main() {
-	// Create contexts so we can shut down the nodes
-	ctx1, _ := context.WithCancel(context.Background())
-	ctx2, _ := context.WithCancel(context.Background())
-
 	// Create two nodes of the Receptor network-layer protocol (Netceptors).
-	n1 := netceptor.New(ctx1, "node1", nil)
-	n2 := netceptor.New(ctx2, "node2", nil)
+	n1 := netceptor.New(context.Background(), "node1", nil)
+	n2 := netceptor.New(context.Background(), "node2", nil)
 
 	// Start a TCP listener on the first node
 	b1, err := backends.NewTCPListener("localhost:3333", nil)
--- a/example/net.go
+++ b/example/net.go
@@ -1,13 +1,13 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"github.com/project-receptor/receptor/pkg/backends"
 	"github.com/project-receptor/receptor/pkg/netceptor"
 	"io"
 	"net"
 	"os"
-	"sync"
 	"time"
 )
 
@@ -15,18 +15,14 @@ import (
    This is an example of the use of Receptor as a Go library.
 */
 
-// Error handler that gets called for backend errors
-func handleError(err error, fatal bool) {
-	fmt.Printf("Error: %s\n", err)
-	if fatal {
-		os.Exit(1)
-	}
-}
-
 func main() {
+	// Create contexts so we can shut down the nodes
+	ctx1, _ := context.WithCancel(context.Background())
+	ctx2, _ := context.WithCancel(context.Background())
+
 	// Create two nodes of the Receptor network-layer protocol (Netceptors).
-	n1 := netceptor.New("node1", nil)
-	n2 := netceptor.New("node2", nil)
+	n1 := netceptor.New(ctx1, "node1", nil)
+	n2 := netceptor.New(ctx2, "node2", nil)
 
 	// Start a TCP listener on the first node
 	b1, err := backends.NewTCPListener("localhost:3333", nil)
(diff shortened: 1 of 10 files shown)


pkg/netceptor/netceptor.go
```diff
@@ -607,23 +630,25 @@ func (s *Netceptor) getEphemeralService() string {
...
 		return
 	}
 	logger.Debug("Known Connections:\n")
 	for conn := range s.knownConnectionCosts {
-		logger.Debug("   %s: ", conn)
+		sb := &strings.Builder{}
+		_, _ = fmt.Fprintf(sb, "   %s: ", conn)
 		for peer := range s.knownConnectionCosts[conn] {
-			logger.Debug("%s(%.2f) ", peer, s.knownConnectionCosts[conn][peer])
+			_, _ = fmt.Fprintf(sb, "%s(%.2f) ", peer, s.knownConnectionCosts[conn][peer])
```
squidboylan, line 637: im not sure i understand why this isnt a `logger.Info` output

example/net.go
```diff
@@ -1,48 +1,48 @@
...
-func handleError(err error, fatal bool) {
-	fmt.Printf("Error: %s\n", err)
-	if fatal {
-		os.Exit(1)
-	}
-}
-
 func main() {
+	// Create contexts so we can shut down the nodes
+	ctx1, _ := context.WithCancel(context.Background())
```
squidboylan: I think this can be `context.Background()` since the netceptor adds the cancel function and we're not using this cancel function anyways. I think we could also pass a nil value and have the netceptor do everything.

fosterseth (approved):

pkg/netceptor/netceptor.go
```diff
@@ -197,38 +172,86 @@ func New(NodeID string, AllowedPeers []string) *Netceptor {
...
+			case <-s.context.Done():
+				return
+			}
+		}
+	}()
+	return nil
+}
+
+// DoneBackend signals to the wait group that the backend is done
+func (s *Netceptor) DoneBackend() {
```
squidboylan: Is this really something we want to export? Im not sure I see the value of that, it seems scary

squidboylan (approved):
Seems reasonable to me, i left some inline comments

pkg/netceptor/netceptor.go
ghjm: I think of this as a normal and routine operation that probably doesn't need to spam the logs, but I could be pretty easily talked into this being Info level.  Particularly since it only gets printed when something changes.

example/net.go
ghjm: Yep, you're right

pkg/netceptor/netceptor.go
ghjm: Yep, this is a holdover from the old code and shouldn't exist any more

pkg/netceptor/netceptor.go
squidboylan: hmm, yeah thinking about it more info is probably too high, but debug seems maybe appropriate. I was mostly just making sure this was intentional. I kinda expected we would want all output to go through the logger unless it cant because the logger isnt setup yet.

squidboylan: In any case, we dont need to block on this, we can always change this later without much (any?) pain

ghjm: Oh, I see what you're saying.  All the output _is_ going through the logger.  The problem is that the logger adds a newline if one isn't already present, so you can't make multiple logger calls to build a single line the way I was originally doing with the printfs.  Instead, I'm constructing a whole line in the loop before sending it to the logger.  The `fmt.Fprintf` calls are writing _to the `string.Builder`_, not to the console.

ghjm:
@fosterseth I made some more changes to the logger, please have another look

pkg/netceptor/netceptor.go
squidboylan: Ah! ok, yeah sorry i was confused

ghjm:
> @fosterseth I made some more changes to the logger, please have another look

Actually I have to go so I'm going to merge this, but let me know if you have any objections to the logger changes.

PR #55, opened 2020-07-14
Test ssl backends

squidboylan:


fosterseth (approved):

ghjm:
In Go it's remarkably easy to generate X.509 certificates on the fly.  Should we maybe do that instead of putting hardcoded certs in the tests?

ghjm:
See https://github.com/project-receptor/receptor/blob/9cc435466f2d29fc3e5d79714b7c5fd6f5a607aa/pkg/netceptor/conn.go#L210 for an example

ghjm (approved):

PR #56, opened 2020-07-15
Receptorctl

ghjm:
Add a new Python module and CLI called receptorctl, that functions as a front-end to the Receptor control socket.  Currently this implements the three built-in control service commands: status, ping and connect.

### DIFF
--- a/pkg/controlsvc/controlsvc.go
+++ b/pkg/controlsvc/controlsvc.go
@@ -139,7 +139,7 @@ func (s *Server) controlPing(params string, cfo ControlFuncOperations) (map[stri
 	cfr := make(map[string]interface{})
 	select {
 	case addr := <-replyChan:
-		cfr["Result"] = fmt.Sprintf("Reply from %s in %s\n", addr.String(), time.Since(startTime))
+		cfr["Result"] = fmt.Sprintf("Reply from %s in %s", addr.String(), time.Since(startTime))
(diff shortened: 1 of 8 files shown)


ghjm:
Quick example of how this works as a Python library:
```
Python 3.7.7 (default, Mar 11 2020, 21:18:58)
[GCC 10.0.1 20200216 (Red Hat 10.0.1-0.8)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import receptorctl
>>> x = receptorctl.ReceptorControl()
>>> x.connect('unix:~/receptor.sock')
>>> x.simple_command('status')
{'Advertisements': [{'NodeID': 'foo', 'Service': 'control', 'Time': '2020-07-14T22:44:21.654845247-04:00', 'Tags': None}], 'Connections': [], 'KnownConnectionCosts': {}, 'NodeID': 'foo', 'RoutingTable': {}}
>>> x.simple_command('status')['NodeID']
'foo'
>>> x.ping('foo')
(True, 'Reply from foo:ping in 38.612µs')
>>> x.ping('bad')
(False, RuntimeError('no route to node'))
>>> x.print_status()
Node ID: foo

Node         Service   Last Seen            Tags
foo          control   2020-07-14 22:42:42  None
>>>
```

fosterseth (approved):

PR #58, opened 2020-07-15
Test shutdown

squidboylan:


tests/functional/mesh/mesh_test.go
```diff
@@ -94,6 +99,56 @@ func TestMeshConnections(t *testing.T) {
...
+			mesh.Shutdown()
+			mesh.WaitForShutdown()
+
+			// Check that the connections are closed
+			pid := os.Getpid()
+			pidString := strconv.Itoa(pid)
+			done := false
+			for timeout := 1000000; timeout > 0 && !done; {
+				var out bytes.Buffer
+				cmd := exec.Command("netstat", "-tunp")
```
squidboylan: This is pretty hacky and im not a huge fan but i think it's fine

pkg/backends/tcp.go
```diff
@@ -113,6 +115,7 @@ func newTCPSession(conn net.Conn, closeChan chan struct{}) *TCPSession {
 		conn:      conn,
 		framer:    framer.New(),
 		closeChan: closeChan,
+		once:      sync.Once{},
```
ghjm: Let's give it a more meaningful name like `closeChanCloser` or something ... calling it `once` gives no hint at all to future programmer/archaeologists digging through this code what its purpose might be.

pkg/backends/tcp.go
```diff
@@ -152,10 +155,12 @@ func (ns *TCPSession) Recv() ([]byte, error) {
 
 // Close closes the session
 func (ns *TCPSession) Close() error {
-	if ns.closeChan != nil {
-		close(ns.closeChan)
-		ns.closeChan = nil
-	}
+	ns.once.Do(func() {
+		if ns.closeChan != nil {
```
ghjm: I feel like this logic should be inverted - we should check for nil-ness first, then wrap only the actual `close` call in the `sync.Once`.

tests/functional/mesh/mesh_test.go
```diff
@@ -94,6 +99,56 @@ func TestMeshConnections(t *testing.T) {
...
+				t.Error(err)
+			}
+			mesh.Shutdown()
+			mesh.WaitForShutdown()
+
+			// Check that the connections are closed
+			pid := os.Getpid()
+			pidString := strconv.Itoa(pid)
+			done := false
+			for timeout := 1000000; timeout > 0 && !done; {
```
ghjm: It's more idiomatic to use the duration constants rather than relying on durations being nanoseconds (and making people count the zeros).  `timeout := 1 * time.Second` would be make it much more immediately apparent what's going on here.

tests/functional/mesh/mesh_test.go
```diff
@@ -94,6 +99,56 @@ func TestMeshConnections(t *testing.T) {
...
+				err := cmd.Run()
+				if err != nil {
+					t.Fatal(err)
+				}
+				if !strings.Contains(out.String(), pidString) {
+					done = true
+					break
+				}
+				time.Sleep(100 * time.Millisecond)
+				timeout -= 100
```
ghjm: You're sleeping for 100 milliseconds but then only subtracting 100 nanoseconds.  Did you intend to do `timeout -= 100 * time.Millisecond`?

tests/functional/mesh/mesh_test.go
ghjm: I'm not keen on shelling out to an external command here either, but if we're going to do that, it should probably be `ss` from the iproute package.  The net-tools package has been considered deprecated for several years and it might be problematic to rely on it existing in CI environments etc.

ghjm (changes requested):

squidboylan:
OK i updated this to use `ss`, looks like TCP connections arent being properly closed (which wasnt being seen before because i forgot the `-l` on the `netstat` command), I'm gonna investigate that, i could have sworn this was working earlier in the week

ghjm (approved):

Issue #54, opened 2020-07-14
Remote cancel and release in Workceptor


ghjm:
Right now the Workceptor `work cancel` and `work release` commands are not aware of remote jobs.  Cancelling or releasing a job created by `work submit` will only cancel or release the local monitor associated with the job - it will keep running, and its files will continue to exist, on the remote node.  The `cancel` and `release` commands should be extended to operate on the remote node.  (One issue with this is defining what the behavior should be if the remote node is unreachable.)

squidboylan:
@fosterseth added coverage for this to our CI that tests cancelling and verifies that when you `work release` on a remote job it deletes the files on the local and remote nodes

PR #59, opened 2020-07-15
Cancel and release extend to remote units of work

fosterseth:
issue #54

`cancelRemote()` uses a goroutine that connects to the remote node and issues a `work cancel` command. This goroutine will not timeout, but will terminate after a successful cancel or a release.

`ReleaseUnit()` will make a one-time attempt to connect to remote and issue a `work release` command. Regardless if it is successful, it will go ahead and delete local files.

Cancel and Release are robust to when nodes are down. If either the control node or remote node go down before the cancellation takes place, it will cancel the work when the nodes come back online.

### DIFF
--- a/pkg/workceptor/workceptor.go
+++ b/pkg/workceptor/workceptor.go
@@ -708,16 +708,22 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
 }
 
 func (w *Workceptor) cancelRemote(remoteNodeID, remoteUnitID string, unit *workUnit, firstAttemptSuccess chan bool) {
+	// to prevent blocking on sending to channel firstAttemptSuccess
+	firstAttemptOnce := sync.Once{}
+	reportFirstResult := func(success bool) {
+		firstAttemptOnce.Do(func() {
+			if firstAttemptSuccess != nil {
+				firstAttemptSuccess <- success
+			}
+		})
+	}
 	retryInterval := SuccessWorkSleep
 	firstTime := true
 	for {
 		if firstTime {
 			firstTime = false
 		} else {
-			select {
-			case firstAttemptSuccess <- false:
-			default:
-			}
+			reportFirstResult(false)
 			time.Sleep(retryInterval)
 			retryInterval = time.Duration(1.5 * float64(retryInterval))
 			if retryInterval > MaxWorkSleep {
@@ -729,10 +735,7 @@ func (w *Workceptor) cancelRemote(remoteNodeID, remoteUnitID string, unit *workU
 		released := unit.released
 		unit.lock.RUnlock()
 		if released {
-			select {
-			case firstAttemptSuccess <- true:
-			default:
-			}
+			reportFirstResult(true)
 			return
 		}
 		conn, reader, err := w.connectToRemote(remoteNodeID)
@@ -746,16 +749,17 @@ func (w *Workceptor) cancelRemote(remoteNodeID, remoteUnitID string, unit *workU
 			conn.Close()
 			continue
 		}
-
 		response, err := reader.ReadString('\n')
+		if err != nil {
+			logger.Error("Read error reading from %s: %s\n", remoteNodeID, err)
+			conn.Close()
+			continue
+		}
 		if response[:5] == "ERROR" {
 			conn.Close()
 			continue
 		} else {
-			select {
-			case firstAttemptSuccess <- true:
-			default:
-			}
+			reportFirstResult(true)
 			conn.Close()
 			return
 		}
@@ -829,14 +833,21 @@ func (w *Workceptor) ReleaseUnit(unitID string) error {
 	unit.waitRemote.Wait()
 
 	if remoteUnitID != "" {
-		conn, _, err := w.connectToRemote(nodeID)
+		conn, reader, err := w.connectToRemote(nodeID)
 		if err != nil {
 			logger.Error("Connection failed to %s: %s\n", nodeID, err)
-		}
-		if conn != nil {
+		} else {
 			_, err = conn.Write([]byte(fmt.Sprintf("work release %s\n", remoteUnitID)))
 			if err != nil {
 				logger.Error("Write error sending to %s: %s\n", nodeID, err)
+			} else {
+				response, err := reader.ReadString('\n')
+				if err != nil {
+					logger.Error("Read error reading from %s: %s\n", nodeID, err)
+				}
+				if response[:5] == "ERROR" {
+					logger.Error("Failed to release %s on remote node %s", remoteUnitID, nodeID)
+				}
 			}
 			conn.Close()
 		}
--- a/pkg/workceptor/workceptor.go
+++ b/pkg/workceptor/workceptor.go
@@ -71,12 +71,12 @@ type workType struct {
 
 // Internal data for a single unit of work
 type workUnit struct {
-	lock             *sync.RWMutex
-	started          bool
-	released         bool
-	monitoringRemote bool
-	worker           WorkType
-	status           *StatusInfo
+	lock         *sync.RWMutex
+	started      bool
+	released     bool
+	waitRemote   *sync.WaitGroup
+	worker       WorkType
+	status       *StatusInfo
 }
 
 // Workceptor is the main object that handles unit-of-work management
@@ -428,9 +428,7 @@ func (w *Workceptor) monitorRemoteStdout(ctx context.Context, cancel context.Can
 func (w *Workceptor) monitorRemoteUnit(unit *workUnit, unitID string) {
 	// TODO: handle cancellation of the overall monitor
 	defer func() {
-		unit.lock.Lock()
-		unit.monitoringRemote = false
-		unit.lock.Unlock()
+		unit.waitRemote.Done()
 	}()
 	var conn net.Conn
 	var nextDelay = SuccessWorkSleep
@@ -440,8 +438,8 @@ func (w *Workceptor) monitorRemoteUnit(unit *workUnit, unitID string) {
 	remoteNodeID := unit.status.Node
 	remoteUnitID := unit.status.RemoteUnitID
 	workType := unit.status.WorkType
-	unit.monitoringRemote = true
 	unit.lock.Unlock()
+	unit.waitRemote.Add(1)
 	for {
 		if conn != nil && !reflect.ValueOf(conn).IsNil() {
 			_ = conn.Close()
@@ -583,6 +581,7 @@ func (w *Workceptor) PreStartUnit(nodeID string, workTypeName string, params str
 		lock:     &sync.RWMutex{},
 		started:  false,
 		released: false,
+		waitRemote: &sync.WaitGroup{},
 		worker:   worker,
 		status:   status,
 	}
@@ -651,7 +650,7 @@ func (w *Workceptor) scanForUnits() {
 					lock:     &sync.RWMutex{},
 					started:  true, // If we're finding it now, we don't want to start it again
 					released: false,
-					monitoringRemote: false,
+					waitRemote: &sync.WaitGroup{},
 					worker:   nil,
 					status:   si,
 				}
@@ -766,19 +765,19 @@ func (w *Workceptor) cancelRemote(remoteNodeID, remoteUnitID string, unit *workU
 }
 
 // CancelUnit cancels a unit, killing it if it is still running
-func (w *Workceptor) CancelUnit(unitID string) error {
-	//TODO: Handle remote cancellation
+func (w *Workceptor) CancelUnit(unitID string) (bool, error) {
+	isPending := false
 	w.scanForUnits()
 	w.activeUnitsLock.RLock()
 	unit, ok := w.activeUnits[unitID]
 	w.activeUnitsLock.RUnlock()
 	if !ok {
-		return fmt.Errorf("unknown work unit %s", unitID)
+		return isPending, fmt.Errorf("unknown work unit %s", unitID)
 	}
 	if unit.worker != nil {
 		err := unit.worker.Cancel()
 		if err != nil {
-			return err
+			return isPending, err
 		}
 	}
 	unit.lock.Lock()
@@ -792,6 +791,7 @@ func (w *Workceptor) CancelUnit(unitID string) error {
 		unit.status.Detail = "Cancel Pending"
 		unit.lock.Unlock()
 		go w.cancelRemote(nodeID, remoteUnitID, unit)
+		isPending = true
 	} else {
 		unit.lock.Lock()
 		unit.status.State = WorkStateFailed
@@ -804,9 +804,9 @@ func (w *Workceptor) CancelUnit(unitID string) error {
 	w.activeUnitsLock.RUnlock()
 	if err != nil {
 		logger.Error("Error saving local status file: %s\n", err)
-		return nil
+		return isPending, nil
 	}
-	return nil
+	return isPending, nil
 }
 
 // ReleaseUnit releases a unit, canceling it if it is still running
@@ -823,20 +823,13 @@ func (w *Workceptor) ReleaseUnit(unitID string) error {
 		return fmt.Errorf("unknown work unit %s", unitID)
 	}
 
-	err := w.CancelUnit(unitID)
+	_, err := w.CancelUnit(unitID)
 	if err != nil {
 		return err
 	}
 
-	for {
-		unit.lock.RLock()
-		monitoringRemote := unit.monitoringRemote
-		unit.lock.RUnlock()
-		if !monitoringRemote {
-			break
-		}
-		time.Sleep(250 * time.Millisecond)
-	}
+	// wait for monitoringRemoteUnit to return
+	unit.waitRemote.Wait()
 
 	if remoteUnitID != "" {
 		conn, _, err := w.connectToRemote(nodeID)
@@ -1071,12 +1064,16 @@ func (w *Workceptor) workFunc(params string, cfo controlsvc.ControlFuncOperation
 		if len(tokens) != 2 {
 			return nil, fmt.Errorf("bad command")
 		}
-		err := w.CancelUnit(tokens[1])
+		isPending, err := w.CancelUnit(tokens[1])
 		if err != nil {
 			return nil, err
 		}
 		cfr := make(map[string]interface{})
-		cfr["cancelled"] = tokens[1]
+		if isPending {
+			cfr["cancel pending"] = tokens[1]
+		} else {
+			cfr["cancelled"] = tokens[1]
+		}
 		return cfr, nil
 	case "results":
 		if len(tokens) < 2 || len(tokens) > 3 {
--- a/pkg/workceptor/workceptor.go
+++ b/pkg/workceptor/workceptor.go
@@ -54,6 +54,7 @@ type StatusInfo struct {
 	Node         string
 	WorkType     string
 	RemoteUnitID string
+	LocalCancel  bool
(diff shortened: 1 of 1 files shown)


pkg/workceptor/workceptor.go
```diff
@@ -689,6 +689,31 @@ func (w *Workceptor) CancelUnit(unitID string) error {
 		return fmt.Errorf("unknown work unit %s", unitID)
 	}
 	w.activeUnitsLock.RUnlock()
+	remoteUnitID := unit.status.RemoteUnitID
```
ghjm: We should get a lock on `unit.lock` here.  (This was my error, we should have already had a lock before manipulating the unit status at the bottom of the function.)

ghjm:
Conceptually, do we want this to be "you call CancelUnit and it either works or fails, and then it's your problem if you want to retry?"  Or do we want this to be "you submit a cancel request and the monitor process will keep trying to cancel remotely until it eventually succeeds?"

pkg/workceptor/workceptor.go
```diff
@@ -689,6 +689,31 @@ func (w *Workceptor) CancelUnit(unitID string) error {
...
+		if err != nil {
+			logger.Error("Write error sending to %s: %s\n", remoteNodeID, err)
+			return err
+		}
+		response, err := reader.ReadString('\n')
+		if err != nil {
+			logger.Error("Read error reading from %s: %s\n", remoteNodeID, err)
+			return err
+		}
+		cancelledRegex := regexp.MustCompile(fmt.Sprintf("cancelled.*%s", remoteUnitID))
```
fosterseth: maybe this is not the proper way to check. We could monitor the remote work status.State instead

ghjm: Yes ^^

pkg/workceptor/workceptor.go
```diff
@@ -322,6 +323,9 @@ func (w *Workceptor) monitorRemoteStatus(ctx context.Context, cancel context.Can
 			logger.Error("Error saving local status file: %s\n", err)
 			return
 		}
+		if unit.status.State == WorkStateFailed {
```
fosterseth: this makes sure we stop monitoring if the cancel goes through on the remote node.

pkg/workceptor/workceptor.go
```diff
@@ -631,6 +636,11 @@ func (w *Workceptor) scanForUnits() {
 				}
 				w.activeUnits[fi.Name()] = unit
 				if si.Node != "" && si.Node != w.nc.NodeID() {
+					if si.LocalCancel {
```
fosterseth, line 639: on control node restart, this will make sure we restart the cancellation goroutine

pkg/workceptor/workceptor.go
```diff
@@ -678,25 +688,91 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
 	return retMap, nil
 }
 
+func (w *Workceptor) attemptRemoteAction(remoteNodeID, remoteUnitID, remoteAction string, unit *workUnit) error {
+	timeout := time.After(300 * time.Second)
```
fosterseth: times out after 5 minutes. This might need to be a configurable parameter

pkg/workceptor/workceptor.go
```diff
@@ -678,25 +688,91 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
...
+			}
+			conn.Close()
+		}
+
+		if shouldTimeout() {
+			return nil
+		}
+	}
+
+   if remoteAction == "cancel" {
```
fosterseth: `attemptRemoteAction` is used both for release and cancel, in the case of cancel we wait around to see if the status actually changes.

pkg/workceptor/workceptor.go
```diff
@@ -631,6 +636,11 @@ func (w *Workceptor) scanForUnits() {
 				}
 				w.activeUnits[fi.Name()] = unit
 				if si.Node != "" && si.Node != w.nc.NodeID() {
+					if si.LocalCancel {
+						remoteNodeID := si.Node
+						remoteUnitID := si.RemoteUnitID
+						go w.attemptRemoteAction(remoteNodeID, remoteUnitID, "cancel", unit)
+					}
 					go w.monitorRemoteUnit(unit, fi.Name())
```
fosterseth, line 663: even if LocalCancel is true, we still want to monitor the remote node until we get a WorkStateFailed from the remote status, thus we run the monitor even if we've issued the cancel.

pkg/workceptor/workceptor.go
```diff
@@ -631,6 +636,11 @@ func (w *Workceptor) scanForUnits() {
 				}
 				w.activeUnits[fi.Name()] = unit
 				if si.Node != "" && si.Node != w.nc.NodeID() {
+					if si.LocalCancel && si.State == WorkStateRunning {
```
fosterseth: on control node restart, this will make sure we restart the cancellation goroutine

fosterseth: I realize we need a way to determine if a previous cancel attempt timed out, and not re-try a cancel in that case. Otherwise, every time we restart the control service it will respawn the go routines to cancel previously timed out cancellations.

ghjm:
What if the timeout logic was:

- On a `cancel` request, we keep retrying indefinitely, including after a restart
- On a `release` request, we only make a single attempt and just report back an error if it fails

The reason I'm suggesting this is that I think a `cancel` without a `release` is the end-user telling us "I do still care about this job and I actively want it cancelled."  And sometimes cancelling a job might be really important (I want to cancel my `rm -rf /` job before it destroys everything).  So I don't think we really have a warrant to say "sorry, user, your cancel didn't go through and we're just giving up on it."  If the user wants us to give up on it, they have a way to do that - they can do a `release`.

On a `release` request, the user is saying "I no longer care about this job."  For all we know, the user has decommissioned the remote Receptor node and we're never going to talk to it again.  In this case I think we should make a best-reasonable-effort attempt at cancelling, and then just give up if it doesn't work.  But maybe the semantics here should be that we don't release control back to the caller until the release goes through or times out, and if it times out, we return an error message saying "remote node unreachable: remote job not released" (or words to that effect).  I don't think we need a timeout here - if the remote Receptor node is unresponsive, not in the routing table, etc, we just return this error immediately.  The reason I think we should hold the client response is that once the local release goes through, there's no way to inform the client that the remote release failed.

Thoughts?

pkg/workceptor/workceptor.go
ghjm: Shouldn't we also stop monitoring the cancel if the status is WorkStateSucceeded?

Also, there's an `IsComplete` helper function, so you can do `if IsComplete(unit.status.State)` which returns true if it's failed or succeeded, and is perhaps a little more future-proof if we ever add more work states.

pkg/workceptor/workceptor.go
ghjm: What if it's WorkStatePending?  Then we don't launch the monitor, and maybe the job goes to WorkStateRunning a second later.  Maybe this should be `if si.LocalCancel && !IsComplete(si.State)`.

pkg/workceptor/workceptor.go
```diff
@@ -678,25 +688,91 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
 	return retMap, nil
 }
 
+func (w *Workceptor) attemptRemoteAction(remoteNodeID, remoteUnitID, remoteAction string, unit *workUnit) error {
+	timeout := time.After(300 * time.Second)
+	retryInterval := 500 * time.Millisecond
```
ghjm: Perhaps instead of a fixed interval, this should be a backoff scheme, where we increase the retry interval on failures, to some maximum limit.

pkg/workceptor/workceptor.go
```diff
@@ -678,25 +688,91 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
...
+	shouldTimeout := func() bool {
+		select {
+		case <-timeout:
+			logger.Error("%s of remote work unit timed out", remoteAction)
+			return true
+		default:
+			time.Sleep(retryInterval)
```
ghjm, line 727: Instead of doing a `time.Sleep` in the `default` section, it would be better to do a `case <- time.After(retryInterval)`

pkg/workceptor/workceptor.go
```diff
@@ -678,25 +688,91 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
 	return retMap, nil
 }
 
+func (w *Workceptor) attemptRemoteAction(remoteNodeID, remoteUnitID, remoteAction string, unit *workUnit) error {
+	timeout := time.After(300 * time.Second)
+	retryInterval := 500 * time.Millisecond
+	shouldTimeout := func() bool {
+		select {
+		case <-timeout:
```
ghjm: The global timeout might be better treated as a `Context` that a caller could pass in.   Then we just check `context.Done()` and it's the caller's business if they want to put a timeout on the context or not.

pkg/workceptor/workceptor.go
```diff
@@ -678,25 +688,91 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
...
+		var state int
+		for {
+			w.activeUnitsLock.RLock()
+			state = unit.status.State
+			w.activeUnitsLock.RUnlock()
+			if state == WorkStateSucceeded || state == WorkStateFailed {
+				return nil
+			}
+
+			if shouldTimeout() {
```
ghjm: It's a bit weird that we're calling a function named `shouldTimeout` that is actually doing the timeout.  Maybe it should be called `doTimeout`, or just be unrolled into this loop and not be a function at all.  (Also I think most of this logic goes away if we use a Context instead.)

pkg/workceptor/workceptor.go
```diff
@@ -678,25 +688,91 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
...
+	}
+	return nil
+}
+
 // CancelUnit cancels a unit, killing it if it is still running
 func (w *Workceptor) CancelUnit(unitID string) error {
 	//TODO: Handle remote cancellation
 	w.scanForUnits()
 	w.activeUnitsLock.RLock()
+	defer w.activeUnitsLock.RUnlock()
```
ghjm: Golang can't stack locks, so if you're holding this lock all the way to the end of the function, you can't call any _other_ function that tries to acquire the lock.  I guess the original reason I did this was for the cancel call, which is now moved to a goroutine.  Are we sure the `unit.status.Save` call is okay?

pkg/workceptor/workceptor.go
```diff
@@ -678,25 +688,91 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
...
 			return err
 		}
 	}
-	unit.status.State = WorkStateFailed
-	unit.status.Detail = "Cancelled"
+	unit.status.LocalCancel = true
+	if unit.status.RemoteUnitID != "" {
+		remoteNodeID := unit.status.Node
+		remoteUnitID := unit.status.RemoteUnitID
+		go w.attemptRemoteAction(remoteNodeID, remoteUnitID, "cancel", unit)
```
ghjm: Should we set the local `unit.status.Detail` to "Cancelling" or "Locally Cancelled" or something like that here?

pkg/workceptor/workceptor.go
fosterseth: I've now implemented a backoff scheme

pkg/workceptor/workceptor.go
fosterseth: changed it to `!IsComplete(si.State)`

pkg/workceptor/workceptor.go
```diff
@@ -679,44 +714,146 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
 	return retMap, nil
 }
 
+func (w *Workceptor) cancelRemote(remoteNodeID, remoteUnitID string, unit *workUnit) error {
+	retryInterval := 500 * time.Millisecond
```
ghjm: Maybe set the initial `retryInterval` to `SuccessWorkSleep` which we already have a constant for?

pkg/workceptor/workceptor.go
```diff
@@ -679,44 +714,146 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
...
+		}
+		if conn != nil {
+			_, err = conn.Write([]byte(fmt.Sprintf("work cancel %s\n", remoteUnitID)))
+			if err != nil {
+				logger.Error("Write error sending to %s: %s\n", remoteNodeID, err)
+			} else {
+				conn.Close()
+				break
+			}
+			conn.Close()
```
ghjm, line 749: We should read the response to the cancel command and report it if it is an error.  Errors should always be of the form "ERROR: <description" so it should be sufficient to just check if the string starts with "ERROR".

pkg/workceptor/workceptor.go
```diff
@@ -679,44 +714,146 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
...
+			return nil
+		}
+		unit.lock.RLock()
+		state = unit.status.State
+		unit.lock.RUnlock()
+		if IsComplete(state) {
+			return nil
+		}
+		time.Sleep(retryInterval)
+	}
```
ghjm: What does this code do?

pkg/workceptor/workceptor.go
```diff
@@ -679,44 +714,146 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
...
 	w.scanForUnits()
 	w.activeUnitsLock.RLock()
 	unit, ok := w.activeUnits[unitID]
+	w.activeUnitsLock.RUnlock()
 	if !ok {
-		w.activeUnitsLock.RUnlock()
 		return fmt.Errorf("unknown work unit %s", unitID)
 	}
-	w.activeUnitsLock.RUnlock()
 	if unit.worker != nil {
```
ghjm, line 780: This might be a bug in the existing code.  Is it legal to access `unit.worker` like this while not holding the lock on the unit?

pkg/workceptor/workceptor.go
```diff
@@ -679,44 +714,146 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
...
 		}
 	}
-	unit.status.State = WorkStateFailed
-	unit.status.Detail = "Cancelled"
+	unit.lock.Lock()
+	unit.status.LocalCancel = true
+	nodeID := unit.status.Node
+	remoteUnitID := unit.status.RemoteUnitID
+	released := unit.released
+	unit.lock.Unlock()
```
ghjm: Probably better not to unlock and immediately re-lock like this, because what happens if some other process changes these values during the interval when we're not holding the lock?

pkg/workceptor/workceptor.go
```diff
@@ -679,44 +714,146 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
...
+		unit.lock.Unlock()
+		go w.cancelRemote(nodeID, remoteUnitID, unit)
+	} else {
+		unit.lock.Lock()
+		unit.status.State = WorkStateFailed
+		unit.status.Detail = "Cancelled"
+		unit.lock.Unlock()
+	}
+	unitdir := path.Join(w.dataDir, unitID)
+	w.activeUnitsLock.RLock()
```
ghjm: I don't see why we need to be holding the activeUnitsLock here?

pkg/workceptor/workceptor.go
```diff
@@ -679,44 +714,146 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
...
+		conn, _, err := w.connectToRemote(nodeID)
+		if err != nil {
+			logger.Error("Connection failed to %s: %s\n", nodeID, err)
+		}
+		if conn != nil {
+			_, err = conn.Write([]byte(fmt.Sprintf("work release %s\n", remoteUnitID)))
+			if err != nil {
+				logger.Error("Write error sending to %s: %s\n", nodeID, err)
+			}
+			conn.Close()
```
ghjm, line 853: Would be nice to check the result here as well, and log an error ... although on a release, I guess we don't care as much.

ghjm:
I experimented with this and tried running it with the remote node up, down, etc.  Everything worked as I would expect.

One thing I found a bit disconcerting was that if you cancel a remote unit of work while the remote node is unreachable, you still get a response like `{"cancelled":"JvnNj9kH"}`.  I feel like if we haven't actually succeeded in doing the remote cancel, this response should say something like "cancel pending" ... maybe?  What do you think?

pkg/workceptor/workceptor.go
fosterseth: yeah I didn't want the lock / unlock to cross over the `if` statement. I'm not sure of a more graceful way to handle that.

pkg/workceptor/workceptor.go
fosterseth: yeah I think was supposed to be `unit.lock.RLock() ` instead

pkg/workceptor/workceptor.go
ghjm: I don't think there's a more graceful way than just unlocking it at the right time in both the if and the else clause.

pkg/workceptor/workceptor.go
fosterseth: my understanding is that no fields of `unit` are read or touched in the calling of `Cancel()`

pkg/workceptor/workceptor.go
ghjm: Actually you could just leave it locked until after the  `unit.status.Save()` below.  The `cancelRemote` call will block until the lock is released, but that's okay, because it's called as a goroutine, so it's not deadlocking _our_ thread.  If the goroutine happens to get scheduled before we're done with the lock, that's okay, it will just block until we release it.

pkg/workceptor/workceptor.go
```diff
@@ -679,44 +714,146 @@ func (w *Workceptor) unitStatusForCFR(unitID string) (map[string]interface{}, er
...
+		unit.status.Detail = "Cancelled"
+		unit.lock.Unlock()
+	}
+	unitdir := path.Join(w.dataDir, unitID)
+	w.activeUnitsLock.RLock()
+	err := unit.status.Save(path.Join(unitdir, "status"))
+	w.activeUnitsLock.RUnlock()
+	if err != nil {
+		logger.Error("Error saving local status file: %s\n", err)
+		return nil
```
ghjm: Shouldn't we return the actual error here?

pkg/workceptor/workceptor.go
fosterseth: this code checks that we are getting a job state change on the remote node. But I realize I don't actually do anything with this information, so it is kind of pointless.

pkg/workceptor/workceptor.go
ghjm: 👍

pkg/workceptor/workceptor.go
fosterseth: fixed this

pkg/workceptor/workceptor.go
fosterseth: I now attempt to read the result here and report back if the remote node did not release correctly

ghjm (approved):

PR #61, opened 2020-07-16
Fix blocking situations that were preventing clean shutdown

ghjm:
This PR adds the following fixes:

- Use a context in backend `Listen()` and `Dial()` calls, with a particular goal of having `Accept()` be released on context cancel
- Use contexts instead of `ErrorChan` to end back-end sessions, so they exit if the whole Netceptor exits without extra logic
- Do not send to `updateRoutingTableChan` and `sendRouteFloodChan` if we are shutting down, because those goroutines may have already exited

### DIFF
--- a/example/net.go
+++ b/example/net.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"github.com/project-receptor/receptor/pkg/backends"
+	"github.com/project-receptor/receptor/pkg/logger"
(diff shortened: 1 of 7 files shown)


fosterseth (approved):

squidboylan:
This doesnt actually shutdown the TCP ports, i think because the context doesnt work like we think, I think we need to call `listener.Close()` in order to cancel the `listener.Accept()`

PR #65, opened 2020-07-17
Use timeouts in BackendSession Recv method to fix leaking goroutine

ghjm:
The goroutine leak in issue https://github.com/project-receptor/receptor/issues/62 is caused by `BackendSession.Recv()` being a blocking call, so the `protoReader` goroutine never exits.  This PR adds a timeout value to `Recv()` so that `protoReader` has a chance to periodically check if the context is done.

The timeout implementation in the websockets backend cannot use `SetReadDeadLine` like the other backends do, because of https://github.com/gorilla/websocket/issues/474.  After a timeout error occurs, the websocket connection is no longer usable and must be closed.  So instead, we start a goroutine called `recvChannelizer` which pushes messages to a channel, and exits after any error.  The websockets backend `Recv()` message then provides "soft" timeouts using a select on this channel or the timeout.

### DIFF
--- a/pkg/backends/tcp.go
+++ b/pkg/backends/tcp.go
@@ -156,13 +156,20 @@ func (ns *TCPSession) Send(data []byte) error {
 }
 
 // Recv receives data via the session
-func (ns *TCPSession) Recv() ([]byte, error) {
+func (ns *TCPSession) Recv(timeout time.Duration) ([]byte, error) {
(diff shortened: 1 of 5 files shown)


squidboylan (approved):
feels like there should be a more event/context based way to do this but it fixes my test so im happy with it

fosterseth (approved):

PR #68, opened 2020-07-20
Debug log the actual address our backends listen on

squidboylan:


ghjm (changes requested):
By having this only in the `Run` method and not in `Start`, we will get a debug log _only_ when running from the command line, not when being called via the API.  I agree we probably don't need two log events, but if we're only going to have one, I think it should be in `Start`.

squidboylan:
> By having this only in the `Run` method and not in `Start`, we will get a debug log _only_ when running from the command line, not when being called via the API. I agree we probably don't need two log events, but if we're only going to have one, I think it should be in `Start`.

Yeah i was actually just thinking about this, i'll update it

ghjm (approved):

PR #71, opened 2020-07-22
Add work-kubernetes service

ghjm:
This adds a simple Kubernetes worker service that can be used as follows:
```
- work-kubernetes:
    service: kubecat
    kubeconfig: /home/graham/.kube/config
    project: graham-test
    image: centos:8
```
With this defined, we can do
```
seq 10 | receptorctl work submit kubecat -p - -f --rm
```
which will launch a pod named `kubecat-<random>` in the graham-test namespace in the k8s cluster we're logged in to in my kubeconfig file, run a container in it using the centos:8 image, and run the bash command.  It will receive my list of numbers from 1 to 10 on its stdin, and repeat them to stdout, which will then be printed to my terminal on the machine running receptorctl (because it has `-f`).  Once the work is complete, the pod will be removed, and the Receptor unit of work will be released (because it has `--rm`).


pkg/workceptor/kubernetes.go
```diff
@@ -0,0 +1,326 @@
...
+
+// **************************************************************************
+// Command line
+// **************************************************************************
+
+// WorkKubeCfg is the cmdline configuration object for a Kubernetes worker plugin
+type WorkKubeCfg struct {
+	Service    string `required:"true" description:"Local Receptor service name to bind to"`
+	KubeConfig string `description:"Kubeconfig file (defaults to environment)"`
+	Project    string `required:"true" description:"Kubernetes project name"`
```
shanemcd: Should this be `Namespace`?

shanemcd:
Left a super helpful comment because I couldn't help it, but will give a more in-depth review tomorrow.

pkg/workceptor/kubernetes.go
ghjm: It's called a Project in all the user-facing screens in the OpenShift GUI, but a Namespace in all the Kubernetes API calls.  Not sure which we would consider more correct here, but I'm happy to change it if you think Namespace is better.

ghjm: I guess nothing else in this is OpenShift specific, and there really isn't such a thing as a "Kubernetes project," so you're right, this should be changed.

ghjm:
This PR is complete as far as the ability to run jobs from a stable Receptor node.  It does not solve the problem of preserving state through a Receptor restart (https://github.com/project-receptor/receptor/issues/86) or passing runtime information such as kubeconfig / k8s authentication (https://github.com/project-receptor/receptor/issues/78).  These issues require a larger rewrite of the workceptor core code.  I'd like to get this and https://github.com/project-receptor/receptor/pull/59 merged first, then look at more significant changes to workceptor.

fosterseth (approved):

PR #76, opened 2020-07-24
Acquire listenerLock before accessing listenerRegistry

ghjm:
Fixes a function where we mistakenly access listenerRegistry without first acquiring listenerLock.  See https://github.com/project-receptor/receptor/issues/75.

### DIFF
--- a/pkg/netceptor/netceptor.go
+++ b/pkg/netceptor/netceptor.go
@@ -802,7 +802,9 @@ func (s *Netceptor) handleMessageData(md *messageData) error {
 		if handled {
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


squidboylan:
There seems to be some sort of deadlock or something (connections timing out) that i need to investigate with this. I'm not entirely sure if it's a receptor problem or a problem in my test suite, it seems like it would be in the receptor code though

squidboylan (approved):
I'm fairly certain the issue im hitting with my updated tests and this is a problem with the tests not this code we can merge this (and it wont break ci since my tests that are breaking are on a different branch)

PR #85, opened 2020-08-07
Add CancelRead to netceptor.Conn and use it in Workceptor

ghjm:
The semantics of `netceptor.Conn.Close()` are that it closes the writer side of the connection but not the reader.  This mirrors the semantics of TCP, where "close" means, essentially, "send a FIN" - and obviously you can't call a function on hostA that causes hostB to send a FIN because that would be spooky action at a distance.

We probably want these semantics to stay the way they are, but Workceptor needs a way to cancel its stdout `io.Copy()` process when we become aware that a unit of work is cancelled or released.  Luckily, go-quic implements a `CancelRead()` function for just this purpose.  So what this PR does is add a `netceptor.Conn.CancelRead()` that can be called by Workceptor to unblock the `io.Copy()`.  HTTP status code 499 is used because it kinda sorta means "closed by user."

### DIFF
--- a/pkg/workceptor/workceptor.go
+++ b/pkg/workceptor/workceptor.go
@@ -322,7 +322,9 @@ func (w *Workceptor) monitorRemoteStatus(ctx context.Context, cancel context.Can
 			logger.Error("Error saving local status file: %s\n", err)
 			return
 		}
-		sleepOrDone(ctx.Done(), 1*time.Second)
+		if sleepOrDone(ctx.Done(), 1*time.Second) {
+			return
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/workceptor/workceptor.go
```diff
@@ -383,7 +385,22 @@ func (w *Workceptor) monitorRemoteStdout(ctx context.Context, cancel context.Can
 				logger.Debug("Could not open stdout file %s: %s\n", stdoutFilename, err)
 				continue
 			}
+			doneChan := make(chan struct{})
+			go func() {
+				select {
+				case <-doneChan:
+					return
+				case <-ctx.Done():
+					cr, ok := conn.(interface{ CancelRead() })
```
fosterseth, line 394: this checks that the interface is defined for `conn`? Trying to understand when `ok` is false

fosterseth (approved):
tested the patch and it worked -- releasing the local job on node B will break out of `monitorRemoteStdout` on node A.

pkg/workceptor/workceptor.go
ghjm: Since `conn` is a `net.Conn`, we only know that it has the functions defined in the `net.Conn` interface declaration, which don't include `CancelRead()`.  But I happen to know that this particular `net.Conn` is probably always going to be a `netceptor.Conn`, which _does_ have a `CancelRead()`.  So what I'm doing here is using an anonymous interface `interface { CancelRead() }` to get access to the `CancelRead()` method via type assertion.  Without the `ok` this would panic if the type assertion failed, and I don't want that - I want it to still try its best to work if you pass it something other than a `netceptor.Conn`.  So I use the two variable version of type assertion, where you get a boolean that tells you if the assertion worked.  If `ok` comes back `True` then we know it's okay to call `cr.CancelRead()`.  If it comes back `False` then we have a non-netceptor Conn and just have to hope it will figure out how to unblock itself when we call `Close()` on it.

PR #92, opened 2020-08-19
Fix a routingtable read without holding the lock

squidboylan:


ghjm (changes requested):
Good find.  But since we're only reading from this, not writing to it, we should use RLock and RUnlock.

ghjm (approved):

PR #93, opened 2020-08-19
Switch CI to using offline-built images instead of building each time

ghjm:
The existing CI has a janky scheme where we start a local Docker registry within the GitHub Actions worker, push to it, and then try to use the Actions cache to save and restore its content.  This is supposed to be marginally faster according to some web site I read once.  In fact what happens is that the Actions cache gets confused about whether something has changed or not, and the CI action always winds up rebuilding the image anyway, so instead of saving ten seconds we're back to spending 5-10 minutes yum installing packages on every commit.

This PR switches to using the GitHub Package Registry to store a CI runner image built once a day.  On pull requests, we pull this image and run the CI in it.  Pulling the image from the registry does take slightly longer than restoring it from the Actions cache, but not enough to matter, and this scheme is much more robust.


fosterseth (approved):
awesome, looking forward to this change

Issue #31, opened 2020-06-30
Allow Listeners to set a cost per dialer


squidboylan:
In the current implementation of connection costs, a listener sets a cost and a dialer must agree with it in order to connect. This restricts the flexibility of graph costs because all dialers into a single listener must have the same cost, it's not hard to imagine a network in which you would want different dialers into the same listener to have different costs (for example if they have different latencies to the same node).

There are reasons to have some negotiation of cost, for example it gives the listener some control over preventing misbehaving dialers from assigning a very cheap cost and everything getting routed through them.

A proposed solution is a listener could have a map of minimum costs for each dialer, this gives some control to the listener while allowing the dialers to set their own cost.

ghjm:
There are two related issues here:

1. The config format is inflexible and requires a listener to have a single connection cost for all nodes that dial into it.
2. Pairs of peers are required to have the same cost statically configured in their configuration, and don't do any run-time negotiation.

Item 1 means that if you want to receive inbound connections from two peers with different path costs, you _must_ run two listeners on different ports (or different `bindaddr`s).  This could be solved by having listener cost configuration be map of costs by peer node ID, instead of just a single number.  I think this is a pretty straightforward feature improvement.  The main reason I haven't done it already is that it's awkward from a CLI perspective (`--tcp-listener cost=<what goes here?>`).

Item 2 could be used to provide better dynamic behavior of the mesh network, for example by increasing the cost of a path at run-time if high latency or packet loss is observed over it.  This would need to be done very carefully to avoid introducing things like routing loops, instability, non-convergence, etc.  This is a much larger effort that I don't foresee tackling in the short term.

ghjm:
Some thoughts on how this should work:
* Both sides of the connection still have to agree on a cost - I don't think there should be any changes to the network protocol.
* Dialers still only get one cost.  This is also unchanged.
* Listeners need to be able to specify a cost per peer (dialer), as a map of node ID to cost.  I think it's okay if this uses the YAML map syntax in the config file.  If you want to specify a map of node IDs to costs on the command line, you'll have to write JSON - but anyone configuring their listener at this level of complexity should be using a config file anyway.
* On connection, a listener will have to decide what it thinks the connection cost should be, and it can only do this after it discovers the node ID of the dialer.  This might get a bit hairy since the protocol start is a single function called by both dialers and listeners, so this change may need to introduce some asymmetry.
[...]

PR #94, opened 2020-08-19
Allow user to define a cost per remote node

fosterseth:
This commit will allow users to specify a map of nodes and costs.

```yaml
- tcp-listener:
    port: 2222
    cost: 1.0
    nodecost:
        bar: 1.5
        fish: 2.0
```

or from command line

`./receptor --node foo --tcp-listener nodecost='{"bar":1.5,"fish":2.0}'`

Nodes must still agree on cost.

If `nodecost[bar]` is not defined, then it uses the value of `cost`.

issue #31

### DIFF
--- a/cmd/receptor.go
+++ b/cmd/receptor.go
@@ -47,7 +47,7 @@ func (cfg nullBackendCfg) Start(ctx context.Context) (chan netceptor.BackendSess
 
 // Run runs the action, in this case adding a null backend to keep the wait group alive
 func (cfg nullBackendCfg) Run() error {
-	err := netceptor.MainInstance.AddBackend(&nullBackendCfg{}, 1.0)
+	err := netceptor.MainInstance.AddBackend(&nullBackendCfg{}, 1.0, make(map[string]float64))
 	if err != nil {
 		return err
 	}
--- a/cmd/receptor.go
+++ b/cmd/receptor.go
@@ -47,7 +47,7 @@ func (cfg nullBackendCfg) Start(ctx context.Context) (chan netceptor.BackendSess
 
 // Run runs the action, in this case adding a null backend to keep the wait group alive
 func (cfg nullBackendCfg) Run() error {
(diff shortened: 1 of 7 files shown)


cmd/receptor.go
```diff
@@ -47,7 +47,7 @@ func (cfg nullBackendCfg) Start(ctx context.Context) (chan netceptor.BackendSess
 
 // Run runs the action, in this case adding a null backend to keep the wait group alive
 func (cfg nullBackendCfg) Run() error {
-	err := netceptor.MainInstance.AddBackend(&nullBackendCfg{}, 1.0)
+	err := netceptor.MainInstance.AddBackend(&nullBackendCfg{}, 1.0, make(map[string]float64))
```
ghjm: You can just pass in `nil` here instead of allocating a map.  If your map value is nil, reads from it will always return non-OK.  (Same comment applies to all the other places where you're passing in an empty map.)

pkg/backends/tcp.go
```diff
@@ -203,6 +203,7 @@ type TCPListenerCfg struct {
 	Port     int     `description:"Local TCP port to listen on" barevalue:"yes" required:"yes"`
 	TLS      string  `description:"Name of TLS server config"`
 	Cost     float64 `description:"Connection cost (weight)" default:"1.0"`
+	NodeCostMap map[string]float64 `description:"Connection cost (weight) for specific node"`
```
ghjm: Let's call this `NodeCosts` instead of `NodeCostMap` just for simplicity and readability in the config file.

ghjm (changes requested):

pkg/backends/tcp.go
fosterseth: fixed

cmd/receptor.go
fosterseth: fixed

ghjm (approved):

PR #97, opened 2020-08-19
Update the image PATH so our built receptor binary is in our PATH

squidboylan:


squidboylan:
#96 does this instead

PR #106, opened 2020-08-24
Cost map testing

squidboylan:


ghjm (approved):
I haven't pulled this down and run it myself, but if CI runs then this must have run right? 😄

PR #109, opened 2020-08-24
Workceptor refactor

ghjm:
Refactoring Workceptor to remove some cruft and fix some issues.

In particular the `StatusInfo` struct has become too much of a catch-all and has a bunch of struct members that are only used by one or another worker type, so any given instance has a lot of dead fields, to the point that it has become confusing.  Further development, particularly of the Kubernetes worker type, would need to add several more fields to this struct.  It is also not clear from the existing code which structs are saved to files and which are used in-memory.

To fix this, a new type, `StatusFileData`, replacing `StatusInfo`, is used only for data that is persisted to the status file.  This struct has a field `ExtraData` that can be used by particular worker implementations to store data relevant only to them.  Each worker type implements the new `WorkUnit` interface.  Functions common to all implementations are provided by `BaseWorkUnit`, with individual implementations extending this via struct composition.

API changes resulting from this PR:

- In the command line and config files, the worker identifier is now called `WorkType` instead of `Service`
- Both `Cancel()` and `Release()` can return ErrPending if the first attempt to do a remote cancel or release fails. They will then continue to run in the background and perform the requested operation when the remote node becomes available.
- `Restart()` now takes a `force` parameter.  If true, we will only make one attempt to release remote resources, and will then do a local release whether or not the remote release succeeded.  This ensures the unit will be deleted locally, but has the potential to leave the remote resources orphaned.
- Several functions now expect to be passed a `Context`, which allows operations to be aborted or timed out more easily.
The `PreStartUnit()` function has been replaced with `AllocateUnit()` and `AllocateRemoteUnit()`.

Other notable changes include:

- File locks are now used to ensure that updates to status files are performed atomically.
- The `sockutils` package is renamed to `utils` because it now contains non-socket-related utilities such as `JobContext`.
- Remote unit handling is now a separate `WorkUnit` implementation called `remoteUnit`.
- Work unit implementations now have a `Restart()` method that creates a worker from an already-existing status file on disk.  `commandUnit` and `remoteUnit` can use this to re-establish monitoring of still-running processes after a Receptor restart.


squidboylan:
are we using `SingleUnlocker` anywhere? It doesnt look like it to me. Other than that I dont see anything that looks out of place

ghjm:
Looks like I'm still using SingleUnlocker in just one place, in remote_work.py, but that could easily be done without it, The function where this was used a lot got refactored out during development.  I'll just take it out.

pkg/workceptor/remote_work.go
```diff
@@ -0,0 +1,479 @@
...
+		ed := status.ExtraData.(*remoteExtraData)
+		ed.LocalStarted = true
+	})
+	return nil
+}
+
+// cancelOrReleaseRemoteUnit makes a single attempt to cancel or release a remote unit.
+func (rw *remoteUnit) cancelOrReleaseRemoteUnit(ctx context.Context, conn net.Conn, reader *bufio.Reader,
+	release bool, force bool) error {
+	closeOnce := sync.Once{}
```
fosterseth: is this for future proofing? seems like we only have a single call to doClose() in this method

ghjm: No, it's this way because the function used to be more complicated and had multiple calls to doClose().  I'll simplify it.

fosterseth (approved):

squidboylan (approved):

PR #111, opened 2020-08-25
Artifact test logs

squidboylan:


build/ci-image/Dockerfile
```diff
@@ -4,10 +4,16 @@ RUN dnf -y update && \
...
+RUN go get -u golang.org/x/lint/golint
 # Warm up the Go build cache
-RUN git clone https://github.com/project-receptor/receptor && cd receptor && make build-all && cd .. && rm -rf receptor
+RUN cd /home/test-user && git clone https://github.com/project-receptor/receptor receptor-tmp && cd receptor-tmp && make build-all && cd .. && rm -rf receptor-tmp
 
-COPY entrypoint.sh /entrypoint.sh
-ENTRYPOINT ["/entrypoint.sh"]
+WORKDIR /home/test-user/receptor
+COPY entrypoint.sh /home/test-user/entrypoint.sh
+ENTRYPOINT ["/home/test-user/entrypoint.sh"]
```
ghjm, line 19: Why move the entry point?

squidboylan: just for consistency sake of having everything in the test-user's homedir, we can move it back if that's preferable

ghjm: So the reason for having this user is so you can map UIDs consistently between inside and outside the container?  Does github give us a guarantee of the username or uid that this will run as in the outer context?

squidboylan: yes that is the reason for it, and i dont know if they explicitly say they make that guarantee (I couldnt find anything about the user in the docs and did some print debugging to figure out the uid) but this is the only way i could figure out to make this work. It seems unlikely that they'll update the user but the totally could

squidboylan:
This should be ready for review now, https://github.com/squidboylan/receptor/runs/1063660826 shows an example of this working with a failing job and artifacting the logs

fosterseth (approved):

ghjm (approved):
This is pretty slick.

PR #112, opened 2020-08-25
Get TLS working again in TCP backend

ghjm:
At some point in the rewrites of the TCP listener, TLS got dropped.  This puts it back, and also fixes a bug where the TCP backend dialer timeout was much too short.


fosterseth (approved):

squidboylan:
in logs when i run this locally with my updated SSL tests i see:
```
ERROR 2020/08/25 09:56:36 Backend receiving error tls: first record does not look like a TLS handshake
ERROR 2020/08/25 09:56:36 Backend sending error tls: first record does not look like a TLS handshake
```
and
```
ERROR 2020/08/25 09:55:48 Backend receiving error tls: unsupported SSLv2 handshake received
ERROR 2020/08/25 09:55:48 Backend sending error tls: unsupported SSLv2 handshake received
```
when connecting to a tls server from a client without a tls cert, i expect that's what's breaking ci as well

squidboylan (approved):

PR #117, opened 2020-08-26
Run more tests in parallel

squidboylan:
by default `go test` will run up to GOMAXPROCS tests in parallel, in github actions i think that's only 1, however our tests are pretty light on cpu load so really we should be fine running many many more tests in parallel, seems like 4 - 16 is probably fine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ build-all:
 	go build example/*.go
 
 test:
-	go test ./... -p 1 -count=1
+	go test ./... -p 1 -parallel=16 -count=1
 
 ci: pre-commit build-all test
 	echo "All done"


ghjm (approved):
Doesn't seem to have actually changed the time CI takes though

squidboylan:
> Doesn't seem to have actually changed the time CI takes though

Compared to https://github.com/project-receptor/receptor/runs/1029308403 it is reduced from ~80s to ~55s (-30%), thats the latest run and that PR adds a bit of time because we have to wait some amount of time before we assume the mesh will never connect.

ghjm:
Ah ok, makes sense

PR #118, opened 2020-08-26
Cleanup test timeouts

squidboylan:


ghjm:
I'm not a fan of how these timeouts don't account for the time taken by the checks. Are you sure you don't want to use a pattern like:

```
ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
defer cancel()
var whateverReady bool
for {
    whateverReady = checkWhatever()
    if whateverReady || ctx.Err() {
        break
    }
    time.sleep(100 * time.Millisecond)
}
```

squidboylan:
hmm, yeah i guess now is a good time to do that

ghjm:
Could use something like this to avoid having to copy and paste this code too many times:
```
func UntilWithTimeout(check func() bool, timeout time.Duration, interval time.Duration) bool {
    ctx, cancel := context.WithTimeout(context.Background(), timeout)
    defer cancel()
    for {
        ready := check()
        if ready || (ctx.Err() != nil) {
            return ready
        }
        time.Sleep(interval)
    }
}
```

ghjm (approved):

PR #119, opened 2020-08-26
Add functional test for cancelling remote work

fosterseth:
Uses a 2-node mesh. `node1` starts work on `node2`. Issues a cancel, then a release.

We confirm that work is properly cancelled by issuing a `work status {workid}` command and asserting that `Detail` is `Cancelled` and `StateName` is `Failed`

We confirm that work is properly released by issue a `work list` command and asserting that workid is not in the list.

### DIFF
--- a/tests/functional/lib/receptorcontrol/receptorcontrol.go
+++ b/tests/functional/lib/receptorcontrol/receptorcontrol.go
@@ -172,7 +172,10 @@ func (r *ReceptorControl) WorkSubmit(node, serviceName string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	r.ReadStr()
+	_, err = r.ReadStr() // flush response
+	if err != nil {
+		return "", err
+	}
 	r.CloseWrite()
 	response, err := r.ReadAndParseJSON()
 	if err != nil {
--- a/tests/functional/lib/receptorcontrol/receptorcontrol.go
+++ b/tests/functional/lib/receptorcontrol/receptorcontrol.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"regexp"
 	"strings"
+	"time"
 )
 
 // ReceptorControl Connects to a control socket and provides basic commands
@@ -103,18 +104,26 @@ func (r *ReceptorControl) Close() error {
 	return err
 }
 
+// CloseWrite closes the write side of the socket
+func (r *ReceptorControl) CloseWrite() error {
+	err := r.socketConn.CloseWrite()
+	return err
+}
+
 // ReadAndParseJSON reads data from the socket and parses it as json
 func (r *ReceptorControl) ReadAndParseJSON() (map[string]interface{}, error) {
 	data, err := r.Read()
 	if err != nil {
 		return nil, err
 	}
+
 	jsonData := make(map[string]interface{})
 
 	err = json.Unmarshal(data, &jsonData)
 	if err != nil {
 		return nil, err
 	}
+
 	return jsonData, nil
 }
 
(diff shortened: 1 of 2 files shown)


tests/functional/lib/receptorcontrol/receptorcontrol.go
```diff
@@ -158,3 +165,152 @@ func (r *ReceptorControl) Status() (*netceptor.Status, error) {
...
+	_, err = r.ReadStr() // flush response
+	if err != nil {
+		return "", err
+	}
+	r.CloseWrite() // clost write half to signal EOF
+	response, err := r.ReadAndParseJSON()
+	if err != nil {
+		return "", err
+	}
+	r.Close()
```
squidboylan: Im not sure i understand why we're doing `r.Close()` here

tests/functional/lib/receptorcontrol/receptorcontrol.go
```diff
@@ -158,3 +165,152 @@ func (r *ReceptorControl) Status() (*netceptor.Status, error) {
...
+	if err != nil {
+		return "", err
+	}
+	r.ReadStr()
+	r.CloseWrite()
+	response, err := r.ReadAndParseJSON()
+	if err != nil {
+		return "", err
+	}
+	unitID := fmt.Sprintf("%v", response["unitid"])
```
squidboylan: We might consider returning the whole response here and updating `ReadAndParseJSON` to return `map[string]interface{}`

tests/functional/lib/receptorcontrol/receptorcontrol.go
```diff
@@ -158,3 +165,152 @@ func (r *ReceptorControl) Status() (*netceptor.Status, error) {
...
+		return true
+	default:
+		return false
+	}
+}
+
+//AssertWorkRunning waits until work status is running
+func (r *ReceptorControl) AssertWorkRunning(workID string) error {
+	timeout := time.After(6 * time.Second)
+	for {
```
squidboylan: We have a new `utils.CheckUntilTimeout` that would be good for this i think, then we could get rid of the local `checkTimeout`

tests/functional/lib/receptorcontrol/receptorcontrol.go
```diff
@@ -158,3 +165,152 @@ func (r *ReceptorControl) Status() (*netceptor.Status, error) {
...
+	select {
+	case <-timeout:
+		return true
+	default:
+		return false
+	}
+}
+
+//AssertWorkRunning waits until work status is running
+func (r *ReceptorControl) AssertWorkRunning(workID string) error {
```
squidboylan: We should also make the timeouts on these `Assert` functions configurable

tests/functional/lib/receptorcontrol/receptorcontrol.go
fosterseth: work submit expects stdin (until EOF) from the socket. A few lines above I close the Write side of the socket (`r.CloseWrite()`), which triggers the end of stdin.

But now that the controller write is closed, the controller is worthless for further use, so I just close the rest of the controller. 

I'd really like to find a better way to handle this, but I am not sure there is one.

tests/functional/lib/receptorcontrol/receptorcontrol.go
fosterseth: okay, either we can use a context with timeout and send that in, or we can just provide a timeout option as an argument into the Assert functions.

squidboylan: we might also consider an `Assert` and an `AssertWithTimeout` that just calls Assert in `CheckWithTimeout` so if you dont want to provide a context you dont have to

tests/functional/lib/receptorcontrol/receptorcontrol.go
squidboylan: ah i saw `WorkSubmit` closes the write side but not the whole socket so i assumed the other test was wrong, but it sounds like `WorkSubmit` is the one that was wrong

tests/functional/lib/receptorcontrol/receptorcontrol.go
squidboylan: > okay, either we can use a context with timeout and send that in, or we can just provide a timeout option as an argument into the Assert functions.

sorry just saw this, a context is more flexible so I think that's the right move

tests/functional/lib/receptorcontrol/receptorcontrol.go
ghjm: Strictly speaking, with a TCP socket, you only _can_ close the write side of the socket, by sending a FIN packet.  Whether the peer decides to send you more bytes is not under your control - though of course you can decide to ignore them, which is what a full close is.

The idea with work submission is that the submitter sends the stdin and then half-closes the socket, which lets the receiver know that this is the end of the data.  The receiver then sends a JSON response including the unit ID, and closes its side.  If you don't care about receiving the JSON, you can go ahead and fully close the socket, which just means the JSON response gets discarded.

tests/functional/lib/receptorcontrol/receptorcontrol.go
squidboylan: i think we also want to close the whole socket here after we get our response?

squidboylan (approved):
LGTM thanks

PR #122, opened 2020-08-27
Add Start, Shutdown CLINode functions

squidboylan:


squidboylan:
I'm pretty torn about how to do this, but I want to unblock the work that depends on this and I dont think it'll be too hard to improve upon this later so I'm happy to do this now, this only adds the restart functionality to CLINodes and not the Node interface, so in order to use it you'll need to do a type assertion

ghjm:
This is fine if we never intend this to work for other than CLI nodes.  If the goal is for it to also work for in-process nodes, then I think it's better to go ahead and add it to the Node interface now, and add stub functions that throw a "not implemented" error if someone calls it on a non-CLI node.

squidboylan:
ok, yeah i think we want this to work for both, im a little worried that it will be hard to get right for the in-process nodes (since it's not as easy as kill a process and run a command again), but I think it's something we should test

fosterseth (approved):

PR #130, opened 2020-09-01
Fix status file locking race condition

ghjm:
There is an unavoidable race condition (or at least I can't think of a way to avoid it) when creating and locking a new file using POSIX advisory file locking.  Because there is no atomic "create and lock" operation, it is possible for one thread to create the file and then find it already locked by another thread.  When this happens between a Workceptor status file save and load operation, the load sees an empty file, resulting in a JSON decoding error.

~~To fix this, this PR adds logic to the load operation to check whether the file we just opened is empty.  If we are experiencing the race condition, that means there is another thread immediately waiting to write to the file as soon as it gains the lock.  So we release the lock once, wait a short time, and then try to re-open the file.  If the file is still empty at this second attempt, we conclude that there really is an empty file in the directory, which is an error condition.~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,31 +2,34 @@ receptor: $(shell find pkg -type f -name '*.go') cmd/receptor.go
 	go build cmd/receptor.go
 
 lint:
-	golint cmd/... pkg/... example/...
+	@golint cmd/... pkg/... example/...
 
 format:
-	find cmd/ pkg/ -type f -name '*.go' -exec go fmt {} \;
+	@find cmd/ pkg/ -type f -name '*.go' -exec go fmt {} \;
 
 fmt: format
 
 pre-commit:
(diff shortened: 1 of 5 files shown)


fosterseth (approved):

squidboylan:
I'm still seeing some failures related to this race condition when i pull it down and run the tests. I saw the `TestWorkCancel` test and `TestStatusFileLock` fail

ghjm:
Ok, new plan: We will use a separate `status.lock` file for locking.  If the create/lock race results in a process creating this as a zero-length file but not obtaining the lock, it doesn't matter, because the file is expected to be zero-length anyway.  All status file operations, including creating it, occur while holding the lock on the status.lock file.

squidboylan (approved):

PR #131, opened 2020-09-01
Add more workceptor tests

fosterseth:
`TestWork` sets up a 3 node linear mesh
Node1  ---- Node2 ---- Node3   (node 1 can only reach node3 via node 2)

This PR includes 3 sub-tests
1. `Node1` starts work on `Node3`, then cancels and releases it.
2. Shutdown remote `Node3`. `Node1` starts work on `Node3`, bring `Node3` back online and make sure the job is successful on `Node1`.
3. `Node1` starts work on `Node3`. `Node2` goes down (the relay node), then comes back online. Make sure job on `Node1` is successful. It also tests that `Node1` receives partial work results before `Node2` goes down, then has the full results at the end.

Notes:
We have no way of indicating from the API that work has been successfully copied from stdout on the remote node to stdout on the controller node. 
`assertStdoutFizeSize()` is a method I wrote that will check the local stdout to make sure it has the data we expect, before calling `AssertWorkResults()`.

This change also adds a `Reconnect()` method for `receptorcontrol`. In some cases we needed to close the socket, so this allows us to reconnect without recreating a new receptorcontrol object

### DIFF
--- a/tests/functional/lib/receptorcontrol/receptorcontrol.go
+++ b/tests/functional/lib/receptorcontrol/receptorcontrol.go
@@ -254,8 +254,8 @@ func (r *ReceptorControl) WorkRelease(unitID string) error {
 }
 
 // GetWorkStatus returns JSON of status file for a given unitID
-func (r *ReceptorControl) GetWorkStatus(unitID string) (workceptor.StatusFileData, error) {
-	status := workceptor.StatusFileData{}
+func (r *ReceptorControl) GetWorkStatus(unitID string) (*workceptor.StatusFileData, error) {
+	status := &workceptor.StatusFileData{}
 	_, err := r.WriteStr(fmt.Sprintf("work status %s\n", unitID))
 	if err != nil {
 		return status, err
@@ -264,7 +264,7 @@ func (r *ReceptorControl) GetWorkStatus(unitID string) (workceptor.StatusFileDat
 	if err != nil {
 		return status, err
 	}
-	err = json.Unmarshal(jsonData, &status)
+	err = json.Unmarshal(jsonData, status)
 	if err != nil {
 		return status, err
 	}
--- a/tests/functional/lib/receptorcontrol/receptorcontrol.go
+++ b/tests/functional/lib/receptorcontrol/receptorcontrol.go
@@ -254,14 +254,19 @@ func (r *ReceptorControl) WorkRelease(unitID string) error {
 }
 
 // GetWorkStatus returns JSON of status file for a given unitID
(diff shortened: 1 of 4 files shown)


squidboylan:
This looks really cool, my main question is: is there a simple work status struct we can unmarshal the work status as, this would make our lives way easier and we'd get to avoid a lot of the type assertions we have to do currently.

Another thing is if we are concerned about the speed of the tests, we can split out the different subtests within the `TestWork` test and they can run in parallel, I'm not sure how much time that would actually save though.

ghjm:
You could do something like:
```
data := StatusFileData{}
data.ExtraData = make(map[string]interface{})
err = json.Unmarshal(bytes, &data)
```
With `ExtraData` already instantiated as a map, `json.Unmarshal` will fill it in with whatever's in the JSON.  You still have to do type assertions when you want to get ExtraData fields from a particular WorkType, but at least you can now say `data.Status` without having to assert its type every time.

fosterseth:
update: 
- `GetWorkStatus()` now returns our `workceptor.StatusFileData struct`
- the tests now run in parallel, which improves overall test speed

Still need to do assertions on `ExtraData` though, e.g. `remoteUnitID := workStatus.ExtraData.(map[string]interface{})["RemoteUnitID"].(string)`

@ghjm I tried to take your suggestion
```golang
data := StatusFileData{}
data.ExtraData = make(map[string]interface{})
err = json.Unmarshal(bytes, &data)
```
hoping that I could do `workStatus.ExtraData["remoteUnitID"].(string)` but that did not work (interface{} does not support indexing error)

tests/functional/lib/receptorcontrol/receptorcontrol.go
```diff
@@ -235,14 +253,20 @@ func (r *ReceptorControl) WorkRelease(workID string) error {
 	return nil
 }
 
-func (r *ReceptorControl) getWorkStatus(workID string) (map[string]interface{}, error) {
-	_, err := r.WriteStr(fmt.Sprintf("work status %s\n", workID))
+// GetWorkStatus returns JSON of status file for a given unitID
+func (r *ReceptorControl) GetWorkStatus(unitID string) (workceptor.StatusFileData, error) {
```
squidboylan: I think we probably want to return a `*workceptor.StatusFileData` here

squidboylan (changes requested):
I left one small inline comment, other than that it looks good

tests/functional/lib/receptorcontrol/receptorcontrol.go
fosterseth: done!

squidboylan (approved):

PR #138, opened 2020-09-03
Readstring uses context instead of timeout

fosterseth:
Also caller closes `conn` in case of context timeout.

### DIFF
--- a/pkg/utils/readstring_context.go
+++ b/pkg/utils/readstring_context.go
@@ -2,8 +2,8 @@ package utils
 
 import (
 	"bufio"
+	"context"
 	"fmt"
-	"time"
 )
 
 type readStringResult = struct {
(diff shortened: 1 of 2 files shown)


pkg/utils/readstring_context.go
```diff
@@ -27,7 +27,7 @@ func ReadStringWithTimeout(reader *bufio.Reader, delim byte, timeout time.Durati
 	select {
 	case res := <-result:
 		return res.str, res.err
-	case <-time.After(timeout):
-		return "", fmt.Errorf("timeout")
+	case <-ctx.Done():
+		return "", fmt.Errorf("ctx timeout")
```
ghjm: Maybe `return "", ctx.Err()`

ghjm (changes requested):

ghjm (approved):

pkg/utils/readstring_context.go
fosterseth: done!

Issue #137, opened 2020-09-03
Add an exponential backoff for redialing backends


squidboylan:
Currently we redial our backends on 5 second intervals, it would be good to redial quickly then have an exponential backoff, this might also help speed up some test times. See https://github.com/project-receptor/receptor/blob/devel/pkg/backends/utils.go#L13

PR #139, opened 2020-09-03
add delay backoff for redial in backends

fosterseth:
issue #137


squidboylan:
We should start our tcp, udp, websocket redials at a much quicker interval, maybe 100ms?

ghjm:
I see where that makes sense in the test context, but it seems a little aggressive in the real-world context, at least with a 1.5 backoff multiplier.  We'd be making 5 attempts within the first second.

ghjm:
Also, we have this same retry logic in multiple places in the code now.  Should we make a retry interval calculator in utils, and use it from all the places that are doing retries with backoffs?

fosterseth:
@ghjm 
I created a `util.IncreaseDuration()` function to handle increasing time.Duration up to a max. Is this what you had in mind?

ghjm:
I was vaguely thinking of a struct that holds the current, multiplier and limit values and has a NextDelay method, but it's not that big a deal.

fosterseth:
@ghjm ok I went ahead and implemented it as an `IncrementalDuration` struct with a `NextDelay()` method

ghjm (approved):

PR #140, opened 2020-09-04
Fix incremental duration to allow reset

fosterseth:
`dialerSession` needs the ability to reset the delay if a connection is successful.

Also, `nextDelay()` was called in the wrong place. I have verified that now it adds delay between retries, and that it resets if a connection is established.


pkg/backends/utils.go
```diff
@@ -57,11 +54,11 @@ func dialerSession(ctx context.Context, redial bool, redialDelay time.Duration,
 				}
 				select {
 				case <-time.After(redialDelayInc.Duration):
+					redialDelayInc.NextDelay()
```
ghjm: How about having the IncrementalDuration object have a method that returns a channel and increments the delay internally?  That way you can just do
```
case <-redialDelayInc.NextTimeout():
```
which might be implemented as something like:
```
func (ID *IncrementalDuration) NextTimeout() chan struct{} {
	ch := time.After(ID.Duration)
	ID.Duration = time.Duration(math.Min(ID.Multiplier*float64(ID.Duration), float64(ID.MaxDuration)))
	return ch
}
```
And maybe the struct values like `Duration` ought to be private?  I'm not sure what value there is for letting user code mess with our internal variables.

fosterseth: okay I've added a `NextTimeout()` that returns a `time.After` type of channel. To aid with the unit test, I retained a method to increase the duration, so that I can test the `maxDuration` without waiting on `NextTimeout`

ghjm (approved):

PR #142, opened 2020-09-04
Change readstring to readstringcontext to prevent possible hangs

fosterseth:
extending https://github.com/project-receptor/receptor/pull/138 to all the other uses of `readstring`


ghjm:
I think a 5 second timeout is too short for some of these.  In particular after we send the stdin, it might take a while to process and I think that's okay.  I was imagining that we would still have a local timeout for the hello message, but just pass the parent context directly into all the others.

fosterseth:
@ghjm I am now just passing in the parent context

ghjm (approved):

PR #143, opened 2020-09-04
Container packaging & cmdline singleton

ghjm:
This pull request does two things.  First, it updates the cmdline library to add the concept of a singleton.  Second, it adds a local RPM and container build process.  The cmdline changes are necessary to allow the Receptor container to set a default node ID based on its host name.

The cmdline changes are:
* `AddConfigType()` now takes an additional boolean parameter that sets the config type as a singleton.
* If a config type is a singleton, the second instantiation of this type will result in an error.
* If a config type is a _required_ singleton, then if it is not given on the command line / config file, it will be instantiated with default values.  (If there are required values with no default, this will error out.)

The change to Receptor itself is:
* If a node ID is not specified, the local hostname will be used rather than erroring out.

The packaging changes are:
* New `make rpms` target that build an RPM builder container and then builds the receptor and receptorctl RPMs in it.
* New `make container` target that builds a Receptor container with these RPMs installed in it.

The Makefile defaults to using podman, but you can run it with Docker by doing `CONTAINERCMD=docker make container`.


ghjm:
Note that these containers assume you're running as root, as you would with local Docker or local rootless Podman.  A future PR will need to do some kerjiggery with users to allow the Receptor container to run on Kubernetes in non-privileged mode.

fosterseth (approved):

ghjm:
Related: https://github.com/project-receptor/receptor/issues/14

squidboylan (approved):

PR #146, opened 2020-09-05
Command service improvements

ghjm:
* Combine the very similar `work start` and `work submit` commands into a single `work submit` command.
* Introduce a `localhost` name, so you can do `work submit localhost <worktype>` (and other things like `ping localhost`).
* Add the possibility of sending control service commands as JSON, to allow some commands to take larger or more complex parameters than are feasible as a simple command line (such as, for example, Kubernetes credentials).

### DIFF
--- a/cmd/receptor.go
+++ b/cmd/receptor.go
@@ -26,10 +26,18 @@ type nodeCfg struct {
 func (cfg nodeCfg) Prepare() error {
 	var err error
 	if cfg.ID == "" {
-		cfg.ID, err = os.Hostname()
+		host, err := os.Hostname()
 		if err != nil {
 			return err
 		}
+		lchost := strings.ToLower(host)
+		if lchost == "localhost" || lchost[:10] == "localhost." {
+			return fmt.Errorf("no node ID specified and local host name is localhost")
+		}
+		cfg.ID = host
+	}
+	if strings.ToLower(cfg.ID) == "localhost" {
+		return fmt.Errorf("node ID \"localhost\" is reserved")
 	}
 	var allowedPeers []string
 	if cfg.AllowedPeers != "" {
--- a/cmd/receptor.go
+++ b/cmd/receptor.go
@@ -23,7 +23,7 @@ type nodeCfg struct {
 	DataDir      string `description:"Directory in which to store node data"`
 }
 
-func (cfg nodeCfg) Prepare() error {
+func (cfg nodeCfg) Init() error {
 	var err error
 	if cfg.ID == "" {
 		host, err := os.Hostname()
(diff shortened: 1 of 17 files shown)


cmd/receptor.go
```diff
@@ -23,22 +23,33 @@ type nodeCfg struct {
 	DataDir      string `description:"Directory in which to store node data"`
 }
 
-func (cfg nodeCfg) Prepare() error {
+func (cfg nodeCfg) Init() error {
```
ghjm, line 26: The point of this change is to make sure that this function, which instantiates all the MainInstance global variables, gets called _first_, before any Prepare() functions that might try to make use of these variables.

pkg/netceptor/netceptor.go
```diff
@@ -538,11 +544,58 @@ func (s *Netceptor) flood(message []byte, excludeConn string) {
 	}
 }
 
+// GetServerTLSConfig retrieves a server TLS config by name
```
ghjm, line 547: The client and server TLS config lists were previously global in scope, not scoped within a Netceptor object.   This should have been fixed long ago when Netceptor was updated to allow multiple instances.

tests/functional/lib/receptorcontrol/receptorcontrol.go
```diff
@@ -377,8 +378,17 @@ func (r *ReceptorControl) AssertWorkReleased(ctx context.Context, unitID string)
...
-	r.ReadStr() // flush "Streaming results for.." line"
+	_, err := r.WriteStr(fmt.Sprintf("work results %s\n", unitID))
+	if err != nil {
+		return nil, err
+	}
+	str, err := r.ReadStr()
+	if err != nil {
+		return nil, err
+	}
+	if str[:5] == "ERROR" {
```
ghjm, line 389: Tests should always check error returns, particularly in obscure places where we don't expect them.  Not checking this particular error made the test hang (since the server side isn't going to send anything for the next `Read()` to read), and also made it a lot harder to figure out what the actual problem was in the tests.

I've added checks to several error returns in this file.  It might be worth looking through the rest of the tests to see if we have other cases like this.

pkg/netceptor/netceptor.go
squidboylan: Awesome, this was something i've been worried about handling in the test suite LibNode stuff

cmd/receptor.go
```diff
@@ -23,22 +23,33 @@ type nodeCfg struct {
...
 	var err error
 	if cfg.ID == "" {
-		cfg.ID, err = os.Hostname()
+		host, err := os.Hostname()
 		if err != nil {
 			return err
 		}
+		lchost := strings.ToLower(host)
+		if lchost == "localhost" || lchost[:10] == "localhost." {
```
squidboylan: should we use https://golang.org/pkg/strings/#HasPrefix here?

ghjm: Yeah, that's probably better. I'll update it.

squidboylan (approved):
Looks good to me, would be good to get @fosterseth 's thoughts too

tests/functional/lib/receptorcontrol/receptorcontrol.go
fosterseth: okay I'll comb through the workceptor tests and make sure we return errors

fosterseth (approved):

PR #149, opened 2020-09-07
Add HopCount field to Netceptor core protocol

ghjm:
This PR adds a HopCount field, to avoid unbounded bandwidth use when there is a routing loop (which the Receptor routing protocol _shouldn't_ allow, but...)   Related: https://github.com/project-receptor/receptor/issues/40.

This requires a change to the `MsgTypeData` packet format.  The initial, one-byte MessageType field is now extended to four bytes, with the first two being the message type and hop count.  Note that the change to the packet format is a breaking change - Receptor binaries compiled before this change will not be able to talk to those compiled after this change.  The 3rd and 4th bytes in the header are reserved for future uses.

Last but not least, this PR also reduces the time for batching routing update requests before actually doing the Dijkstra update.  On close examination of the code, we are never spuriously requesting routing updates - we only make such requests if the routing table has really changed, or a local connection has really come up or gone down.  Also, we have locking in place so if multiple requests are made within less than the time an update takes to complete, they will be serialized and the second update won't start until the first update is complete.  So it is safe to run these updates more aggressively.

### DIFF
--- a/pkg/netceptor/netceptor.go
+++ b/pkg/netceptor/netceptor.go
@@ -32,6 +32,9 @@ const ServiceAdTime = 60 * time.Second
 // SeenUpdateExpireTime is the age after which routing update IDs can be discarded
 const SeenUpdateExpireTime = 1 * time.Hour
 
+// MaxForwardingHops is the maximum number of times that Netceptor will forward a data packet
+const MaxForwardingHops = 30
+
 // MainInstance is the global instance of Netceptor instantiated by the command-line main() function
 var MainInstance *Netceptor
 
@@ -134,6 +137,7 @@ type messageData struct {
 	FromService string
 	ToNode      string
 	ToService   string
+	HopCount    byte
 	Data        []byte
 }
 
@@ -239,7 +243,7 @@ func New(ctx context.Context, NodeID string, AllowedPeers []string) *Netceptor {
 	s.clientTLSConfigs["default"] = &tls.Config{}
 	s.addNameHash(NodeID)
 	s.context, s.cancelFunc = context.WithCancel(ctx)
-	s.updateRoutingTableChan = tickrunner.Run(s.context, s.updateRoutingTable, time.Hour*24, time.Second*1)
+	s.updateRoutingTableChan = tickrunner.Run(s.context, s.updateRoutingTable, time.Hour*24, time.Millisecond*100)
 	s.sendRouteFloodChan = tickrunner.Run(s.context, s.sendRoutingUpdate, RouteUpdateTime, time.Millisecond*100)
 	s.sendServiceAdsChan = tickrunner.Run(s.context, s.sendServiceAds, ServiceAdTime, time.Second*5)
 	go s.monitorConnectionAging()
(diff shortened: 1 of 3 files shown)


pkg/netceptor/netceptor.go
```diff
@@ -31,6 +31,9 @@ const ServiceAdTime = 60 * time.Second
 // SeenUpdateExpireTime is the age after which routing update IDs can be discarded
 const SeenUpdateExpireTime = 1 * time.Hour
 
+// MaxForwardingHops is the maximum number of times that Netceptor will forward a data packet
+const MaxForwardingHops = 30
```
squidboylan, line 36: should this be configurable? We dont need that to merge this but I figure we should think about that

pkg/netceptor/netceptor.go
```diff
@@ -587,43 +591,48 @@ func fixedLenBytesFromString(s string, l int) []byte {
 
 // Translates an incoming message from wire protocol to messageData object.
 func (s *Netceptor) translateDataToMessage(data []byte) (*messageData, error) {
-	if len(data) < 33 {
+	if len(data) < 36 {
```
squidboylan, line 647: should we make a `MinimumPacketSize` const for this

pkg/netceptor/netceptor.go
```diff
@@ -587,43 +591,48 @@ func fixedLenBytesFromString(s string, l int) []byte {
...
 		return nil, fmt.Errorf("data too short to be a valid message")
 	}
-	fromNode, err := s.getNameFromHash(binary.BigEndian.Uint64(data[1:9]))
+	fromNode, err := s.getNameFromHash(binary.BigEndian.Uint64(data[4:12]))
```
squidboylan, line 650: might also be useful to have consts for these operations too

squidboylan (commented):
I left some inline comments.

Im also wondering if it might be better to keep track of the "HopsRemaining" and decrement that on each hop (it is my impression that this is how IP works but I havent looked too closely) instead of keeping track of the number of hops we've made and each node tracking the max number of hops it thinks a packet should be allowed.

pkg/netceptor/netceptor.go
ghjm: The two functions `translateDataToMessage` and `translateDataFromMessage` are the only places where these values are used.  I think the functions are considerably more readable with the values inline like this, rather than defined in constants far away from the code.  If the values needed to be used anywhere else, I would agree with you, but as things stand I think it's better to leave this as is.

ghjm:
> Im also wondering if it might be better to keep track of the "HopsRemaining" and decrement that on each hop (it is my impression that this is how IP works but I havent looked too closely) instead of keeping track of the number of hops we've made and each node tracking the max number of hops it thinks a packet should be allowed.

Yes, that's a good idea.  I'll update it to work that way.  That might also allow us to do a Receptor traceroute at some point in the future.

pkg/netceptor/netceptor.go
ghjm: Most of these constants probably ought to be configurable.

squidboylan (approved):

fosterseth (approved):

PR #155, opened 2020-09-09
Make RSA key generation at least 2048 bits to be FIPS compliant

fosterseth:


squidboylan (approved):

fosterseth:
We'll keep 1024 bit rsa keys for now, as they are faster to generate and keeps the test suite completion time lower. We can look into adding a FIPS mode to testing later.

PR #158, opened 2020-09-11
Add support for testing receptor in containers

squidboylan:


Makefile
```diff
@@ -53,6 +53,10 @@ container: rpms
 	@cp -av rpmbuild/RPMS/ packaging/container/RPMS/
 	@$(CONTAINERCMD) build packaging/container -t receptor
 
+tc-image: receptor
+	@cp receptor packaging/tc-image/
+	@docker build packaging/tc-image -t receptor-tc
```
ghjm: This should be `@$(CONTAINERCMD)` rather than `@docker`

Makefile
```diff
@@ -53,6 +53,10 @@ container: rpms
 	@cp -av rpmbuild/RPMS/ packaging/container/RPMS/
 	@$(CONTAINERCMD) build packaging/container -t receptor
 
+tc-image: receptor
```
ghjm: It seems to me like it would be better to do `tc-image: container` here, and then have the test container be a layer on top of a Makefile-built Receptor container.

packaging/tc-image/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM fedora:32
```
ghjm: See above comment; I think this should be `FROM: receptor:latest`

tests/functional/mesh/dockermesh.go
```diff
@@ -0,0 +1,611 @@
...
+func init() {
+	containerImage = os.Getenv("CONTAINER_IMAGE")
+	if containerImage == "" {
+		containerImage = "receptor-tc"
+	}
+}
+
+// DockerNode holds a Netceptor, this layer of abstraction might be unnecessary and
+// go away later
+type DockerNode struct {
```
ghjm: Should probably call this ContainerNode since it's not always Docker

tests/functional/mesh/dockermesh.go
```diff
@@ -0,0 +1,611 @@
...
+			fmt.Printf("applying tc rule, duplication: %s\n", n.TCRules.Duplication)
+			dockerArgs = append(dockerArgs, "duplication")
+			dockerArgs = append(dockerArgs, n.TCRules.Duplication)
+		}
+		if n.TCRules.Corrupt != "" {
+			fmt.Printf("applying tc rule, corrupt: %s\n", n.TCRules.Corrupt)
+			dockerArgs = append(dockerArgs, "corrupt")
+			dockerArgs = append(dockerArgs, n.TCRules.Corrupt)
+		}
+		tcCmd := exec.Command("docker", dockerArgs...)
```
ghjm: Let's make it possible to do this with podman as well as Docker

ghjm (changes requested):

ghjm:
I left some comments but overall this looks really cool.  How do you envision making use of it?  It seems like this is more something you would run manually when doing scale tests, rather than a normal test that becomes part of CI.  Maybe we should add a simple CLI that runs a mesh from a mesh definition file, or something like that?

packaging/tc-image/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM receptor:latest
+
+RUN dnf install tc -y
+
+ENTRYPOINT ["/bin/bash"]
```
ghjm, line 5: It probably doesn't matter for this use case, but in a Receptor container that's going to be running a busy Workceptor, we expect it to be exec-ing lots of worker processes, including processes that themselves exec other processes ... which is exactly the situation where we really want our pid 1 to be an init process that does proper zombie process reaping.  This is why I've got the entrypoint running dumb-init in the Receptor container.

squidboylan:
Ok this has been updated to support podman, however i'm still working on getting podman to work with `tc` and i expect i'll only be able to get it working with podman run by root if i can get it working at all due to podman's security policy

>  Maybe we should add a simple CLI that runs a mesh from a mesh definition file, or something like that?

This is a good idea, I'll start thinking about how something like this would be designed

packaging/tc-image/Dockerfile
squidboylan: the reason im using bash is because i want to redirect stdout and stderr, as far as i can tell i can't do that with `dumb-init`, if you know a way to do that with `dumb-init` i can switch to using that over bash

ghjm: I don't think there's a way.  We aren't likely to be running heavy Workceptor workload in this container though.

ghjm (approved):

ghjm:
Yes, agreed, this certainly isn't ever going to work with rootless Podman.  Seems like it should be possible to make it work with Podman-as-root though.

PR #162, opened 2020-09-13
Change tun_proxy to ip_router with IPv4/IPv6 routing

ghjm:
This change makes the Receptor tun_proxy, now called ip_router, aware of the IPv4 or IPv6 destinations of the packets it is forwarding, and able to choose a Receptor node based on routes received via the Netceptor service advertising mechanism.

The network name must be the same for all participating nodes.  Each node will listen on a local Receptor service using the network name, and will broadcast routes it can reach.  When an IP packet arrives at any node, it will forward it to the node advertising a route to that destination.  If there are multiple nodes advertising, the lowest Receptor path cost will be used.

Local addressing of the tun interface is always a /30 subnet, with the .1 and .2 addresses used as the local address and the point-to-point destination of the tun interface, respectively.  Additional subnets reachable via a node can be advertised using the routes parameter.

Here is a sample configuration for a simple three node network making use of this service:

```
- node:
    id: node1
- tcp-peer:
    address: node2:2222
- ip-router:
    networkname: nodenet
    localnet: 172.16.1.0/30
```

```
- node:
    id: node2
- tcp-listener:
    port: 2222
- ip-router:
    networkname: nodenet
    localnet: 172.16.1.4/30
```

```
- node:
    id: node3
- tcp-peer:
    address: node2:2222
- ip-router:
    networkname: nodenet
    localnet: 172.16.1.8/30
```

Once all three nodes are up and running, kernel routes will be added for all three subnets, and packets to remote subnets will be forwarded through the Receptor connection.  So users on all three nodes will be able to ping, ssh, etc node1 at 172.16.1.1, node 2 at 172.16.1.5 and node3 at 172.16.1.9.


ghjm:
@squidboylan I had a couple CI failures on this PR that seemed to be caused by the CLI node early shutdown checking reporting an error even though no node had actually shut down.  I made some changes to how the check is done:

- Move the select that receives from the fail channel outside the loop, since it only needs to receive once.
- Instead of a channel with a buffer size of 1, use non-blocking sends to the channel.  This way if multiple nodes fail, the 2nd and subsequent ones will just skip sending (since nobody is listening).
- Pass the CLI node causing the failure through the channel, so the error can say which node was the problem.

I don't know for sure whether this fixed anything or if I just coincidentally got a successful result after this change.  Either way, I wanted to call it out here to make sure you get a chance to look at it.

PR #165, opened 2020-09-15
Simplify packaging

ghjm:
* Don't use RPMs in the Receptor container - just copy the built files.  This saves the time-consuming RPM build when doing a container build.
* Calculate the version, release and officialness in the Makefile based on git tag and the VERSION file.
* Add prep/socketpair to RPM build in order to unbreak it.
* Bump VERSION to 0.9.1


fosterseth (approved):

squidboylan (approved):

ghjm:
Instead of storing a version number, generate it entirely from tags.  On official builds, which is to say versions built directly from a tagged commit, the version will just be the tag, and the RPM NVR will be the tag -1.  On unofficial builds, the version will be the most recent tag _plus one_ on the least significant version component (so if the latest tag is 0.9.0, the version will be 0.9.1), with a revision of `-0.datetime.commit`.

PR #169, opened 2020-09-16
Override labels in container image, to avoid getting the Fedora ones

ghjm:
The Receptor container is inheriting labels with the Fedora license, vendor etc.  We should override these with our own.

### DIFF
--- a/packaging/container/Dockerfile
+++ b/packaging/container/Dockerfile
@@ -1,6 +1,11 @@
 FROM fedora:32
 ARG VERSION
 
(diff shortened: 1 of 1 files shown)


packaging/container/Dockerfile
```diff
@@ -1,6 +1,11 @@
 FROM fedora:32
 ARG VERSION
 
+LABEL license="ASL2"
```
squidboylan, line 4: Is this the right license?

squidboylan (approved):

PR #171, opened 2020-09-16
Add arbitrary parameter passing to work units

ghjm:
* Move Params string field (representing command-line parameters) from the base StatusFileData to commandExtraData/kubeExtraData, since those are the only places where it is relevant
* Add a new params map[string]string to AllocateUnit/AllocateRemoteUnit, which represents generic and extensible parameters
* Have remoteWorkUnit pass arbitrary parameters to its remote peer, so it can be used with any current or future work type
* Clarify that work unit Init method is supposed to just allocate memory, and create a separate SetParamsAndSave for when we want to initialize a new unit and overwrite its status file

On the last bullet, the idea is that if you are creating a _new_ work unit, you do Init and then SetParamsAndSave.  If you are reloading an _existing_ work unit, you do Init and then Load.  Everything that writes to the unit is contained in SetParamsAndSave, so we don't get cases where loading a unit inadvertently changes something because we were only thinking about the new-unit case when we wrote the constructor code.

### DIFF
--- a/pkg/workceptor/command.go
+++ b/pkg/workceptor/command.go
@@ -20,14 +20,17 @@ import (
 // commandUnit implements the WorkUnit interface for the Receptor command worker plugin
 type commandUnit struct {
 	BaseWorkUnit
-	command    string
-	baseParams string
-	done       bool
+	command             string
+	baseParams          string
+	acceptRuntimeParams bool
+	cmdParams           string
+	done                bool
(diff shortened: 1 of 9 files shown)


ghjm:
Related https://github.com/project-receptor/receptor/issues/78

fosterseth (approved):

PR #172, opened 2020-09-17
Kubernetes worker improvements

ghjm:
* Combine `newWorker()` with `Init()` to provide the initialization behavior the k8s worker needs
* Add k8s auth from provided parameters as well as kubeconfig files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ else
 endif
 
 receptor: $(shell find pkg -type f -name '*.go') cmd/receptor.go
-	@go build -ldflags "-X main.version=$(APPVER)" $(TAGPARAM) cmd/receptor.go
+	go build -ldflags "-X main.version=$(APPVER)" $(TAGPARAM) cmd/receptor.go
 
 lint:
 	@golint cmd/... pkg/... example/...
--- a/pkg/workceptor/command.go
+++ b/pkg/workceptor/command.go
@@ -145,7 +145,7 @@ func (cw *commandUnit) SetFromParams(params map[string]string) error {
 	if !ok {
 		cmdParams = ""
 	}
-	if ok && !cw.allowRuntimeParams {
+	if cmdParams != "" && !cw.allowRuntimeParams {
 		return fmt.Errorf("extra params provided but not allowed")
 	}
 	var allParams string
@@ -162,6 +162,11 @@ func (cw *commandUnit) SetFromParams(params map[string]string) error {
 
 // Status returns a copy of the status currently loaded in memory
 func (cw *commandUnit) Status() *StatusFileData {
+	return cw.UnredactedStatus()
+}
+
+// UnredactedStatus returns a copy of the status currently loaded in memory, including secrets
+func (cw *commandUnit) UnredactedStatus() *StatusFileData {
 	cw.statusLock.RLock()
 	defer cw.statusLock.RUnlock()
 	status := cw.getStatus()
--- a/pkg/workceptor/command.go
+++ b/pkg/workceptor/command.go
@@ -139,9 +139,9 @@ loop:
 	return nil
 }
 
-// SetParams sets the unit's in-memory status from command line parameters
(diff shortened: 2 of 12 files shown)


ghjm:
There's still more work to be done on this, but I think this is a good point where we can merge the work so far.

What this includes:

* We have the concept of secret fields.  Units now have a new `UnredactedStatus()` function.  `Status()` now returns status data with any secrets blanked out.  `UnredactedStatus()` returns the full data including secrets.
* The command service `work submit` function will return an error if you try to submit a command including JSON parameters that start with `secret_`, but you are proposing to send them over a non-TLS connection to a remote node.
* The remote worker's `work submit` implementation now always uses JSON commands to start the work on the remote node.
* The `SetParamsAndSave` function now only sets params, to allow decoupling of the param setting from the saving.  Also, it is now named `SetFromParams` to better reflect what's happening (the object is being set from the provided params).
* Receptorctl is updated to include the ability to provide arbitrary parameters to a work submittal via `--param/-a`.

Still to be done (outside the scope of this PR):

* Add a mode where the worker pod communicates with Receptor by cluster-local TCP instead of through the slow and problematic k8s spdy/logger implementation.
* Encryption at rest of status/stdout data.

fosterseth (approved):

squidboylan (approved):

fosterseth:
Tried this out with a local minikube setup

everything mostly worked fine, except I got a hang when trying the following (I see the pod is created and running, but it never terminates)

```yaml
- work-kubernetes:
    worktype: kubeit
    namespace: default
    authmethod: params
    kubehost: http://127.0.0.1:44355/
    kubeapipath: api/v1
    kubeusername: sbf
    kubepassword: sbf
    allowruntimecommand: true
```
`seq 10 | receptorctl --socket /tmp/foo.sock work submit kubeit --payload - -f --param kube_image=centos:8 --rm`

if I add in my own `kube_command` it will work

`seq 10 | receptorctl --socket /tmp/foo.sock work submit kubeit --payload - -f --param kube_image=centos:8 --rm --param kube_command='echo HI'`

also if I change `authmethod` to `kubeconfig` it will work

```yaml
- work-kubernetes:
    worktype: kubeit
    namespace: default
    kubeconfig: /home/sfoster/.kube/config
    authmethod: kubeconfig
    allowruntimecommand: true
```

Maybe I am missing some detail?

ghjm:
I wasn't able to reproduce this - in my environment the first example works.  But I am using CRC rather than minikube.

I have seen a couple times when we didn't get any results until the whole process finished, then we get them all at once at the end.  The delay seemed to be happening in the k8s logger, not in Receptor.  Is there any chance this is what's going on for you - i.e. did you wait long enough that the whole process could have finished?  Also, when you see the container running, does it go to a finished state or does it stay in running?

fosterseth:
staying in running state

```
[sfoster@localhost ghjm]$ kubectl get pods
NAME           READY   STATUS    RESTARTS   AGE
kubeit-tvp6s   1/1     Running   0          3m58s
```
from describe

```
Events:
  Type    Reason     Age    From               Message
  ----    ------     ----   ----               -------
  Normal  Scheduled  4m30s  default-scheduler  Successfully assigned default/kubeit-tvp6s to minikube
  Normal  Pulled     4m29s  kubelet, minikube  Container image "centos:8" already present on machine
  Normal  Created    4m29s  kubelet, minikube  Created container worker
  Normal  Started    4m29s  kubelet, minikube  Started container worker
```

ghjm:
I wonder if it's not getting stdin?

fosterseth:
hmm you might be right, given my echo command works fine in those cases

ghjm:
So then the question is whether Receptor is failing to pass the stdin, or if it's getting to the Kube API but not making it to the actual container.

ghjm:
Merging this as is for now, will deal with the stdin issue (if that's what it is) in a future PR

PR #174, opened 2020-09-17
Add direct support for Gorilla websockets in ExternalBackend

ghjm:
This PR alters ExternalBackend so that instead of taking a `net.Conn`, it takes a new interface called a `MessageConn`, which can be a wrapper around either `net.Conn` or Gorilla `websocket.Conn`.  Talking directly to `websocket.Conn` allows us to avoid the use of `Framer` when it isn't required.

So if you have existing code:
```
var eb ExternalBackend
var conn net.Conn

...

eb.NewConnection(conn, true)
```

it will need to be changed to:

```
eb.NewConnection(MessageConnFromNetConn(conn), true)
```

and if you have a `websockets.Conn`, you can now do:

```
eb.NewConnection(MessageConnFromWebsocketConn(wsconn), true)
```

### DIFF
--- a/pkg/netceptor/external_backend.go
+++ b/pkg/netceptor/external_backend.go
@@ -3,6 +3,7 @@ package netceptor
 import (
 	"context"
 	"fmt"
+	"github.com/gorilla/websocket"
 	"github.com/project-receptor/receptor/pkg/framer"
 	"net"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/netceptor/external_backend.go
```diff
@@ -16,6 +17,128 @@ type ExternalBackend struct {
...
+		conn: conn,
+	}
+}
+
+// WriteMessage writes a message to the connection
+func (mc *websocketMessageConn) WriteMessage(ctx context.Context, data []byte) error {
+	if ctx.Err() != nil {
+		return fmt.Errorf("session closed: %s", ctx.Err())
+	}
+	return mc.conn.WriteMessage(websocket.BinaryMessage, data)
```
jhjaggars, line 117: Is there a use-case where you might want to be sending text messages?

jhjaggars (approved):

pkg/netceptor/external_backend.go
ghjm: No, because this is a backend connection, so the only thing Receptor will ever send over it will be Receptor wire protocol packets, which are binary data structures.

PR #178, opened 2020-09-18
Submit work to remote has option to provide a tlsclient config

fosterseth:
Works with `InitFromJson` by providing a `tlsclient` option

e.g.

```json
{"command":"work","subcommand":"submit","worktype":"100cat","tlsclient":"client","node":"bar","params":""}
```

Therefore, I changed `receptorctl work submit` to use JSON encoded strings when writing to the control socket.

**Behavior if wrong TLS configuration is provided**
If remote job has not yet started, we fail the job and add `Incorrect tlsclient to remote service` message to the status Detail field.
If remote job has already started (maybe tls server was reconfigured), we do not fail the job but still provide the same Detail message.

Here are the logger Warning messages that appear when providing incorrect tls information
_client provided no tlsclient_
`WARNING Connection to bar failed with error: CRYPTO_ERROR: insecure connection to secure service`

_client provided wrong ca_
`WARNING Connection to bar failed with error: CRYPTO_ERROR: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "Easy-RSA CA")`

_client provided wrong certs and keys_
`WARNING Connection to bar failed with error: CRYPTO_ERROR: tls: bad certificate`

**Other changes**,

`LocalStarted` in `remoteExtraData` renamed to `RemoteStarted`

If `RemoteStarted` is false, we can do a `release` with force, since nothing has actually started on the remote. Otherwise if we are having issues connecting to remote, we must do a `force-release` even though we know nothing has started there.

added a `made version` target which saves the .VERSION files

**To do in other PRs**:
provide tls expectations in service advertisement
add a test case that provides tlsclient

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,6 @@ ci: pre-commit build-all test
 
 version:
 	@echo $(VERSION) > .VERSION
-	@echo $(VERSION) > receptorctl/.VERSION
 	@echo ".VERSION created for $(VERSION)"
 
 SPECFILES = packaging/rpm/receptor.spec packaging/rpm/receptorctl.spec packaging/rpm/receptor-python-worker.spec
--- a/Makefile
+++ b/Makefile
@@ -86,6 +86,11 @@ testloop: receptor
 ci: pre-commit build-all test
 	@echo "All done"
 
+version:
+	@echo $(VERSION) > .VERSION
+	@echo $(VERSION) > receptorctl/.VERSION
+	@echo ".VERSION created for $(VERSION)"
+
 SPECFILES = packaging/rpm/receptor.spec packaging/rpm/receptorctl.spec packaging/rpm/receptor-python-worker.spec
 
 specfiles: $(SPECFILES)
--- a/pkg/workceptor/remote_work.go
+++ b/pkg/workceptor/remote_work.go
@@ -34,6 +34,7 @@ type remoteExtraData struct {
 	LocalStarted   bool
 	LocalCancelled bool
 	LocalReleased  bool
+	TLSConfigName  string
 }
 
 type actionFunc func(context.Context, net.Conn, *bufio.Reader) error
@@ -48,7 +49,13 @@ func (rw *remoteUnit) connectToRemote(ctx context.Context) (net.Conn, *bufio.Rea
 	}
 	node := red.RemoteNode
 	rw.statusLock.RUnlock()
-	conn, err := rw.w.nc.DialContext(ctx, node, "control", nil)
+	tlsConfigName := rw.Status().ExtraData.(*remoteExtraData).TLSConfigName
+	expectedHostName := rw.Status().ExtraData.(*remoteExtraData).RemoteNode
+	tlsConfig, err := rw.w.nc.GetClientTLSConfig(tlsConfigName, expectedHostName)
+	if err != nil {
+		return nil, nil, err
+	}
+	conn, err := rw.w.nc.DialContext(ctx, node, "control", tlsConfig)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/pkg/workceptor/remote_work.go
+++ b/pkg/workceptor/remote_work.go
@@ -89,17 +89,17 @@ func (rw *remoteUnit) connectToRemote(ctx context.Context) (net.Conn, *bufio.Rea
 }
 
 // getConnection retries connectToRemote until connected or the context expires
-func (rw *remoteUnit) getConnection(mw *utils.JobContext) (net.Conn, *bufio.Reader) {
+func (rw *remoteUnit) getConnection(ctx context.Context) (net.Conn, *bufio.Reader) {
 	connectDelay := utils.NewIncrementalDuration(SuccessWorkSleep, MaxWorkSleep, 1.5)
 	for {
-		conn, reader, err := rw.connectToRemote(mw)
+		conn, reader, err := rw.connectToRemote(ctx)
 		if err == nil {
 			return conn, reader
 		}
 		logger.Warning("Connection to %s failed with error: %s",
 			rw.Status().ExtraData.(*remoteExtraData).RemoteNode, err)
 		select {
-		case <-mw.Done():
+		case <-ctx.Done():
 			return nil, nil
 		case <-connectDelay.NextTimeout():
 		}
@@ -121,9 +121,9 @@ func (rw *remoteUnit) connectAndRun(ctx context.Context, action actionFunc) erro
 // getConnectionAndRun retries connecting to a host and, once the connection succeeds, runs an action function.
 // If firstTimeSync is true, a single attempt is made on the calling goroutine. If the initial attempt fails to
 // connect or firstTimeSync is false, we run return ErrPending to the caller.
-func (rw *remoteUnit) getConnectionAndRun(mw *utils.JobContext, firstTimeSync bool, action actionFunc) error {
+func (rw *remoteUnit) getConnectionAndRun(ctx context.Context, firstTimeSync bool, action actionFunc) error {
 	if firstTimeSync {
-		err := rw.connectAndRun(mw, action)
+		err := rw.connectAndRun(ctx, action)
 		if err == nil {
 			return nil
 		} else if !utils.ErrorIsKind(err, "connection") {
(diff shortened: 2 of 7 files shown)


pkg/workceptor/remote_work.go
```diff
@@ -68,17 +75,34 @@ func (rw *remoteUnit) connectToRemote(ctx context.Context) (net.Conn, *bufio.Rea
...
+			if status.Detail != detailMsg {
+				rw.UpdateFullStatus(func(status *StatusFileData) {
+					status.Detail = detailMsg
+				})
+			}
+			if status.ExtraData.(*remoteExtraData).RemoteStarted == false {
+				rw.UpdateFullStatus(func(status *StatusFileData) {
+					status.State = WorkStateFailed
+				})
+				mw.WorkerDone()
```
fosterseth: not calling `WorkDone()` here causes subsequent `release` or `cancel` commands on this work unit to hang.

pkg/workceptor/remote_work.go
```diff
@@ -68,17 +75,34 @@ func (rw *remoteUnit) connectToRemote(ctx context.Context) (net.Conn, *bufio.Rea
...
 			return conn, reader
 		}
-		logger.Debug("Connection to %s failed with error: %s",
-			rw.Status().ExtraData.(*remoteExtraData).RemoteNode, err)
+		status := rw.Status()
+		logger.Warning("Connection to %s failed with error: %s",
+			status.ExtraData.(*remoteExtraData).RemoteNode, err)
+		errStr := err.Error()
+		if strings.Contains(errStr, "CRYPTO_ERROR") {
+			detailMsg := "Incorrect tlsclient to remote service"
```
ghjm: I think it would be a good idea to use fmt.Sprintf here to insert errStr into detailMsg.  Wherever possible, we want to pass error information up rather than mask it, to make life easier for anyone troubleshooting this in the future.

pkg/workceptor/remote_work.go
```diff
@@ -68,17 +75,34 @@ func (rw *remoteUnit) connectToRemote(ctx context.Context) (net.Conn, *bufio.Rea
...
+			if status.Detail != detailMsg {
```
ghjm: The semantics of UpdateFullStatus are that it gets a lock on the status file, reads its current content, then runs the update function you pass in.  By checking `status.Detail` _outside_ the UpdateFullStatus block, it's possible that someone else was ahead of us in line for the lock, and updated the file before UpdateFullStatus began.  So I would do this as something more like
```
	if strings.Contains(errStr, "CRYPTO_ERROR") {
		var shouldExit = false
			status.Detail = fmt.Sprintf("TLS error connecting to remote service: %s", errStr)
			if status.ExtraData.(*remoteExtraData).RemoteStarted == false {
				shouldExit = true
				status.State = WorkStateFailed
			}
		})
		if shouldExit {
			mw.Cancel()
			return
		}
	}
```

pkg/workceptor/remote_work.go
ghjm: I think the problem here is that you ought to `return` - as things stand you're calling `WorkerDone` but then continuing around the for loop, which is blocking.  If you _did_ return then I think the outer function would take care of calling `WorkerDone` for you.  It's also super important to make sure that `WorkerDone` only gets called _once_ per dispatched worker goroutine, which I think won't be the case here (since, whenever this function _does_ manage to exit the loop, it will return to its caller, who has a deferred WorkerDone ready to go).

pkg/workceptor/remote_work.go
ghjm: To put this another way, everything inside the UpdateFullStatus call is guaranteed(*) to be atomic.

(*) if I wrote the locking code right

pkg/workceptor/remote_work.go
```diff
@@ -113,7 +111,7 @@ func (rw *remoteUnit) getConnection(mw *utils.JobContext) (net.Conn, *bufio.Read
 func (rw *remoteUnit) connectAndRun(ctx context.Context, action actionFunc) error {
 	conn, reader, err := rw.connectToRemote(ctx)
 	if err != nil {
-		return utils.WrapErrorWithKind(err, "connection")
+		return utils.WrapErrorWithKind(err, "connection first attempt")
```
ghjm: What's going on here?

fosterseth: accidentally left this in, I'll fix in my next commit

fosterseth:
@ghjm I had to rewrite this a bit

I shifted the logic that checks for CRYPTO errors into `connectToRemote` instead

The previous solution we came up with wasn't going to work -- `getConnection` runs in a go-routine so it never returns to our `runAndMonitor` function where we do the `WorkerDone` calls.

However, `connectAndRun` does provide immediate errors back to `runAndMonitor` so if we hit a crypto error we can just call `workerDone()` there. I think this better preserves the original intent of getConnection and connectandRun anyways -- the former will keep attempting, while the latter does a one-time attempt.

When we start job we always do a first one time attempt, so if we hit tls errors we still fail the job. In the case where tls errors come after the job starts, we spin in the getConnection loop as expected, until tls configuration is fixed.

we need to wrap the error with "crypto", because if it is wrapped with "connection" we eventually return an ErrPending and don't call `WorkerDone()`

pkg/workceptor/remote_work.go
fosterseth: this is fixed up now

now all WorkerDone() calls for the job are in a single method, `runAndMonitor`

pkg/workceptor/remote_work.go
fosterseth: fixed

fosterseth:
I found some issues with the implementation, so my previous comment no longer applies.

pkg/workceptor/remote_work.go
```diff
@@ -97,19 +119,21 @@ func (rw *remoteUnit) connectAndRun(ctx context.Context, action actionFunc) erro
...
 		}
 	}
 	go func() {
-		conn, reader := rw.getConnection(ctx)
+		conn, reader := rw.getConnection(mw)
 		if conn != nil {
-			_ = action(ctx, conn, reader)
+			_ = action(mw, conn, reader)
+		} else {
+			mw.WorkerDone()
```
fosterseth: action() already calls WorkerDone() in the right places, so we need to call WorkerDone() only if action() isn't executed

ghjm: This can't be right ... action is of type actionFunc:
```
type actionFunc func(context.Context, net.Conn, *bufio.Reader) error
```
which takes a Context, not a JobContext, so it doesn't have a WorkerDone function to call.

fosterseth: it appears it grabs the mw out of scope?

https://github.com/project-receptor/receptor/blob/e862c28921c3edd780ffbcfb2a0fd0dccac5e09c/pkg/workceptor/remote_work.go#L407

Makefile
```diff
@@ -86,6 +86,11 @@ testloop: receptor
 ci: pre-commit build-all test
 	@echo "All done"
 
+version:
+	@echo $(VERSION) > .VERSION
+	@echo $(VERSION) > receptorctl/.VERSION
```
ghjm: It should not be necessary to create a copy in the receptorctl directory, because receptorctl's setup.py also looks for `../.VERSION`.

pkg/workceptor/remote_work.go
```diff
@@ -68,17 +75,32 @@ func (rw *remoteUnit) connectToRemote(ctx context.Context) (net.Conn, *bufio.Rea
 }
 
 // getConnection retries connectToRemote until connected or the context expires
-func (rw *remoteUnit) getConnection(ctx context.Context) (net.Conn, *bufio.Reader) {
+func (rw *remoteUnit) getConnection(mw *utils.JobContext) (net.Conn, *bufio.Reader) {
```
ghjm: I don't see where we're making use of any JobContext functionality, so I think this parameter should still be a `context.Context`

pkg/workceptor/remote_work.go
ghjm: That's being called from the outer function, not the action passed in.

pkg/workceptor/remote_work.go
fosterseth: we call `mw.Cancel()` in this method, line 98

ghjm (approved):

PR #179, opened 2020-09-18
Add netceptor.SubscribeRoutingUpdates() function

ghjm:
This adds a `SubscribeRoutingUpdates()` function to Netceptor, which allows API users to be proactively notified when the routing table changes.


ghjm:
This PR has gone somewhat beyond its original intent.  While adding `SubscribeRoutingUpdates()`, it was necessary to rework the broker code somewhat, so I took the opportunity to pass a proper struct instead of a string map for unreachable messages.  While trying to troubleshoot CI failures, I also made some changes related to protecting against blocking behavior.

dehort (approved):
Looks like I can get routing table updates from netceptor with this change.  Looks good to me

fosterseth (approved):

PR #183, opened 2020-09-22
Test server tls verification by clients

squidboylan:


ghjm:
While you're looking at this, do you think you can slip in some client cert generation and verification?  😀

Edit: Never mind, you already did!

ghjm (approved):

PR #184, opened 2020-09-23
Add tests for submitting work via JSON with a tlsclient

fosterseth:
Recent API changes allow us to submit work with JSON

```golang
command := `{"command":"work","subcommand":"submit","worktype":"echosleepshort","tlsclient":"","node":"node2","params":""}`
unitID, err := controllers["node3"].WorkSubmitJSON(command)
```

This API supports adding a tlsclient that will be used when connecting to a remote control service that has TLS configured

**Other changes:**

You can now generate a certificate with a specified Common Name, which allows us to test with `insecureskipverify: false` at the receptor level.

You can now setup a control-service with service `control` in the node definitions. This allows us to set up control services with tls enabled.

```golang 
map[interface{}]interface{}{
	"control-service": map[interface{}]interface{}{
		"service":  "control",
		"tls":      "control_tls",
	},
},
```

When generating the mesh, it will skip over creating a control socket if the node already has a control-service with the `control` service defined.

It will however generate a control socket filename and append it to the definition.

### DIFF
--- a/tests/functional/lib/mesh/climesh.go
+++ b/tests/functional/lib/mesh/climesh.go
@@ -362,9 +362,9 @@ func NewCLIMeshFromYaml(MeshDefinition YamlData, dirPrefix string) (*CLIMesh, er
 		}
 		controlSocket := filepath.Join(tempdir, "controlsock")
 		node.controlSocket = controlSocket
-		tmp := make(map[interface{}]interface{})
-		tmp["filename"] = controlSocket
 		if needsControlService {
+			tmp := make(map[interface{}]interface{})
+			tmp["filename"] = controlSocket
 			controlServiceYaml := make(map[interface{}]interface{})
 			controlServiceYaml["control-service"] = tmp
 			MeshDefinition.Nodes[k].Nodedef = append(MeshDefinition.Nodes[k].Nodedef, controlServiceYaml)
--- a/tests/functional/lib/mesh/climesh.go
+++ b/tests/functional/lib/mesh/climesh.go
@@ -334,6 +334,28 @@ func NewCLIMeshFromYaml(MeshDefinition YamlData, dirPrefix string) (*CLIMesh, er
 
 	// Setup the controlsvc and sockets
 	for k, node := range nodes {
+		controlSocket := ""
+		needsControlService := true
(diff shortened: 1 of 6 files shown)


squidboylan:
It seems like in most/all cases a generated filename is good enough but you might still want to specify tls for the control service. Because of this, I think we should allow for using TLS on a control service with a generated filename, it doesnt look possible to do this at first glance.

We should also consider whether we want to allow specifying the control service filename at all, that could complicate things long term, for example with docker nodes the control service needs to be exposed inside and outside the container, if we always generate the filename for the node we can make assumptions about how to do that, if we let a node specify the filename that gets a little trickier.

fosterseth:
@squidboylan 

okay now the mesh creation process will generate a filename for the control-service. If the user specifies a filename in the definition, it will throw an error.

tests/functional/lib/mesh/climesh.go
```diff
@@ -345,27 +345,32 @@ func NewCLIMeshFromYaml(MeshDefinition YamlData, dirPrefix string) (*CLIMesh, er
...
 		}
-		node.controlSocket = filepath.Join(tempdir, "controlsock")
-		controlServiceYaml := make(map[interface{}]interface{})
+		controlSocket := filepath.Join(tempdir, "controlsock")
+		node.controlSocket = controlSocket
 		tmp := make(map[interface{}]interface{})
-		tmp["filename"] = node.controlSocket
-		controlServiceYaml["control-service"] = tmp
-		MeshDefinition.Nodes[k].Nodedef = append(MeshDefinition.Nodes[k].Nodedef, controlServiceYaml)
+		tmp["filename"] = controlSocket
```
squidboylan: this `tmp` variable can go in the `if` now (as well as line 365 where we create it), makes it a bit more clear that that's only used when we are creating the whole controlservice definition.

squidboylan:
I left one small inline comment but overall this looks right, thanks!

tests/functional/lib/mesh/climesh.go
fosterseth: fixed!

squidboylan (approved):
LGTM

ghjm (approved):

PR #196, opened 2020-10-01
Fixes some flake in TestHopCountLimit

squidboylan:
There is a race condition in SubscribeRoutingUpdates where we may miss
the first routing update because the channel is unbuffered and the send
is non-blocking, blocks the send so it is less likely we will miss any
routing updates.


ghjm (approved):
I'm OK with this.  It means less protection against a misbehaved client that takes the channel but fails to listen to it.  But there are a lot of other ways you can misuse the library to produce bad results, and I don't think this particular one needs this much special treatment.

PR #199, opened 2020-10-02
Kubernetes TCP streaming

ghjm:
This PR adds a configurable ability to stream to/from a worker pod using cluster-local TCP instead of the SPDY streaming built into Kubernetes, which is a bit janky and slow.  To use this, run a Receptor container within Kubernetes, configure its service account to be able to create pods, and give it a configuration like:

```
- work-kubernetes:
    worktype: tcpcat
    image: alpine/socat
    command: sh -c "nc $RECEPTOR_HOST $RECEPTOR_PORT -e awk '{system(\"echo \"$1\"; sleep 0.5\")}'"
    authmethod: incluster
    streammethod: tcp
    namespace: test
```

Receptor creates the worker pod and sets the RECEPTOR_HOST and RECEPTOR_PORT environment variables.  The pod is then expected to make a TCP connection back to this host and port, and then receive its stdin and produce its stdout from/to the TCP stream socket.

This PR also includes several improvements that were necessary for writing or testing this feature:
* Change the order of operations in the Kubernetes logger streaming method to avoid a case where the log fails to open
* Add the ability for a control socket to listen directly on a TCP port
* Add the ability for receptorctl to connect to a control socket via TCP (which I apparently just forgot to write earlier)
* When someone gives an integer value for a string field in a YAML config, do the right thing instead of passing a weird string representation of the binary value of the integer (in case you do `tcplisten: 12345`)
* Move the `monitorLocalStatus()` function from the command work unit to the base work unit, so the Kubernetes worker can use it

### DIFF
--- a/pkg/cmdline/cmdline.go
+++ b/pkg/cmdline/cmdline.go
@@ -235,7 +235,7 @@ func setValue(field *reflect.Value, value interface{}) error {
 	valueType := reflect.TypeOf(value)
 
 	// If the value is directly convertible to the field, just set it
-	if valueType.ConvertibleTo(fieldType) {
+	if fieldKind == valueType.Kind() && valueType.ConvertibleTo(fieldType) {
 		field.Set(reflect.ValueOf(value).Convert(fieldType))
 		return nil
 	}
@@ -318,6 +318,13 @@ func setValue(field *reflect.Value, value interface{}) error {
 			return nil
 		}
 	}
+
+	// If the destination is a string, try a last ditch string conversion
+	if fieldKind == reflect.String {
+		field.Set(reflect.ValueOf(fmt.Sprintf("%v", value)))
+		return nil
+	}
+
 	return fmt.Errorf("type error (expected %s)", fieldType)
 }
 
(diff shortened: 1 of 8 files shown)


fosterseth (approved):

fosterseth:
played around with it, and TCP streaming worked great. Also tried TCP for the control service, which also worked as expected

I did experience a scenario where a container failed, but `work status` said the job finished and succeeded.

```yaml
- work-kubernetes:
    worktype: kubeit
    image: alpine/socat
    namespace: default
    kubeconfig: /home/sfoster/.kube/config
    authmethod: kubeconfig
```
the image doesn't have bash, I can see the pod error `Error: failed to start container "worker": Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"bash\": executable file not found in $PATH": unknown`

However the work results didn't reflect that error

```
{'iyU524yu': {'Detail': 'Finished',
              'ExtraData': {'Command': 'bash -c "while read -r line; do echo '
                                       '$line; sleep 0.5; done"',
                            'Image': 'alpine/socat',
                            'KubeAPIPath': '',
                            'KubeBearerToken': '',
                            'KubeHost': '',
                            'KubeNamespace': 'default',
                            'KubePassword': '',
                            'KubeTLSCAData': '',
                            'KubeUsername': '',
                            'KubeVerifyTLS': True,
                            'Params': '',
                            'PodName': 'kubeit-bfp8m'},
              'State': 2,
              'StateName': 'Succeeded',
              'StdoutSize': 0,
              'WorkType': 'kubeit'}}
```

Issue #185, opened 2020-09-24
Check CN in receptor-level client certificates


ghjm:
In Receptor-level mutual TLS, the listener node establishes the legitimacy of its claim to a node ID by supplying a certificate, signed by a CA the client trusts, with a CN matching the node ID.  However, when the peer (dialer) node provides a client certificate, we only check that it is signed by a client-signing CA trusted by the listener. We do _not_ check that the client cert's CN matches the node ID claimed by the peer.

We should check this. The Receptor-level security model isn't that we need a valid user - it's that we need to verify the specific claim of being a given node ID, and we ought to verify it in both directions.

This is a bit different than how web server mutual TLS usually works, so implementing this might (or might not!) require writing a custom certificate verification function.

squidboylan:
@fosterseth added testing for this as part of his #201 PR, I believe it is enough for us to consider this done

PR #201, opened 2020-10-02
Allow tls-server to verify that client node id matches certificate CN

fosterseth:
related #185 

adds the following option to `--tls-server`,
`verifyclientnodeid=<bool>: Verify certificate CA matches client node id (default: true)`

If true, the server will check that the connecting client node id matches the Common Name in the client certificate.

It does this through the `tls.Config` callback, `VerifyPeerCertificate`. This callback will be called after the regular tls verification takes place.

In order to know which client is connecting, we must use yet another callback called `GetConfigForClient` which is called during the handshake. This callback sends in `net.Conn` information, where we can extract the connecting `RemoteAddr`.

I've updated two tests in `TestWork` which ensures the verification works when a correct tls client cert is provided, and that a failure occurs when providing a tls client cert with the wrong CN.

### DIFF
--- a/pkg/netceptor/tlsconfig.go
+++ b/pkg/netceptor/tlsconfig.go
@@ -80,6 +80,8 @@ func (cfg TLSServerCfg) Prepare() error {
 	}
 
 	if cfg.VerifyClientNodeID {
+		// make GetConfigForClient non-nil for now, and later fill in the callback
+		// after cloning the tls
 		tlscfg.GetConfigForClient = func(hi *tls.ClientHelloInfo) (*tls.Config, error) {
 			return nil, nil
 		}
--- a/pkg/netceptor/tlsconfig.go
+++ b/pkg/netceptor/tlsconfig.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"github.com/project-receptor/receptor/pkg/cmdline"
 	"io/ioutil"
+	"strings"
 )
 
 // **************************************************************************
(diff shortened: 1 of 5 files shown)


pkg/netceptor/tlsconfig.go
```diff
@@ -65,7 +79,16 @@ func (cfg TLSServerCfg) Prepare() error {
 		tlscfg.ClientAuth = tls.NoClientCert
 	}
 
-	return MainInstance.SetServerTLSConfig(cfg.Name, tlscfg)
+	serverConf := tlscfg
+	if cfg.VerifyClientNodeID {
+		tlscfg.NextProtos = []string{"netceptor"}
+		serverConf = &tls.Config{}
```
ghjm: Maybe I'm misunderstanding this, but it seems to be throwing away the previous settings in serverConf?

pkg/netceptor/tlsconfig.go
```diff
@@ -65,7 +79,16 @@ func (cfg TLSServerCfg) Prepare() error {
...
+		serverConf.GetConfigForClient = func(hi *tls.ClientHelloInfo) (*tls.Config, error) {
+			tlscfg.VerifyPeerCertificate = getClientValidator(hi, tlscfg.ClientCAs)
```
ghjm: This seems to be modifying the outer-scope tlscfg?

ghjm:
In addition to what you've done here, I think we want to save the node ID that was claimed in the TLS hi message, and then add it to the check here: https://github.com/project-receptor/receptor/blob/06c12db072561454f2e17e8db65ebf34ff135226/pkg/netceptor/netceptor.go#L1305 so that your Netceptor-level connection will fail if you try to start the protocol using a different node ID than the one you asserted in TLS.

pkg/netceptor/tlsconfig.go
fosterseth: yea in the case of `VerifyClientNodeID`, we set serverConf to an empty tls.Config{} and only set the `GetConfigForClient` field, which ultimately returns the tlscfg we just defined.

if it is more clear I can do an if / else

```golang
if cfg.VerifyClientNodeID {
	tlscfg.NextProtos = []string{"netceptor"}
	serverConf = &tls.Config{}
	serverConf.GetConfigForClient = func(hi *tls.ClientHelloInfo) (*tls.Config, error) {
		tlscfg.VerifyPeerCertificate = getClientValidator(hi, tlscfg.ClientCAs)
		return tlscfg, nil
	}
} else {
	serverConf := tlscfg
}
return MainInstance.SetServerTLSConfig(cfg.Name, serverConf)
```

pkg/netceptor/tlsconfig.go
fosterseth: yeah it does, and I am not to sure how to get around it.

fosterseth:
@ghjm I'm also realizing my naming and description of the parameter might not be best, as this parameter could be set for the backend connections too, not just the receptor-level connections.

pkg/netceptor/tlsconfig.go
ghjm: Doesn't this drop the server's certificates when VerifyClientNodeID is true?

ghjm:
Yeah, we talked about that a day or two ago.  While it _could_ be used for backend connections, this can only provide actual security assurances at the IP address level (since the client can set its own local hostname to whatever it wants).  I don't think this is actually useful.

In the Receptor protocol, where names and network addresses are the same thing, we know the name that the connection came from, so it makes sense to verify it with the certificate.

pkg/netceptor/tlsconfig.go
fosterseth: the `GetConfigForClient` callback returns `tlscfg`, which will have `Certificates` as defined on line 62.
When `VerifyClientNodeID` is set, nothing is dropped from tlscfg, but `VerifyPeerCertificate` and `NextProtos` are added

squidboylan (approved):
Looks like we have tests that verify an incorrect CN fails and a correct CN with verification works, :+1:  from me from the testing standpoint, thanks!

ghjm (approved):
Using a null function as a flag to indicate we ought to create a closure at connection time is a _bit_ weird, but I can't think of anything much better, so LGTM.

Issue #207, opened 2020-10-06
Receptor occassionally panics when a backend disconnects


squidboylan:
I got the following panic when running `TestWork/work_submit_with_incorrect_tlsclient_CN`:

```
panic: send on closed channel

goroutine 50 [running]:
github.com/project-receptor/receptor/pkg/netceptor.(*Netceptor).handleMessageData(0xc0002c0c80, 0xc00032fe00, 0xc0000ebeb0, 0x6)
        /home/squid/work/project-receptor/receptor/pkg/netceptor/netceptor.go:1047 +0x35d
github.com/project-receptor/receptor/pkg/netceptor.(*Netceptor).runProtocol(0xc0002c0c80, 0x19393a0, 0xc0004a2000, 0x3ff0000000000000, 0x0, 0x0, 0x0)
        /home/squid/work/project-receptor/receptor/pkg/netceptor/netceptor.go:1249 +0x19a4
github.com/project-receptor/receptor/pkg/netceptor.(*Netceptor).AddBackend.func1.1(0xc0002c0c80, 0x19393a0, 0xc0004a2000, 0x3ff0000000000000, 0x0)
        /home/squid/work/project-receptor/receptor/pkg/netceptor/netceptor.go:318 +0x5f
created by github.com/project-receptor/receptor/pkg/netceptor.(*Netceptor).AddBackend.func1
        /home/squid/work/project-receptor/receptor/pkg/netceptor/netceptor.go:317 +0xe9
```

This is very rare and seems to be caused by something continuing to use a channel that has been closed because a backend has exited. I'm unsure if this is what caused the test to fail or if this happened when we were shutting down the nodes after the test failed, in either case it seems like a good idea to make note of.

Attached are the logs of the `go test` output with the `-test.v` flag and an archive of the mesh logs, the relevant node is `node1`, this was produced on commit `1396c8d72a697f532957ed353f1a8fea60f67465`. I will look into making a reproducer soon.

[go-test.log.zip](https://github.com/project-receptor/receptor/files/5335319/go-test.log.zip)
[mesh-logs.zip](https://github.com/project-receptor/receptor/files/5335321/mesh-logs.zip)

squidboylan:
I think this was fixed by #208, if it turns out this is still an issue we can reopen this

PR #208, opened 2020-10-06
Fix race condition when send happens after close

ghjm:
This PR fixes a potential race condition that I think may be the cause of https://github.com/project-receptor/receptor/issues/207.  In `handleMessageData`, the `listenerLock` read lock is released immediately on obtaining the result from the lookup.  This means that potentially, `packetconn.Close()` could be called after we look up the listener, but before we send to `pc.recvChan`.

The fix is to hold the lock while also checking whether the packetconn's context is still good, and if it is, continue holding the lock through the channel send.  Any call to `Close()` will be blocked (because it requests a write lock on `listenerLock`) until the channel send completes.

In the case that the connection is still in the listener table but the context is cancelled, we send an unreachable message.  This is future-proofing - in the current code, it is not possible for this to happen because `packetconn.Close()` requests the lock before cancelling the context.  Checking the context here protects against the case where a future code change unwittingly decides to cancel the context in some other circumstance, like an error handler.

### DIFF
--- a/pkg/netceptor/packetconn.go
+++ b/pkg/netceptor/packetconn.go
@@ -182,10 +182,10 @@ func (pc *PacketConn) Close() error {
 	pc.s.listenerLock.Lock()
 	defer pc.s.listenerLock.Unlock()
 	delete(pc.s.listenerRegistry, pc.localService)
-	close(pc.recvChan)
 	if pc.cancel != nil {
 		pc.cancel()
 	}
+	close(pc.recvChan)
 	if pc.advertise {
 		err := pc.s.removeLocalServiceAdvertisement(pc.localService)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/netceptor/packetconn.go
```diff
@@ -182,10 +182,10 @@ func (pc *PacketConn) Close() error {
 	pc.s.listenerLock.Lock()
 	defer pc.s.listenerLock.Unlock()
 	delete(pc.s.listenerRegistry, pc.localService)
-	close(pc.recvChan)
 	if pc.cancel != nil {
 		pc.cancel()
 	}
+	close(pc.recvChan)
```
squidboylan, line 188: I'm not sure I understand this change

squidboylan (approved):
this seems like it fixes #207 (though it's hard to tell due to the rarity of the bug). In any case I think this should be merged since it looks like it fixes a race condition, even if it doesnt fix #207

PR #212, opened 2020-10-07
Add --no-payload and pass-through command arguments

ghjm:
This addresses https://github.com/project-receptor/receptor/issues/210 by allowing a more natural syntax for additional run-time parameters to a work command.  Previously, you could do this with `receptorctl work submit echo ---param params="thing to echo"`, but this is awkward.  This PR allows you to do `receptorctl work submit echo thing to echo`, similar to the non-JSON command to the control service.

It is confusing that we have both Receptor-level "parameters" that go in the top of the `work submit` JSON command, and also "parameters" sent to a `work-command` when `allowruntimeparams=true` that go into the `params` key of the JSON command.  Perhaps we should choose different vocabulary terms and use them consistently.

This PR also adds a `--no-payload` option.  People have been trying to accomplish this with `--payload-literal=""` which doesn't work, so clearly there's a need for it.


shanemcd:
Wow, that was fast 😄

shanemcd (approved):
Beautiful.

squidboylan (approved):
LGTM

PR #213, opened 2020-10-07
Test adding runtime params

squidboylan:


squidboylan:
if i understand this right, this should add some coverage for https://github.com/project-receptor/receptor/issues/78 , if i understand that issue wrong it's still valuable coverage to have

ghjm (approved):

Issue #22, opened 2020-06-23
Remote units of work should have expiration


ghjm:
When issuing a unit of work to be executed on a remote node via `work submit`, the originator should specify an expiration time or time-to-live.  If the job hasn't been able to be sent to the remote node and started by this time, it should be marked as failed.

ghjm:
Ping @fosterseth 😊

squidboylan:
@fosterseth added testing for this in his PR

PR #214, opened 2020-10-07
Add --ttl (time to live) option to remote work submission commands

fosterseth:
related #22 

**What is added**
Adds `--ttl` to `receptorctl` and our JSON work submission, e.g.

`receptorctl --socket /tmp/foo.sock work submit 100cat --node bar --tlsclient client --ttl 3s --payload -`

If the remote work does not start before this time, the work unit is terminated and marked with a Failed status.

**--ttl format**
`--ttl` is a string and is flexible, described here https://golang.org/pkg/time/#ParseDuration

e.g. `10h30m10s`, `0.5h` are valid

**Internal representation**
Receptor converts this duration into an `Expiration time.Time` field on our RemoteExtraData structure

Here is what the new status looks like

```
'Y0NSwpjQ': {'Detail': 'Running: PID 228982',
              'ExtraData': {'Expiration': '2020-10-07T16:42:38.018557837-04:00',
                            'LocalCancelled': False,
                            'LocalReleased': False,
                            'RemoteNode': 'bar',
                            'RemoteParams': {'ttl': '3s'},
                            'RemoteStarted': True,
                            'RemoteUnitID': 'IgIKO6od',
                            'RemoteWorkType': '100cat',
                            'TLSClient': 'client'},
              'State': 1,
              'StateName': 'Running',
              'StdoutSize': 12,
              'WorkType': 'remote'}}
```

Here is the error message that hits if the work unit times out

`{'3weUcmLn': {'Detail': 'Work unit expired on Wed Oct 7 16:43:13',`

**Additional Notes**
The "nil" `time.Time` is Zero time, so work units without a user-provided TTL look like this

```
{'dO0iAgiv': {'Detail': 'Starting Worker',
              'ExtraData': {'Expiration': '0001-01-01T00:00:00Z',
```

Not very pretty unfortunately, but maybe not a big deal. An alternative is to just store the expiration as a string (if ttl not set, "") and do conversions to `time.Time` wherever needed.

### DIFF
--- a/pkg/workceptor/controlsvc.go
+++ b/pkg/workceptor/controlsvc.go
@@ -173,6 +173,10 @@ func (c *workceptorCommand) ControlFunc(nc *netceptor.Netceptor, cfo controlsvc.
 		if err != nil {
 			tlsclient = "" // optional so don't return
 		}
+		ttl, err := strFromMap(c.params, "ttl")
+		if err != nil {
+			ttl = ""
+		}
 		workParams := make(map[string]string)
 		for k, v := range c.params {
 			if k == "command" || k == "subcommand" || k == "node" || k == "worktype" || k == "tlsclient" {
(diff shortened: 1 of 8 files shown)


ghjm (commented):
I pulled this down and ran it, and it looks good in all the cases I could think of.  The only comment I have is that since (at least for now) the ttl parameter is only meaningful for remote work, I think we ought to return an error if you try to submit local work with a ttl.  (I think returning an error is better than just silently failing to respect it.)

ghjm (approved):
LGTM if CI passes

fosterseth:
@ghjm thanks, it now returns an error if `ttl` is specified when submitting local work.

PR #233, opened 2020-10-13
Fix traceroute

ghjm:
See https://github.com/project-receptor/receptor/issues/223.  The problem was that at some previous stage, the `SubscribeUnreachable` mechanism was switched from sending `UnreachableMessage` to sending `map[string]string`, but this change was only partially completed.  This didn't cause any errors because the `Broker` class takes `interface{}`, so it passed these messages through just fine, but then they weren't processed on the receiving side.

This PR adds type enforcement to the `Broker` class, and adds a new `UnreachableNotification` struct, which includes `UnreachableMessage` and also the additional information of which node the message was received from.


fosterseth (approved):

ghjm:
I added more stuff to this @fosterseth.  Requesting re-review

fosterseth (approved):

Issue #86, opened 2020-08-11
Kubernetes worker does not preserve state on Receptor restart


ghjm:
If you launch a Kubernetes job and then restart Receptor, it loses track of the job - stdout is never made available, the job is not cancellable, etc.  To fix this we need to do a rewrite of Workceptor to allow storing worker-specific data in the status file, most notably the pod ID.  This Workceptor rewrite is also likely needed for https://github.com/project-receptor/receptor/issues/78.

squidboylan:
looks like the stdout we get in this scenario is incorrect, see https://github.com/project-receptor/receptor/pull/257 for my test. It looks like the kubernetes job stdout is:

```
1
1
2
3
4
5
```
so the first line is being duplicated (im guessing that's where the restart is happening).

PR #234, opened 2020-10-14
k8s work units can restart

fosterseth:
related #86 

kubeunits can restart if both conditions are met

1. Work unit is in running state
2. `streammethod` is `logger`

If those fail, we have an option to delete the pod that is associated with the kubeunit,

`deletepodonrestart=<bool>: On restart, delete the pod if in pending state (default: true)`

streammethod `tcp` is complicated because we would need the pod to reconnect back to the tcp server after the receptor restart, so this PR does not address that.

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -182,21 +182,31 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
 }
 
 func (kw *kubeUnit) runWorkUsingLogger() {
-
-	// Create the pod
 	skipStdin := false
-	err := kw.createPod(nil)
-	if err == ErrPodCompleted {
+	status := kw.Status()
+	ked := status.ExtraData.(*kubeExtraData)
+	var err error
+	if status.State != WorkStateRunning {
+		// Create the pod
+		err = kw.createPod(nil)
+		if err == ErrPodCompleted {
+			skipStdin = true
+		} else if err != nil {
+			errMsg := fmt.Sprintf("Error creating pod: %s", err)
+			kw.UpdateBasicStatus(WorkStateFailed, errMsg, 0)
+			logger.Error(errMsg)
+			return
+		}
+	} else {
 		skipStdin = true
-	} else if err != nil {
-		errMsg := fmt.Sprintf("Error creating pod: %s", err)
-		kw.UpdateBasicStatus(WorkStateFailed, errMsg, 0)
-		logger.Error(errMsg)
-		return
+		kw.pod, err = kw.clientset.CoreV1().Pods(ked.KubeNamespace).Get(kw.ctx, ked.PodName, metav1.GetOptions{})
+		if err != nil {
+			logger.Error(err.Error())
+			return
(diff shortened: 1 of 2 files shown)


pkg/workceptor/kubernetes.go
```diff
@@ -661,7 +669,37 @@ func (kw *kubeUnit) Start() error {
...
+	if status.State == WorkStateRunning && !isTCP {
+		return kw.startOrRestart()
+	}
+	// Work unit is in Pending state
+	if kw.deletePodOnRestart {
+		err := kw.connectToKube()
+		if err != nil {
+			logger.Warning("Pod %s could not be deleted: %s", ked.PodName, err.Error())
+		} else {
+			err := kw.clientset.CoreV1().Pods(ked.KubeNamespace).Delete(context.Background(), ked.PodName, metav1.DeleteOptions{})
```
fosterseth, line 690: I noticed in `kw.Cancel()` we put this delete in a go routine, but I didn't have any issues with blocking on this api call. Should I also put this into a go routine?

pkg/workceptor/kubernetes.go
```diff
@@ -182,21 +183,33 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
...
-
-	// Create the pod
 	skipStdin := false
-	err := kw.createPod(nil)
-	if err == ErrPodCompleted {
+	status := kw.Status()
+	ked := status.ExtraData.(*kubeExtraData)
+	var err error
+	var errMsg string
+	if status.State != WorkStateRunning {
```
ghjm: I don't think this check is sufficient.  We leave a pod in WorkStatePending until we've sent complete stdin to it.  I think a better check here would be whether kubeExtraData.PodName is set (which we do right when we create the pod).

pkg/workceptor/kubernetes.go
```diff
@@ -259,12 +271,12 @@ func (kw *kubeUnit) runWorkUsingLogger() {
 		kw.UpdateBasicStatus(WorkStateFailed, errMsg, 0)
 		return
 	}
-	kw.UpdateBasicStatus(WorkStatePending, "Sending stdin to pod", 0)
 
 	// Goroutine to update status when stdin is fully sent to the pod, which is when we
 	// update from WorkStatePending to WorkStateRunning.
 	finishedChan := make(chan struct{})
 	if !skipStdin {
+		kw.UpdateBasicStatus(WorkStatePending, "Sending stdin to pod", 0)
```
ghjm: I think it's possible that a pod may already have started sending stdout when this line runs - in which case it is wrong to set the stdoutSize to 0.  Maybe switch this to an UpdateFullStatus that only changes the state & description.

pkg/workceptor/kubernetes.go
ghjm: As discussed, there doesn't seem to be a reason for either to be in separate goroutines.

ghjm (changes requested):

ghjm (approved):

PR #235, opened 2020-10-15
Only restart remote work if RemoteStarted is true

fosterseth:
When we flip this boolean to True, it is safe to resume monitoring of that remote work.

### DIFF
--- a/pkg/workceptor/remote_work.go
+++ b/pkg/workceptor/remote_work.go
@@ -531,7 +531,11 @@ func (rw *remoteUnit) Start() error {
 
 // Restart resumes monitoring a job after a Receptor restart
 func (rw *remoteUnit) Restart() error {
-	return rw.startOrRestart(false)
+	red := rw.Status().ExtraData.(*remoteExtraData)
+	if red.RemoteStarted {
+		return rw.startOrRestart(false)
+	}
+	return fmt.Errorf("remote work had not previously started")
 }
 
 // cancelOrRelease is a shared implementation of Cancel() and Release()


pkg/workceptor/remote_work.go
```diff
@@ -531,7 +531,11 @@ func (rw *remoteUnit) Start() error {
 
 // Restart resumes monitoring a job after a Receptor restart
 func (rw *remoteUnit) Restart() error {
-	return rw.startOrRestart(false)
+	red := rw.Status().ExtraData.(*remoteExtraData)
+	if red.RemoteStarted {
+		return rw.startOrRestart(false)
+	}
+	return fmt.Errorf("remote work had not previously started")
```
ghjm, line 538: Should this be an error?  I was thinking work in this state would just be silently discarded.

fosterseth: returning an error here results in a Warning in caller https://github.com/project-receptor/receptor/blob/0374f78d160383fbc92a1a2bc28a1c1a319729a4/pkg/workceptor/workceptor.go#L227

are you suggesting we do a `release` in this scenario?

ghjm: A warning sounds fine.  I think my concern here was not to keep retrying.

I don't think we should do a release until the user actually asks for one.  Do we mark the unit as failed?

ghjm (approved):

pkg/workceptor/remote_work.go
fosterseth: Yep, unit is marked as failed

PR #237, opened 2020-10-20
Add a receptorctl test suite

squidboylan:


Makefile
```diff
@@ -149,6 +149,14 @@ tc-image: container
 	@cp receptor packaging/tc-image/
 	@$(CONTAINERCMD) build packaging/tc-image -t receptor-tc
 
+receptorctl-test-venv: receptor
```
ghjm: Why does receptorctl-test-venv depend on receptor?

Makefile
```diff
@@ -149,6 +149,14 @@ tc-image: container
...
+	virtualenv receptorctl-test-venv -p python3
+	receptorctl-test-venv/bin/pip install -e receptorctl
+	receptorctl-test-venv/bin/pip install -r receptorctl/test-requirements.txt
+
+receptorctl-tests: receptor receptorctl-test-venv
+	cd receptorctl && ../receptorctl-test-venv/bin/pytest tests/tests.py
+
 clean:
```
ghjm, line 160: Make clean should remove receptorctl-test-venv

Makefile
ghjm: The make target here should be a file rather than a directory (I would suggest `receptorctl-test-venv/bin/pytest`) so the Makefile can avoid rebuilding when not necessary.

squidboylan: this was juts a copy paste error

ghjm (approved):

PR #239, opened 2020-10-21
Improvements to receptorctl.socket_interface.submit_work interface

shanemcd:
Usage before:

```
receptor_ctl.submit_work('localhost',
                         'worker',
                         sockout.makefile('rb'), '', '', {})
```

After:

```
receptor_ctl.submit_work(worktype='worker',
                         payload=sockout.makefile('rb'))
```

Note: `worktype` and `payload` are *required* keyword arguments, with no defaults. All other arguments are optional keyword arguments.

### DIFF
--- a/receptorctl/receptorctl/socket_interface.py
+++ b/receptorctl/receptorctl/socket_interface.py
@@ -98,17 +98,23 @@ class ReceptorControl:
         if not str.startswith(text, "Connecting"):
             raise RuntimeError(text)
 
-    def submit_work(self, node, worktype, payload, tlsclient, ttl, params):
+    def submit_work(self, worktype, payload, node=None, tlsclient=None, ttl=None, params=None):
         if node is None:
             node = "localhost"
+
         commandMap = {
             "command": "work",
             "subcommand": "submit",
             "node": node,
             "worktype": worktype,
-            "tlsclient": tlsclient,
-            "ttl": ttl,
         }
+
+        if tlsclient:
+            commandMap['tlsclient'] = tlsclient
+
+        if ttl:
+            commandMap['ttl'] = ttl
+
         if params:
             for k,v in params.items():
                 if k not in commandMap:


receptorctl/receptorctl/socket_interface.py
```diff
@@ -98,7 +98,7 @@ def connect_to_service(self, node, service, tlsclient):
         if not str.startswith(text, "Connecting"):
             raise RuntimeError(text)
 
-    def submit_work(self, node, worktype, payload, tlsclient, ttl, params):
+    def submit_work(self, *, worktype, payload, node=None, tlsclient='', ttl='', params={}):
```
ghjm: AIUI `params={}` creates a new dict _once_ at module load time, and then uses it as the default.  So if params is ever modified by the function, the modification will stick around forever.  So we probably want to have the default be None and then test for that and instantiate a new dict in the function body (assuming the code can't deal with it being None).

Also, what does the `*` do?  I haven't seen this syntax before.

ghjm: Speaking of which, None is probably a better default than the empty string for tlsclient and (particularly) ttl.  And if the code can't deal with None values, it should be fixed, because someday someone will pass one in.

shanemcd: > what does the * do?

https://www.python.org/dev/peps/pep-3102/

This syntax makes `worktype` and `payload` required keyword arguments, with no default value. 

This should illustrate it:

```
def myfunc(*, foo, bar):
    print(foo)
    print(bar)

myfunc()
```

Throws the error:

```
  File "test2.py", line 6, in <module>
    myfunc()
TypeError: myfunc() missing 2 required keyword-only arguments: 'foo' and 'bar
```

As an aside: if we ever wanted to add a new positional argument, it would need to be _before_ the `*`:

```
def myfunc(buzz, *, foo, bar):
    print(foo)
    print(bar)
    print(buzz)


myfunc('', foo='', bar='')
```

ghjm: Why do we want this?  If we just do

```
def submit_work(self, worktype, payload, node=None, tlsclient=None, ttl=None, params=None):
```
then if we ever want to add more parameters, that's fine - as long as they have defaults they won't break existing code.  I don't see any reason why we would ever need variadic arguments to this function.

shanemcd: @ghjm Check out the latest diff. I've changed them all to `None`. But I much prefer keyword arguments over positional arguments. It makes reading the implementation code easier. It's possible to add whatever kinds of parameters we need to in the future, so I'm not sure I understand the concern.

receptorctl/receptorctl/socket_interface.py
```diff
@@ -98,9 +98,12 @@ def connect_to_service(self, node, service, tlsclient):
...
             raise RuntimeError(text)
 
-    def submit_work(self, node, worktype, payload, tlsclient, ttl, params):
-        if node is None:
-            node = "localhost"
+    def submit_work(self, *, worktype, payload, node=None, tlsclient=None, ttl=None, params=None):
+        node = node or "localhost"
+        tlsclient = tlsclient or ''
+        ttl = ttl or ''
+        params = params or ''
```
ghjm: I think it would be better to conditionally add these optional parameters to commandMap rather than setting them to the empty string.  If params is None then the `if params` below will skip the section, for example.  And the rest are optional and don't need to be in the json if we aren't using them.

shanemcd: @ghjm Done. How about now?

receptorctl/receptorctl/socket_interface.py
ghjm: The concern is to reduce cognitive demands on API users, who are just using our library as one of hundreds they have to deal with as they go about their day.  Imagine if every library had a stylistic preference that they enforced on their callers.  I think we should stick to vanilla-looking Python that any user will easily understand and be able to use.

receptorctl/receptorctl/socket_interface.py
ghjm: Yep, this looks good, thanks!

ghjm (approved):

PR #240, opened 2020-10-22
Add duplicate node detection

ghjm:
This PR adds duplicate node detection (see https://github.com/project-receptor/receptor/issues/19).  When we launch a node with a node ID that already exists on the mesh, something like this happens:

```
(graham-dev) ✔ ~/git/receptor [dupe_detector L|✚ 1]
21:29 $ receptor --node id=foo --tcp-peer address=localhost:4444 --log-level debug
DEBUG 2020/10/21 21:29:34 Running TCP peer connection localhost:4444
DEBUG 2020/10/21 21:29:34 Sending initial connection message
INFO 2020/10/21 21:29:34 Connection established with baz
DEBUG 2020/10/21 21:29:34 Stopping initial updates
INFO 2020/10/21 21:29:35 Initialization complete
DEBUG 2020/10/21 21:29:35 Sending routing update A3Low6Hl. Connections: baz(1.00)
DEBUG 2020/10/21 21:29:35 flooding to baz
DEBUG 2020/10/21 21:29:35 Re-calculating routing table
INFO 2020/10/21 21:29:35 Known Connections:
INFO 2020/10/21 21:29:35    foo: baz(1.00)
INFO 2020/10/21 21:29:35    baz: foo(1.00)
INFO 2020/10/21 21:29:35 Routing Table:
INFO 2020/10/21 21:29:35    baz via baz
DEBUG 2020/10/21 21:29:35 Received routing update HoGyHAIY from baz via baz
ERROR 2020/10/21 21:29:35 We are a duplicate node.  Shutting down.
```

The process of detecting a duplicate is as follows:
* Routing update messages now have an extra field, SuspectedDuplicate.
* When we receive a routing update from a node with our name, we look to see if the epoch is newer than ours.  If it is, then we suspect the other node is a duplicate.  (If clocks are in sync, then the other node is the one that joined later.  Even if clocks are not in sync, the comparison deterministically chooses which node is considered the legitimate owner of the node ID.)
* When we find a duplicate, we send a routing update with the SuspectedDuplicate field filled in.
* On receiving this, the duplicate node exits.
* When other nodes on the network see the SuspectedDuplicate, they adjust their known node epochs back to the older epoch, to avoid the situation where they are "stuck" at the higher epoch they previously saw.

The following changes were also made to support this:
* Epoch numbers, which were previously just a Unix timestamp, now have 24 low-order random bits added.  This gives a 99.99999404% chance that duplicate nodes will have different epochs, even if they joined in the same second (as in testing).  If two nodes joined in the same second, the one with the higher random data will be the one that exits.
* A new lock, seenUpdatesLock, is added to protect access to seenUpdates (previously knownNodeLock was used for both).
* Debug logging of routing updates now gives the routing update ID so you can trace the flow of specific updates across nodes.
* New connections are refused if the connecting node has our node ID, or the node ID of one of our direct connections.

### DIFF
--- a/pkg/netceptor/netceptor.go
+++ b/pkg/netceptor/netceptor.go
@@ -15,6 +15,7 @@ import (
 	"github.com/project-receptor/receptor/pkg/utils"
 	"io"
 	"math"
+	"math/rand"
 	"reflect"
 	"strings"
 	"sync"
@@ -82,8 +83,9 @@ type Netceptor struct {
 	connLock               *sync.RWMutex
 	connections            map[string]*connInfo
 	knownNodeLock          *sync.RWMutex
-	seenUpdates            map[string]time.Time
 	knownNodeInfo          map[string]*nodeInfo
+	seenUpdatesLock        *sync.RWMutex
+	seenUpdates            map[string]time.Time
 	knownConnectionCosts   map[string]map[string]float64
 	routingTableLock       *sync.RWMutex
 	routingTable           map[string]string
@@ -167,12 +169,13 @@ type nodeInfo struct {
 }
 
 type routingUpdate struct {
-	NodeID         string
-	UpdateID       string
-	UpdateEpoch    uint64
-	UpdateSequence uint64
-	Connections    map[string]float64
-	ForwardingNode string
+	NodeID             string
+	UpdateID           string
+	UpdateEpoch        uint64
+	UpdateSequence     uint64
+	Connections        map[string]float64
+	ForwardingNode     string
+	SuspectedDuplicate uint64
 }
 
 // ServiceAdvertisement is the data associated with a service advertisement
@@ -235,13 +238,14 @@ func New(ctx context.Context, NodeID string, AllowedPeers []string) *Netceptor {
 	s := Netceptor{
 		nodeID:                 NodeID,
 		allowedPeers:           AllowedPeers,
-		epoch:                  uint64(time.Now().Unix()),
+		epoch:                  uint64(time.Now().Unix()*(1<<24)) + uint64(rand.Intn(1<<24)),
 		sequence:               0,
 		connLock:               &sync.RWMutex{},
 		connections:            make(map[string]*connInfo),
 		knownNodeLock:          &sync.RWMutex{},
-		seenUpdates:            make(map[string]time.Time),
 		knownNodeInfo:          make(map[string]*nodeInfo),
+		seenUpdatesLock:        &sync.RWMutex{},
+		seenUpdates:            make(map[string]time.Time),
 		knownConnectionCosts:   make(map[string]map[string]float64),
 		routingTableLock:       &sync.RWMutex{},
 		routingTable:           make(map[string]string),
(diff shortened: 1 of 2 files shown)


fosterseth:
> On receiving this, the duplicate node exits.

I tried the following, start `foo.yml`, then `bar.yml`, then `bardup.yml`

I see errors on `foo`

```
ERROR 2020/10/22 15:08:43 Backend receiving error read tcp [::1]:2222->[::1]:46018: use of closed network connection
ERROR 2020/10/22 15:08:43 Backend error: rejected connection with node bar because it connected using a node ID we are already connected to
```

but my duplicate `bar` node is not shutting down

foo.yml
```yaml
---
- node:
    id: foo
    allowedpeers: bar

- tcp-listener:
    port: 2222
```

bar.yml and bardup.yml
```yaml
---
- node:
    id: bar
    allowedpeers: foo

- tcp-peer:
    address: localhost:2222
```

ghjm:
Yeah, that situation is being treated as a failed connection rather than a duplicate detection, so if you have retries enabled or are a listener, it just keeps retrying or listening.  I guess in that situation we should probably shutdown?  I'm not totally sure what the right behavior is here.

To exercise the duplicate node detection without triggering the connection checks, you'll need to have a mesh where no single node can "see" that there's a duplicate within its local connections, like `foo - bar - baz - foo`.  In the case of `foo - bar - foo`, the bar node is immediately aware of the duplicate foo, and doesn't need to use any fancy network shenanigans to discover that there's a problem.

fosterseth:
I'd be in favor of leaving it as is. It seems if you bring down the first bar, the duplicate bar is picked up and works just fine. I'd be more worried of a receptor node that hangs around in a broken state and cannot properly function no matter how the topology changes.

fosterseth (approved):

fosterseth (approved):

squidboylan:
this sounds like it'll work for a stable mesh, but what will happen if our mesh is already in a state of flux, is there potential for both nodes to be told they're both duplicates and shut down? I'm trying to wrap my head around this and I don't think there's a problem but I'm not sure

ghjm:
@squidboylan this is one of the things I struggled with.  The question is, how do you tell the difference between a legitimately duplicate node, and a node that just happens to have dropped and rejoined the mesh, and maybe there are routing updates still circulating from its old instantiation.

The idea is that in order for a node to be detected as a duplicate, two things have to happen:
* One node sees a routing update from another with a _newer_ epoch than itself.  That means the other node isn't just an echo of its own past routing updates.
* This node then sends a special routing update, with SuspectedDuplicate set, that says "I think the node with such-and-so epoch is a duplicate of me."
* The duplicate node has to receive this routing update, and only then will it shut down.

Because it takes concerted action from _both_ the original node and the duplicate, it's unlikely that any sort of lingering messages could cause a node to drop off.

There are still some concerns:
* The concept of "newer" is only valid if the two nodes' clocks are reasonably in sync.  If that's not the case, then we'll still deterministically select one node to drop (not both), but it might be the older node.
* The fact that the nodes need to cooperate means this is a solution only for configuration errors.  We don't attempt to handle the case where a node is running corrupted or malicious code or otherwise misbehaving.
* During the time the duplicate node exists, it is sending routing updates, which other nodes will believe.  We attempt to patch this up after the fact by having the other nodes see the SuspectedDuplicate message and revert their local data.  But if the mesh is in flux, a node might see the duplicate node's update and then _not_ see the SuspectedDuplicate message, which means it might get stuck with a wrong epoch and ignore future routing updates from the legitimate node.

PR #242, opened 2020-10-29
Minikube in action

squidboylan:


squidboylan:
Im expecting CI for this to fail but wanted to open a PR so others could take a look, i *think* we might have a bug in receptor that's causing the CI to fail.

squidboylan:
```
{"State":3,"Detail":"Error creating pod: expected 1 container in pod but there were 0","StdoutSize":0,"WorkType":"remote","ExtraData":{"RemoteNode":"node3","RemoteWorkType":"echosleepshort","RemoteParams":{},"RemoteUnitID":"BttZIgvp","RemoteStarted":true,"LocalCancelled":false,"LocalReleased":false,"TLSClient":"","Expiration":"0001-01-01T00:00:00Z"}}
```

from the work status file. See the test artifacts for more info

ghjm:
Would really like to see the Kubernetes logs here

squidboylan:
> Would really like to see the Kubernetes logs here

I'm not sure how we can do this in CI in a good way, it might be best to reproduce locally against a local k8s cluster if you can. It should be fairly easy to do that if you download this branch. If that can't be done let me know and I can try think of alternative.

ghjm:
I'm in meetings all day and was hoping if I said that, you would just magically figure out how to get the minikube logs wedged into test-logs.zip 😀

squidboylan:
heh, ok, yeah i figured that might be the case. I'll see what i can do

ghjm:
After this PR, what is the expectation for running `make ci` locally?  Do you need to have a minikube running (or access to full kubernetes, or something) in order to get through ci tests?

squidboylan:
you need a kubectl config that can talk to a k8s where the tests can run. Currently that's all and i'd like to keep it that way, but that may limit some of what we can do. Currently we're not actually using the minikube registry that we're starting

ghjm:
I fixed up the makefile for the kubetest target, but it never occurred to me to wonder: what is this actually for?

ghjm:
Also, FWIW, I'm seeing the same CI failures with a local `make ci`.

ghjm (approved):

ghjm:
@squidboylan   Tests are passing.  This is your PR so I'll leave it to you to merge it.  It looks good to me.

PR #244, opened 2020-11-03
Add tls support to receptorctl

fosterseth:
related https://github.com/project-receptor/receptor/issues/224

Can now do the following,
`receptorctl --socket tls://localhost:11114 --rootcas ca.crt --key client.key --cert client.crt --insecureskipverify true status`

One can also point to a receptor-compatible yaml file with tls client information

foo.yaml
```yaml
- tls-client:
    name: thisclient
    rootcas: ca.crt
    insecureskipverify: true
    cert: client.crt
    key: client.key
```
`receptorctl --socket tls://localhost:11114 --yaml  foo.yaml --tlsclient thisclient status`

or set up environment variables

`export RECEPTORCTL_YAML=foo.yaml` and `export RECEPTORCTL_TLSCLIENT=thisclient`

if `rootcas` is None (default), then it _should_ use system certs, although I could not get this to work properly on my machine.

https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_default_certs

### DIFF
--- a/packaging/rpm/receptorctl.spec.j2
+++ b/packaging/rpm/receptorctl.spec.j2
@@ -20,6 +20,7 @@ BuildRequires: python36-setuptools
 Requires: python36
 Requires: python36-setuptools
 Requires: python36-dateutil
+Requires: python36-pyyaml
 %else
 BuildRequires: python3
 BuildRequires: python3-setuptools
--- a/packaging/rpm/receptorctl.spec.j2
+++ b/packaging/rpm/receptorctl.spec.j2
@@ -27,6 +27,7 @@ Requires: python3
 Requires: python3-setuptools
 Requires: python3-dateutil
 Requires: python3-click
+Requires: python3-pyyaml
 %endif
 
 %description
--- a/receptorctl/receptorctl/cli.py
+++ b/receptorctl/receptorctl/cli.py
@@ -31,17 +31,23 @@ class IgnoreRequiredWithHelp(click.Group):
 @click.pass_context
 @click.option('--socket', envvar='RECEPTORCTL_SOCKET', required=True, show_envvar=True,
               help="Control socket address to connect to Receptor (defaults to Unix socket, use tcp:// for TCP socket)")
-@click.option('--rootcas', default="", envvar='RECEPTORCTL_ROOTCAS', required=False, show_envvar=True,
-              help="Root CA bundle to use instead of system trust when connecting with tls")
-def cli(ctx, socket, rootcas):
+# @click.option('--rootcas', default="", envvar='RECEPTORCTL_ROOTCAS', required=False, show_envvar=True,
+#               help="Root CA bundle to use instead of system trust when connecting with tls")
+@click.option('--tlsclient', default=None, envvar='RECEPTORCTL_TLSCLIENT', required=False, show_envvar=True,
+              help="TLS client name specified in yaml")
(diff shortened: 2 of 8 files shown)


receptorctl/receptorctl/cli.py
```diff
@@ -31,14 +31,28 @@ def parse_args(self, ctx, args):
 @click.pass_context
 @click.option('--socket', envvar='RECEPTORCTL_SOCKET', required=True, show_envvar=True,
               help="Control socket address to connect to Receptor (defaults to Unix socket, use tcp:// for TCP socket)")
-def cli(ctx, socket):
+@click.option('--yaml', default=None, envvar='RECEPTORCTL_YAML', required=False, show_envvar=True,
```
ghjm: I think this should be named `--config` for consistency with the main receptor CLI.

receptorctl/receptorctl/cli.py
```diff
@@ -31,14 +31,28 @@ def parse_args(self, ctx, args):
...
+              help="Yaml filename configured for receptor")
+@click.option('--tlsclient', default=None, envvar='RECEPTORCTL_TLSCLIENT', required=False, show_envvar=True,
```
ghjm: Maybe call this tls-client instead of tlsclient, for the same reason.

receptorctl/receptorctl/socket_interface.py
```diff
@@ -33,27 +34,49 @@ def read_and_parse_json(self):
...
+        rootcas = ctxobj["rootcas"]
+        insecureskipverify = ctxobj["insecureskipverify"]
         if self.socket is not None:
             raise ValueError("Already connected")
-        m = re.compile("tcp:(//)?([a-zA-Z0-9-]+):([0-9]+)|(unix:(//)?)?([^:]+)").fullmatch(address)
+        m = re.compile("(tcp|tls):(//)?([a-zA-Z0-9-]+):([0-9]+)|(unix:(//)?)?([^:]+)").fullmatch(address)
         if m:
-            if m[6]:
-                path = os.path.expanduser(m[6])
+            if m[7]:
```
ghjm: This isn't great code style (I can say that since it was me who originally wrote it).  It would be better to assign the `m[x]` values to named variables up top, and then refer to them by name in later code.

ghjm (approved):

packaging/rpm/receptorctl.spec.j2
```diff
@@ -27,6 +27,7 @@ Requires: python3
 Requires: python3-setuptools
 Requires: python3-dateutil
 Requires: python3-click
+Requires: python3-pyyaml
```
ghjm, line 31: RHEL 7 section also needs a Requires: for pyyaml, probably `python36-pyyaml`

ghjm (changes requested):

receptorctl/receptorctl/cli.py
fosterseth: fixed, can now do `--config` or `-c`

receptorctl/receptorctl/cli.py
fosterseth: fixed

receptorctl/receptorctl/socket_interface.py
fosterseth: fixed

ghjm (approved):

PR #250, opened 2020-11-10
Fix receptorctl work submit

shanemcd:
I broke this in https://github.com/project-receptor/receptor/pull/239, sorry!

### DIFF
--- a/receptorctl/receptorctl/cli.py
+++ b/receptorctl/receptorctl/cli.py
@@ -259,7 +259,7 @@ def submit(ctx, worktype, node, payload, no_payload, payload_literal, tlsclient,
         if node == "":
             node = None
         rc = get_rc(ctx)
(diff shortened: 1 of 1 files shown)


receptorctl/receptorctl/cli.py
```diff
@@ -245,7 +245,7 @@ def submit(ctx, worktype, node, payload, no_payload, payload_literal, tlsclient,
         if node == "":
             node = None
         rc = get_rc(ctx)
-        work = rc.submit_work(node, worktype, payload_data, tlsclient, ttl, params)
+        work = rc.submit_work(worktype, payload_data, tlsclient=tlsclient, ttl=ttl, params=params)
```
ghjm: Do we need `node=node` as well?

shanemcd: yup. good catch.

ghjm (approved):

PR #251, opened 2020-11-11
Remove ctxobj in ReceptorControl

fosterseth:
Now TLS information is supplied to the ReceptorControl `__init__`, instead of the `.connect()`

the connection remains open so that you can sequence commands together

```python
rc = receptorctl.ReceptorControl("tls://localhost:8888", config="foo.yml", tlsclient="client")
rc.simple_command("status")
rc.simple_command("work list")
rc.get_work_results("6ByU8r4b")
```

Note:
Also addresses an issue where shutting down the write side of the socket does not behave correctly for SSLSocket
https://bugs.python.org/issue18880

### DIFF
--- a/receptorctl/receptorctl/socket_interface.py
+++ b/receptorctl/receptorctl/socket_interface.py
@@ -9,15 +9,30 @@ import ssl
 import yaml
 
 class ReceptorControl:
-    def __init__(self):
+    def __init__(self, socketaddress, config=None, tlsclient=None, rootcas=None, key=None, cert=None, insecureskipverify=False):
+        if config and any((rootcas, key, cert)):
+            raise RuntimeError("Cannot specify both config and rootcas, key, cert")
+        if config and not tlsclient:
+            raise RuntimeError("Must specify both config and tlsclient")
+        self.socketaddress = socketaddress
         self.socket = None
         self.sockfile = None
         self.remote_node = None
+        self.rootcas = rootcas
+        self.key = key
+        self.cert = cert
+        self.insecureskipverify = insecureskipverify
+
+        if config and tlsclient:
+            self.readconfig(config, tlsclient)
+
+        self.connect()
 
     def readstr(self):
         return self.sockfile.readline().decode().strip()
 
     def writestr(self, str):
+        import pdb; pdb.set_trace()
         self.sockfile.write(str.encode())
         self.sockfile.flush()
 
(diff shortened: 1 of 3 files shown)


receptorctl/receptorctl/socket_interface.py
```diff
@@ -9,10 +9,22 @@
...
         self.sockfile = None
         self.remote_node = None
+        self.socketaddress = socketaddress
+        self.rootcas = rootcas
+        self.key = key
+        self.cert = cert
+        self.insecureskipverify = insecureskipverify
+        if config and tlsclient:
+            self.readconfig(config, tlsclient)
+        self.connect()
```
fosterseth: is it okay to go ahead and call connect here?

receptorctl/receptorctl/socket_interface.py
```diff
@@ -34,40 +46,34 @@ def read_and_parse_json(self):
...
-        address = ctxobj["socket"]
-        key = ctxobj["key"]
-        cert = ctxobj["cert"]
-        rootcas = ctxobj["rootcas"]
-        insecureskipverify = ctxobj["insecureskipverify"]
+    def connect(self):
         if self.socket is not None:
-            raise ValueError("Already connected")
-        m = re.compile("(tcp|tls):(//)?([a-zA-Z0-9-]+):([0-9]+)|(unix:(//)?)?([^:]+)").fullmatch(address)
+            return
```
fosterseth, line 75: I think it's okay to remove the error here. User can call `.close()` first if a different connection is needed

receptorctl/receptorctl/socket_interface.py
ghjm: I think it would be better to track connection state and lazily connect when someone calls a method that requires a connection.

receptorctl/receptorctl/socket_interface.py
ghjm: Yep, agreed

receptorctl/receptorctl/socket_interface.py
```diff
@@ -124,12 +130,14 @@ def close(self):
...
 
     def connect_to_service(self, node, service, tlsclient):
+        self.connect()
         self.writestr(f"connect {node} {service} {tlsclient}\n")
         text = self.readstr()
         if not str.startswith(text, "Connecting"):
             raise RuntimeError(text)
 
     def submit_work(self, worktype, payload, node=None, tlsclient=None, ttl=None, params=None):
+        self.connect()
```
ghjm, line 143: If this method succeeds, it's going to leave the socket in an unusable state for further commands, so we need to note that the next command will need a fresh connection, probably by setting `self.socket = None`.

Also, we're returning the filelike object generated by `socket.makefile()`.  We want to leave this open so the caller can read from it, but we also want the caller to be able to close it.  The way `makefile` works is it duplicates the file descriptor, so in order to close the underlying socket connection, you have to close _both_ the socket _and_ the sockfile.  So I think we should go ahead and close the socket here, meaning that when the caller closes the sockfile, the connection will be dropped.

receptorctl/receptorctl/socket_interface.py
```diff
@@ -180,6 +188,7 @@ def submit_work(self, worktype, payload, node=None, tlsclient=None, ttl=None, pa
         return result
 
     def get_work_results(self, unit_id):
+        self.connect()
```
ghjm, line 194: Same comment here as on submit_work

receptorctl/receptorctl/socket_interface.py
fosterseth: Updated to remove `connect` on init. 
The way I'm tracking connection state is by checking if `self.socket` is None in the connect method. Do you think that suffices?

ghjm: Yes, I think so, but see my other comments.  We need to make sure to set `self.socket` back to None in cases where the socket remains open, but is no longer usable for subsequent commands.

ghjm (approved):

Issue #206, opened 2020-10-06
Add service protocol to service advertisements


ghjm:
Add flags to `ServiceAdvertisement` that indicates whether the service listener is a PacketConn (datagram) or a Conn (stream) service, and in the latter case, whether it is TLS or non-TLS.

PR #253, opened 2020-11-13
Add connection info to service advertisement 

fosterseth:
related #206 

calling `status` now has connection information

```
Node ID: foo

foo          control   StreamTLS  2020-11-13 17:37:11  None
foo          echo      Stream     2020-11-13 17:37:11  {'type': 'Command Service'}
foo          udp       Datagram   2020-11-13 17:37:11  {'address': 'localhost:8989', 'type': 'UDP Proxy'}
```

I added a single byte to the ServiceAdvertisement struct to keep track of the connection information.
As it stands, cli.py maps the byte to the string. That means calling `status` from the unix socket will only display the byte (0,1, or 2), not the string, which is less useful.

### DIFF
--- a/pkg/netceptor/netceptor.go
+++ b/pkg/netceptor/netceptor.go
@@ -179,9 +179,12 @@ type routingUpdate struct {
 }
 
 const (
-	typePacketConn = 0
-	typeConnNonTLS = 1
-	typeConnTLS    = 2
+	// ConnTypeDatagram indicates a packetconn (datagram) service listener
+	ConnTypeDatagram = 0
+	// ConnTypeStream indicates a conn (stream) service listener, without a user-defined TLS
+	ConnTypeStream = 1
+	// ConnTypeStreamTLS indicates the service listens on a packetconn connection, with a user-defined TLS
+	ConnTypeStreamTLS = 2
 )
 
 // ServiceAdvertisement is the data associated with a service advertisement
--- a/pkg/netceptor/netceptor.go
+++ b/pkg/netceptor/netceptor.go
@@ -178,12 +178,19 @@ type routingUpdate struct {
 	SuspectedDuplicate uint64
 }
 
+const (
+	typePacketConn = 0
+	typeConnNonTLS = 1
+	typeConnTLS    = 2
+)
+
 // ServiceAdvertisement is the data associated with a service advertisement
 type ServiceAdvertisement struct {
-	NodeID  string
-	Service string
-	Time    time.Time
-	Tags    map[string]string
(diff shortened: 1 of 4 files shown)


pkg/netceptor/netceptor.go
```diff
@@ -178,12 +178,19 @@ type routingUpdate struct {
 	SuspectedDuplicate uint64
 }
 
+const (
+	typePacketConn = 0
+	typeConnNonTLS = 1
+	typeConnTLS    = 2
+)
```
ghjm: These consts should be public, since someone getting a ServiceAdvertisement can see a ConnType byte, and needs to know what it means.  Also, I think they should be named ConnTypeXXX rather than TypeXXX.

receptorctl/receptorctl/cli.py
```diff
@@ -97,10 +97,16 @@ def status(ctx):
...
+        print(f"{'Node':<{longest_node}} Service   Type        Last Seen            Tags")
         for ad in ads:
             time = dateutil.parser.parse(ad['Time'])
-            print(f"{ad['NodeID']:<{longest_node}} {ad['Service']:<8}  {time:%Y-%m-%d %H:%M:%S}  ", end="")
+            if ad['ConnType'] == 0:
+                conn_type = 'PacketConn'
+            elif ad['ConnType'] == 1:
+                conn_type = 'ConnNonTLS'
+            elif ad['ConnType'] == 2:
+                conn_type = 'ConnTLS'
```
ghjm: Conn and PacketConn are only meaningful to a Go developer.  For end-user-visible status, we should call these something like Datagram, Stream and TLS_Stream (or StreamTLS?).  And maybe the constants in the Go code should reflect these names, just for consistency.

ghjm:
> That means calling status from the unix socket will only display the byte (0,1, or 2), not the string, which is less useful.

I can see arguments to both (leaving it a byte, or making it a string).  I don't have a strong preference one way or the other - I think it's fine the way it is.  As long as `receptorctl status` gives you something human readable, I think we're good.

receptorctl/receptorctl/cli.py
fosterseth: renamed them to `Datagram`, `Stream`, and `StreamTLS`

pkg/netceptor/netceptor.go
fosterseth: made public, and renamed to `ConnTypeDatagram`, `ConnTypeStream`, and `ConnTypeStreamTLS`

ghjm (approved):

PR #256, opened 2020-11-17
Check receptor and receptorctl versions

fosterseth:
This PR addresses the situation where a new receptorctl is used with an older receptor (and vice versa), which may lead to incompatibility errors.

Adds a "Version" field to status output,  which represents the receptor binary app version number

e.g.
```
[sfoster@localhost sockceptor]$ nc -U /tmp/foo.sock
Receptor Control, node foo
{"command":"status","fields":"Version"}            
{"Version":"0.9.2-0.git.35.g2518d6c"}
```
Note: you can now pass "fields" into the JSON command, which will only retrieve the status fields in the specified list.

Adds a `version` command to cli.py, which grabs the version from the receptor binary, and then cross check it with the version in the python package for receptorctl.

```
[sfoster@localhost receptorctl]$ receptorctl --socket /tmp/foo.sock version
	receptorctl  0.9.2-0.git.35.gc584677
	receptor     0.9.2-0.git.36.gc584618
Error: Could not submit work
```

cli.py `status` now includes version number in its output,
```
Node ID: foo
Version: 0.9.2-0.git.35.g2518d6c

foo          control   StreamTLS  2020-11-18 16:12:04  None
foo          echo      Stream     2020-11-18 16:12:04  {'type': 'Command Service'}
foo          udp       Datagram   2020-11-18 16:12:04  {'address': 'localhost:8989', 'type': 'UDP Proxy'}
```


cmd/receptor.go
```diff
@@ -95,6 +95,7 @@ func (cfg nullBackendCfg) Run() error {
 }
 
 func main() {
+	controlsvc.Version = version
```
ghjm: Maybe what we want to do here is have a new package that holds the version resource, instead of having it be in main, so that other packages can import it without dependency loops?

pkg/controlsvc/status.go
```diff
@@ -17,17 +22,42 @@ func (t *statusCommandType) InitFromString(params string) (ControlCommand, error
...
 
 func (c *statusCommand) ControlFunc(nc *netceptor.Netceptor, cfo ControlFuncOperations) (map[string]interface{}, error) {
 	status := nc.Status()
+	precfr := make(map[string]interface{})
+	precfr["Version"] = Version
+	precfr["NodeID"] = status.NodeID
+	precfr["Connections"] = status.Connections
+	precfr["RoutingTable"] = status.RoutingTable
+	precfr["Advertisements"] = status.Advertisements
+	precfr["KnownConnectionCosts"] = status.KnownConnectionCosts
```
ghjm: This is a lot of moving around of data that maybe we're not going to use.  It probably doesn't matter that much because status probably isn't going to be called in a tight inner loop, but it would be better not to pay the runtime cost of copying a data element we know we don't need.

One way to do this would be to us a function table, like:
```
statusGetters := make(map[string]func() interface{})
statusGetters["Connections"] = func() interface{} { return status.Connections }
```

Then when assigning to cfr, you call the getter as a function, so the data copy doesn't actually happen until we know this particular item is needed.  We should also check for unknown items and ignore them rather than panic.
```
if c.fields != nil {
	for _, f := range c.fields {
		getter, ok := statusGetters[f]
		if ok {
			cfr[f] = getter()
		}
	}
}
```

receptorctl/receptorctl/cli.py
```diff
@@ -51,6 +51,8 @@ def get_rc(ctx):
 def status(ctx):
     rc = get_rc(ctx)
     status = rc.simple_command("status")
+    version = status.pop('Version')
+    print(f"Version: {version}")
```
ghjm: Not sure exactly why, but it seems aesthetically unpleasing to me to put the version before the node ID.  I guess because I think of the node ID as the label for the whole node, and the other items as properties accruing to that label.  It doesn't really matter but I'd personally rather keep the node ID as the first thing printed.

receptorctl/receptorctl/socket_interface.py
```diff
@@ -65,6 +66,23 @@ def readconfig(self, config, tlsclient):
                     self._insecureskipverify = key.get("insecureskipverify", self._insecureskipverify)
                     break
 
+    def check_version(self):
+        receptorVersion = self.simple_command('{"command":"status","requested_fields":["Version"]}')["Version"]
+        receptorctlVersion = pkg_resources.get_distribution('receptorctl').version
+        if receptorVersion != receptorctlVersion:
+            print("Warning: receptorctl and receptor are different versions, they may not be compatible")
+        else:
+            print("Versions match and should be compatible")
```
ghjm: I don't think it's necessary to print this in the case where the versions match.  It just distracts from whatever the real error is.

receptorctl/receptorctl/socket_interface.py
```diff
@@ -65,6 +66,23 @@ def readconfig(self, config, tlsclient):
                     self._insecureskipverify = key.get("insecureskipverify", self._insecureskipverify)
                     break
 
+    def check_version(self):
+        receptorVersion = self.simple_command('{"command":"status","requested_fields":["Version"]}')["Version"]
```
ghjm: Maybe skip this if self.socket is None?

ghjm (approved):

PR #259, opened 2020-11-23
Add system CPU cores and total memory installed in status output

fosterseth:
```
Node ID: foo
Version: 0.9.2-0.git.35.gcadd579
System CPU Count: 8
System Memory MB: 16651

foo          control   StreamTLS  2020-11-23 15:00:37  None
foo          echo      Stream     2020-11-23 15:00:37  {'type': 'Command Service'}
foo          udp       Datagram   2020-11-23 15:00:37  {'address': 'localhost:8989', 'type': 'UDP Proxy'}
```

We can get number of cores from the built-in go library `runtime`. There doesn't seem to be a way to get total memory from this package.

For memory I use https://github.com/pbnjay/memory

It appears to be minimal and only relies on standard library. It claims to supports windows, linux, and macOS (I only tested on linux)

It seems there is hesitance to expose system memory as a function in the go standard library, which is discussed here https://github.com/golang/go/issues/21816

### DIFF
--- a/pkg/utils/sysinfo.go
+++ b/pkg/utils/sysinfo.go
@@ -10,7 +10,7 @@ func GetSysCPUCount() int {
 	return runtime.NumCPU()
 }
 
-// GetSysMemoryMB returns the capacity (in MB) of the physical memory installed on the system
-func GetSysMemoryMB() uint {
-	return uint(memory.TotalMemory() / 1e6)
+// GetSysMemoryMiB returns the capacity (in mebibytes) of the physical memory installed on the system
+func GetSysMemoryMiB() uint64 {
+	return memory.TotalMemory() / 1048576
 }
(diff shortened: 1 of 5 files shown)


pkg/utils/sysinfo.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+// GetSysCPUCores returns number of CPU cores on the system
+func GetSysCPUCores() int {
+	return runtime.NumCPU()
+}
+
+// GetSysMemoryMB returns the capacity (in MB) of the physical memory installed on the system
+func GetSysMemoryMB() uint {
+	return uint(memory.TotalMemory() / 1e6)
```
ghjm: Maybe use `2e20` here to get binary MB?  (And call it GetSysMemoryMiB if you really want to be correct per SI.)

ghjm: Or I guess this would be `1<<20` not `2e20`.  (Or just `1048576`, it's not that many more characters.)

ghjm (approved):

PR #261, opened 2020-11-24
Fix stdout bug on k8s restart

fosterseth:
related https://github.com/project-receptor/receptor/issues/86#issuecomment-732345001

We cannot seek into the logger stream object, so if there is stdout already present (before the restart) we cannot append to the stdout file. Instead, we can rewrite the full k8s output starting from the beginning of the stdout file.

### DIFF
--- a/pkg/workceptor/stdio_utils.go
+++ b/pkg/workceptor/stdio_utils.go
@@ -27,16 +27,7 @@ type stdoutWriter struct {
 
 // newStdoutWriter allocates a new stdoutWriter, which writes to both the stdout and status files
 func newStdoutWriter(unitdir string) (*stdoutWriter, error) {
-	var osFlags int
-	stdoutPath := path.Join(unitdir, "stdout")
-	if _, err := os.Stat(stdoutPath); err == nil {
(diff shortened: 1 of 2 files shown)


pkg/workceptor/stdio_utils.go
```diff
@@ -27,7 +27,16 @@ type stdoutWriter struct {
 
 // newStdoutWriter allocates a new stdoutWriter, which writes to both the stdout and status files
 func newStdoutWriter(unitdir string) (*stdoutWriter, error) {
-	writer, err := os.OpenFile(path.Join(unitdir, "stdout"), (os.O_APPEND|os.O_CREATE)+os.O_WRONLY+os.O_SYNC, 0600)
+	var osFlags int
+	stdoutPath := path.Join(unitdir, "stdout")
+	if _, err := os.Stat(stdoutPath); err == nil {
```
fosterseth: my assumption here is that if the stdout exists at all (regardless of its content), we are safe to open without the create flag

ghjm (approved):

squidboylan (approved):

PR #263, opened 2020-11-24
fix hang when getting results for nonexistent work ID

fosterseth:
related https://github.com/project-receptor/receptor/issues/262

If `GetResults` returns with an error, writing to `doneChan` will hang because our only reader is in `GetResults`. Instead I think we can safely close the channel.


ghjm:
I think the point of the doneChan is that if the send of the results fails (for example, if the network connection to the client breaks), we terminate the goroutine that is sending data to resultsChan.  It might be that there's a race or deadlock and this scheme needs to be reworked, but I don't think it's as simple as just getting rid of doneChan entirely.

ghjm (approved):

PR #271, opened 2020-12-08
Fix test reporting

squidboylan:
https://github.com/squidboylan/receptor/actions/runs/409119313 has an example run of this with some failed tests and the new image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,9 @@ build-all:
 	go build cmd/receptor.go
 
 test: receptor
-	@go test ./... -p 1 -parallel=16 -count=1
+	@echo "Running receptor tests" && \
+	go test -v ./... -p 1 -parallel=16 -count=1 | tee test-output.log | grep FAIL ; \
+	echo "See test-output.log for a more comprehensive report"
 
 RUNTEST ?=
 ifeq ($(RUNTEST),)
@@ -162,7 +164,7 @@ receptorctl-test-venv/bin/pytest:
 	receptorctl-test-venv/bin/pip install -r receptorctl/test-requirements.txt
 
 receptorctl-tests: receptor receptorctl-test-venv/bin/pytest
-	cd receptorctl && ../receptorctl-test-venv/bin/pytest tests/tests.py
+	receptorctl-test-venv/bin/pytest --junitxml=../receptorctl-test-junit.xml receptorctl/tests/tests.py
 
 clean:
 	@rm -fv receptor receptor.exe receptor.app net $(SPECFILES)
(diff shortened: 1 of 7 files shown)


tests/functional/mesh/mesh_test.go
```diff
@@ -51,6 +51,7 @@ func TestMeshStartup(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
+			t.Fatal("Testing our failure reporting")
```
ghjm: Is this intended to be removed before merging?

ghjm (approved):

Makefile
```diff
@@ -162,7 +164,7 @@ receptorctl-test-venv/bin/pytest:
 	receptorctl-test-venv/bin/pip install -r receptorctl/test-requirements.txt
 
 receptorctl-tests: receptor receptorctl-test-venv/bin/pytest
-	cd receptorctl && ../receptorctl-test-venv/bin/pytest tests/tests.py
+	cd receptorctl && ../receptorctl-test-venv/bin/pytest --junitxml=../receptorctl-test-junit.xml tests/tests.py
```
ghjm: Perhaps this receptorctl-test-junit.xml should also be saved as an artifact from the GitHub workflow?

Makefile
```diff
@@ -162,7 +164,7 @@ receptorctl-test-venv/bin/pytest:
...
 
 clean:
 	@rm -fv receptor receptor.exe receptor.app net $(SPECFILES)
```
ghjm, line 169: Make clean should remove test-output.log, test-junit.xml and receptorctl-test-junit.xml.  These should also be added to .gitignore.

Makefile
```diff
@@ -70,7 +70,9 @@ build-all:
 	go build cmd/receptor.go
 
 test: receptor
-	@go test ./... -p 1 -parallel=16 -count=1
+	@echo "Running receptor tests" && \
+	go test -v ./... -p 1 -parallel=16 -count=1 | tee test-output.log | grep FAIL ; \
+	echo "See test-output.log or test-junit.xml for a more comprehensive report"
```
ghjm: If I understand correctly, test-junit.xml isn't being created by anything in the Makefile - it's only a thing that happens in the GitHub Actions container.  So it probably shouldn't be mentioned here.

Makefile
ghjm: Also, do receptorctl errors here get picked up by GitHub and reported in the GUI?  And are there relative paths in the file that would be broken by moving the file up a level?

squidboylan: yes the receptorctl-test-junit.xml will be saved as an artifact, it just ends up in the same zip/tar as the other logs. Also yes the thing that reports the junit in the GUI will report the receptorctl test junit as well. Not sure about the relative paths but i expect you're right, i should probably fix that by removing the cd

squidboylan:
sadly im not sure we can do this right now due to https://github.com/ScaCap/action-surefire-report/issues/31 which has to do with github and the permissions given via the `GITHUB_TOKEN` to forks. I'm still looking around but it's not looking good so far

ghjm:
😢

Issue #267, opened 2020-12-04
Allow for passing in kubeconfig as runtime param


shanemcd:
Currently it is only possible to pass in individual config items: https://github.com/project-receptor/receptor/blob/820c4155becdb8839e3025469987ad00f4e68fe4/pkg/workceptor/kubernetes.go#L608-L618

shanemcd:
I actually might retract this request. It's more secure if we never write this out to disk and therefore dont have to worry about cleaning it up.

ghjm:
@shanemcd I'm not sure what you mean about writing it to disk.  Are you talking about the Tower code or the disk writes to the status file within Receptor?

shanemcd:
@ghjm The former. Right now we write out a kubeconfig file and have to go to great lengths to make sure it's cleaned up. But your question raises the same concern. Will this status file always be cleaned up when the work is "released"?

ghjm:
Yes, the status file is in a temporary directory specific to the unit of work, and the "release" operation deletes this directory.  (Of course this assumes "release" is called.  If a work unit is never released then its files are never deleted.)

squidboylan:
@fosterseth  added testing for this as part of his PR, closing

PR #272, opened 2020-12-08
Allow passing in kubeconfig path at runtime

fosterseth:
related #267 
e.g. 
`receptorctl --socket /tmp/foo.sock work submit kubeit --param kube_config=/home/sfoster/.kube2/config -n`


ghjm (approved):

ghjm:
Well this didn't turn out to be too complicated, did it?

PR #276, opened 2020-12-15
Allow passing in podspec during runtime

fosterseth:
e.g. `receptorctl ... --param secret_kube_podspec=@/home/sfoster/sockceptor/podspec`

podspec can be in `yaml` or `json` formats

If podspec is passed in, it will overwrite the `command`, `image` and `params` defined in the `work-kubernetes`.

As such, the `allowruntimecommand` and `allowruntimeparams` must be set to true

```yaml
- work-kubernetes:
    worktype: kubeit
    authmethod: kubeconfig
    namespace: sbf
    allowruntimecommand: true
    allowruntimeparams: true
```

The podspec must specify a container named "worker", with `stdin` and `stdinOnce` set to True. This is the container that we get stdin/stdout to and from.

Podspec can also be a config option

```yaml
- work-kubernetes:
    worktype: kubeit2
    authmethod: kubeconfig
    namespace: default
    podspec: /home/sfoster/sockceptor/podspec.yml
```

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -11,6 +11,7 @@ import (
 	"github.com/project-receptor/receptor/pkg/cmdline"
 	"github.com/project-receptor/receptor/pkg/logger"
 	"io"
+	"io/ioutil"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -46,7 +47,6 @@ type kubeUnit struct {
 	allowRuntimeCommand bool
 	allowRuntimeParams  bool
 	deletePodOnRestart  bool
-	kubeConfig          string
 	namePrefix          string
 	config              *rest.Config
 	clientset           *kubernetes.Clientset
@@ -109,20 +109,15 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
 	spec := corev1.PodSpec{}
 	if ked.KubePodSpec != "" {
 		reader := bytes.NewReader([]byte(ked.KubePodSpec))
-		decoder := yaml.NewYAMLOrJSONDecoder(reader, 1024)
-		err := decoder.Decode(&spec)
+		err := yaml.NewYAMLOrJSONDecoder(reader, 1024).Decode(&spec)
 		if err != nil {
 			return err
 		}
 		foundWorker := false
 		for _, container := range spec.Containers {
 			if container.Name == "worker" {
-				if !container.Stdin {
-					container.Stdin = true
-				}
-				if !container.StdinOnce {
-					container.StdinOnce = true
-				}
+				container.Stdin = true
+				container.StdinOnce = true
 				foundWorker = true
 				break
 			}
@@ -130,14 +125,7 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
 		if !foundWorker {
 			return fmt.Errorf("At least one container must be named worker")
 		}
-		if spec.RestartPolicy != corev1.RestartPolicyNever {
-			spec.RestartPolicy = corev1.RestartPolicyNever
-		}
-		kw.UpdateFullStatus(func(status *StatusFileData) {
-			status.ExtraData.(*kubeExtraData).Image = ""
-			status.ExtraData.(*kubeExtraData).Params = ""
-			status.ExtraData.(*kubeExtraData).Command = ""
-		})
+		spec.RestartPolicy = corev1.RestartPolicyNever
 	} else {
 		spec = corev1.PodSpec{
 			Containers: []corev1.Container{{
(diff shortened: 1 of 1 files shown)


fosterseth:
@ghjm having `--param kube_image` override whatever is in podspec is a little tricky because podspec could have more than one container. I think it's better to just throw an error if they try to provide both of these params at once.

pkg/workceptor/kubernetes.go
```diff
@@ -103,12 +106,40 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
...
+			}
+		}
+		if !foundWorker {
+			return fmt.Errorf("At least one container must be named worker")
+		}
+		if spec.RestartPolicy != corev1.RestartPolicyNever {
+			spec.RestartPolicy = corev1.RestartPolicyNever
+		}
+		kw.UpdateFullStatus(func(status *StatusFileData) {
+			status.ExtraData.(*kubeExtraData).Image = ""
```
ghjm: Aren't we already guaranteeing these are empty with the error checking on submission?

pkg/workceptor/kubernetes.go
```diff
@@ -606,6 +648,11 @@ func (kw *kubeUnit) SetFromParams(params map[string]string) error {
 			}
 		}
 	}
+	if ked.KubePodSpec != "" && (userParams != "" || userCommand != "" || userImage != "") {
+		return fmt.Errorf("params kube_command, kube_image, kube_params not compatible with secret_kube_podspec")
```
ghjm, line 659: Probably need to check that `kw.baseParams` is also empty

pkg/workceptor/kubernetes.go
```diff
@@ -103,12 +106,40 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
...
+		reader := bytes.NewReader([]byte(ked.KubePodSpec))
+		decoder := yaml.NewYAMLOrJSONDecoder(reader, 1024)
+		err := decoder.Decode(&spec)
+		if err != nil {
+			return err
+		}
+		foundWorker := false
+		for _, container := range spec.Containers {
+			if container.Name == "worker" {
+				if !container.Stdin {
```
ghjm: There's no need to check this before setting it.  This is just a boolean variable - there ere are no side effects from setting it.

pkg/workceptor/kubernetes.go
fosterseth: these will contain the image/command strings from the `worker-kubernetes` definition

ghjm: Hmm ... I wonder if we should make this an error as well?

And the more I think about it, the more I think there ought to be an option for a statically configured pod spec.  To avoid the yaml-inside-yaml marshalling issue, we could have the config option take a filename that contains the pod spec yaml.

pkg/workceptor/kubernetes.go
fosterseth: This check makes sure podspec isn't passed in alongside the other runtime options, but it's okay to have these (image, command, params) still defined in the `worker-kubernetes` definition in the yaml.

But it does make me realize that I probably should set kw.baseParams to be empty, given it won't be used at all.

fosterseth:
here are the rules for podspec
if podspec is in the yaml file, then image/command/params cannot be set in that yaml definition.
if podspec is passed in at runtime, image/command/params cannot also be passed in at runtime.

pkg/workceptor/kubernetes.go
```diff
@@ -557,6 +581,17 @@ func (kw *kubeUnit) connectToKube() error {
 	return nil
 }
 
+func readFileToString(filename string) (string, error) {
+	if filename == "" {
```
ghjm, line 586: In the unlikely event that some future person stumbles across this function and uses it for something, the "return an empty result if given an empty filename" behavior is pretty unexpected.  I think we should at least call it out with a comment.

pkg/workceptor/kubernetes.go
```diff
@@ -606,7 +654,17 @@ func (kw *kubeUnit) SetFromParams(params map[string]string) error {
 			}
 		}
 	}
-	ked.Params = combineParams(kw.baseParams, userParams)
+	if userPodSpec != "" && (userParams != "" || userCommand != "" || userImage != "") {
+		return fmt.Errorf("params kube_command, kube_image, kube_params not compatible with secret_kube_podspec")
+	}
+	ked.Command = userCommand
```
ghjm: For clarity of which things are happening in which circumstance, what do you think of arranging this more like:

```
	if userPodSpec == "" {
		ked.Command = userCommand
		ked.Image = userImage
		ked.Params = combineParams(kw.baseParams, userParams)
	} else {
		ked.KubePodSpec = userPodSpec
		kw.baseParams = ""
	}
```

fosterseth: yep that is clearer, fixed!

pkg/workceptor/kubernetes.go
fosterseth: added a comment

ghjm (approved):

PR #281, opened 2020-12-22
Receptor TLS improvements

ghjm:
- [x]  Use x.509 OtherName for Receptor node IDs so we can distinguish between them and DNS names
- [x]  Update tests to generate and work with the new certificates
- [x]  Add a command line CA that generates the new certificate types
- [ ]  Integrate the CA into the Receptor mesh so nodes can join, be approved, self-renew, etc


ghjm:
This now has the first three items working: above-the-mesh TLS now uses Receptor-specific names that cannot be mixed up with DNS names, there are new `cert-*` commands that allow you to create a CA and generate and sign Receptor-style certificates, and the tests have been updated to use the new certificate generation functions.

There is as yet no online CA service offered via the Receptor mesh.  I'm thinking perhaps that would be better handled as a separate PR subsequent to this one.

pkg/cmdline/cmdline.go
```diff
@@ -84,32 +83,64 @@ func printableTypeName(typ reflect.Type) string {
 	} else if typ.Kind() == reflect.Map {
 		return fmt.Sprintf("JSON dict of %s to %s", printableTypeName(typ.Key()), printableTypeName(typ.Elem()))
 	} else if typ.Kind() == reflect.Slice {
+		if typ.Elem() == reflect.TypeOf("") {
+			return fmt.Sprintf("%s (may be repeated)", typ.String())
```
fosterseth, line 87: does this mean comma-separated?
e.g.
`dnsname: foo.com,foo2.com`

pkg/certificates/cli.go
```diff
@@ -0,0 +1,179 @@
...
+	"net"
+	"strings"
+	"time"
+)
+
+type initCA struct {
+	CommonName string `description:"Common name to assign to the certificate" required:"Yes"`
+	Bits       int    `description:"Bit length of the encryption keys of the certificate" required:"Yes"`
+	NotBefore  string `description:"Effective (NotBefore) date/time, in RFC3339 format"`
+	NotAfter   string `description:"Expiration (NotAfter) date/time, in RFC3339 format"`
```
fosterseth, line 19: not too critical, but we could provide an example of RFC3339 format,
`Effective (NotBefore) date/time, in RFC3339 format, e.g. 2022-07-04T00:00:00.00Z`

pkg/cmdline/cmdline.go
ghjm: No, it means you can give the option more than once, adding a single string each time, and the cmdlibrary will build a slice of strings that it passes to the app.  So you can do `receptor --cert-makereq [...] dnsname=foo.com dnsname=bar.com`

pkg/certificates/cli.go
ghjm: My only concern about this is that I've been generally trying to keep things so that when you do `receptor --help` it looks good on a standard size (80x25) terminal.  (I haven't always succeeded.)

squidboylan (approved):

Issue #283, opened 2021-01-05
Do not require image w/ work-kubernetes when using allowruntimeparams


shanemcd:
With the following config:

```
- work-kubernetes:
    worktype: ocp
    namespace: awx
    authmethod: kubeconfig
    allowruntimeauth: true
    allowruntimeparams: true
```

I am seeing the error:

```
$ ./receptor --config ~/awx/tools/docker-compose/receptor.conf
Error: must specify a container image to run
```

I want to pass this only via runtime pod spec.

ghjm:
Related to this: we should allow a runtime-only kubeconfig mode, perhaps `authmethod: runtime`, where we suppress trying to load a kubeconfig from the default location and fail if one isn't passed in at runtime.

PR #284, opened 2021-01-07
Add tests for passing in k8s podspec and kubeconfig at runtime

fosterseth:
#283 #280 
- Also adds a new authmethod called `runtime`.
- Adds a new param permission called `allowruntimepodspec` which must be true if podspec is to be passed in during runtime.

If authmethod is `runtime`, then `secret_kube_config` must be passed at runtime, otherwise the job will fail. With authmethod `kubeconfig` (the older way, which is still an option), it's possible for the system to find and load a default kubeconfig on the system (e.g. `/home/sfoster/.kube/config`), which could lead to unintended side effects.

here is a an example work-kubernetes definition with the changes, and issuing a `receptorctl` command to target this definition.
```yaml
- work-kubernetes:
    worktype: kubeit2
    authmethod: runtime
    namespace: default
    allowruntimeauth: true
    allowruntimepodspec: true
```
`receptorctl --socket /tmp/foo.sock work submit kubeit2 -n --param secret_kube_podspec=@/home/sfoster/sockceptor/podspec.yml --param secret_kube_config=@/home/sfoster/.kube/config`

### DIFF
--- a/tests/functional/mesh/work_test.go
+++ b/tests/functional/mesh/work_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/project-receptor/receptor/tests/functional/lib/mesh"
 	"github.com/project-receptor/receptor/tests/functional/lib/receptorcontrol"
 	"github.com/project-receptor/receptor/tests/functional/lib/utils"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
@@ -58,7 +59,6 @@ func TestWork(t *testing.T) {
 			"command":    "sh -c \"for i in `seq 1 5`; do echo $i;done\"",
 		},
 	}
-
 	testTable := []struct {
 		testGroup    string
 		shortCommand map[interface{}]interface{}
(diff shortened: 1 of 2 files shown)


tests/functional/mesh/work_test.go
```diff
@@ -591,6 +591,62 @@ func TestRuntimeParams(t *testing.T) {
 	}
 }
 
+func TestKubeRuntimeParams(t *testing.T) {
```
squidboylan, line 594: we should run the `checkSkipKube` in this test which skips a test if we set `SKIP_KUBE` in our environment

squidboylan (commented):
Looks mostly right to me except for one inline comment i left

tests/functional/mesh/work_test.go
fosterseth: thanks, fixed

squidboylan (approved):

ghjm:
In `connectUsingKubeconfig`, we call `NewDefaultClientConfigLoadingRules` to get a kube api object that can search for a kubeconfig in the default places.  With authmethod=runtime we shouldn't do this.  In the function this only gets used if `ked.KubeConfig == ""` .  Are we 100% confident that this could never be true under runtime kubeconfig auth?  For example, if a user intentionally passes in an empty string?

If we're not 100% confident, then I think we need to add a boolean parameter to `connectUsingKubeconfig` to suppress the search behavior when using runtimeonly auth.  This could be a security issue because who knows if the user has a higher-privileged kubeconfig lying around in the account Receptor happens to be running under.

fosterseth:
@ghjm 

I believe we are covered here, in `SetFromParams`

```golang
if kw.authMethod == "runtime" && ked.KubeConfig == "" {
	return fmt.Errorf("param secret_kube_config must be provided if AuthMethod=runtime")
}
```

ghjm (approved):
Ok, looks good then!

PR #285, opened 2021-01-10
Allow for passing full Pod definition at runtime

shanemcd:
I should have looked more closely at https://github.com/project-receptor/receptor/pull/276 before we merged it. For the AWX use case we need to be able to pass in a full `Pod` definition, not just the `spec:` field. This was my fault, because I was used the term 'pod spec' when making this request, without knowing that it was its own distinct object type.

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -3,22 +3,26 @@
 package workceptor
 
 import (
-	"bytes"
 	"context"
 	"encoding/pem"
 	"fmt"
+	"io"
+	"io/ioutil"
+	"net"
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+
 	"github.com/google/shlex"
 	"github.com/project-receptor/receptor/pkg/cmdline"
 	"github.com/project-receptor/receptor/pkg/logger"
-	"io"
-	"io/ioutil"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -27,11 +31,6 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/remotecommand"
 	watch2 "k8s.io/client-go/tools/watch"
-	"net"
-	"os"
-	"strconv"
-	"strings"
-	"sync"
 )
 
 // kubeUnit implements the WorkUnit interface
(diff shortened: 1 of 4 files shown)


pkg/workceptor/kubernetes.go
```diff
@@ -107,18 +106,20 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
...
 		foundWorker := false
-		for _, container := range spec.Containers {
-			if container.Name == "worker" {
-				container.Stdin = true
-				container.StdinOnce = true
+		spec = &pod.Spec
+		for i := range spec.Containers {
+			if spec.Containers[i].Name == "worker" {
+				spec.Containers[i].Stdin = true
+				spec.Containers[i].StdinOnce = true
```
shanemcd, line 122: I would appreciate if someone with more Go knowledge could explain why I needed to do this when I changed how `spec` is defined. With `for _, container := range spec.Containers {`, the changes did not persist when used below on line 159.

pkg/workceptor/kubernetes.go
```diff
@@ -127,8 +128,16 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
 			return fmt.Errorf("At least one container must be named worker")
 		}
 		spec.RestartPolicy = corev1.RestartPolicyNever
+		userNamespace := pod.ObjectMeta.Namespace
+		if userNamespace != "" {
+			ked.KubeNamespace = userNamespace
+		}
+		userPodName := pod.ObjectMeta.Name
+		if userPodName != "" {
+			kw.namePrefix = userPodName + "-"
```
shanemcd, line 137: I am not sure about this. If a `name` is passed as part of the Pod spec, I'm taking that and using it for `GenerateName`.

pkg/workceptor/kubernetes.go
```diff
@@ -127,8 +128,16 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
 			return fmt.Errorf("At least one container must be named worker")
 		}
 		spec.RestartPolicy = corev1.RestartPolicyNever
+		userNamespace := pod.ObjectMeta.Namespace
+		if userNamespace != "" {
```
shanemcd, line 132: This allows for omitting the `namespace` attribute in the receptor config. It will also override any value present in the config.

pkg/workceptor/kubernetes.go
shanemcd: I'm really stumped by this one. Everything I'm reading online is saying that `range` will create a copy of the items in the slice it's iterating over. How was this working before? 😕

fosterseth: yeah this was a bug, and I think passing in Stdin wouldn't have worked before (seems our tests don't use stdin). How you do it here seems like the right way to go.

shanemcd: Yep, came here to say this. Thanks!

fosterseth:
I feel like we should rename our internal variables from "\*podspec\*" (e.g. `ked.KubePodSpec`) to "\*pod\*" with this change.

shanemcd:
Sure, I'll whip that up now.

pkg/workceptor/kubernetes.go
```diff
@@ -804,14 +804,14 @@ type WorkKubeCfg struct {
 	Params              string `description:"Command-line parameters to pass to the entrypoint"`
 	AuthMethod          string `description:"One of: kubeconfig, incluster" default:"incluster"`
 	KubeConfig          string `description:"Kubeconfig filename (for authmethod=kubeconfig)"`
-	PodSpec             string `description:"Podspec filename"`
+	Pod                 string `description:"Podspec filename"`
```
fosterseth: maybe "Pod definition filename, in json or yaml format"

fosterseth (approved):

PR #298, opened 2021-01-28
Prevent hangs when work-kubernetes encounters an error while sending stdin

shanemcd:
When using a service account without the appropriate permissions, the unit of work was hanging forever with the status `Sending stdin to pod`.

After some spelunking, I discovered that the `streamWait` WaitGroup was never finishing because we never receive any stdout.

This PR attempts to solve the problem by utilizing a separate WaitGroup for both stdin and stdout.

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -338,34 +338,37 @@ func (kw *kubeUnit) runWorkUsingLogger() {
 	// Actually run the streams.  This blocks until the pod finishes.
 	var errStdin error
 	var errStdout error
-	streamWait := sync.WaitGroup{}
-	streamWait.Add(2)
+	stdinWait := sync.WaitGroup{}
+	stdinWait.Add(1)
+	stdoutWait := sync.WaitGroup{}
+	stdoutWait.Add(1)
 	if skipStdin {
-		streamWait.Done()
+		stdinWait.Done()
 	} else {
 		go func() {
 			errStdin = exec.Stream(remotecommand.StreamOptions{
 				Stdin: stdin,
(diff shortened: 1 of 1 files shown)


pkg/workceptor/kubernetes.go
```diff
@@ -338,34 +338,37 @@ func (kw *kubeUnit) runWorkUsingLogger() {
...
 	}()
-	streamWait.Wait()
+	stdinWait.Wait()
+	if errStdin != nil {
+		var errDetail string
+		errDetail = fmt.Sprintf("%s", errStdin)
+		kw.UpdateBasicStatus(WorkStateFailed, fmt.Sprintf("Stdin error running pod: %s", errDetail), stdout.Size())
+		return
+	}
+	stdoutWait.Wait()
```
fosterseth, line 367: I am wondering if we are potentially leaking a `io.Copy` go-routine here, since we are returning before `stdoutWait.Wait()` is called.

Maybe we could start the stdout copy routine only after the stdin is passed in successfully? Or do these need to be concurrent.

I think to close that io.copy routine, we can close either the `logStream` or stdout connection.

We ran into a similar issue before when trying to cancel the io.Copy operation, and Graham came up with this solution
https://github.com/project-receptor/receptor/blob/fed9f8698b967f513ed48db36bf88261e522d792/pkg/workceptor/remote_work.go#L391

shanemcd: This is good feedback. Thank you. I'll look at it more closely next week.

shanemcd:
Closing in favor of https://github.com/project-receptor/receptor/pull/300

PR #300, opened 2021-02-04
Close logStream if Stdin is not passed correctly, to prevent hang

fosterseth:
If `errStdin` is not `nil`, we need to close out of the `io.Copy` go routine. To do this, we can simply close the `logStream` connection.

see https://github.com/project-receptor/receptor/pull/298 for more details about the scenario that causes the hang.


shanemcd (approved):
It works! Thanks @fosterseth

PR #301, opened 2021-02-05
include an array of available work commands in service announcements

ryanpetrello:
related: https://github.com/project-receptor/receptor/issues/21

### DIFF
--- a/pkg/netceptor/netceptor.go
+++ b/pkg/netceptor/netceptor.go
@@ -23,7 +23,7 @@ import (
 	"time"
 )
 
-// defaultMTU is the largest message sendable over the Netecptor network
+// defaultMTU is the largest message sendable over the Netceptor network
 const defaultMTU = 16384
 
 // defaultRouteUpdateTime is the interval at which regular route updates will be sent
@@ -88,6 +88,7 @@ type Netceptor struct {
 	maxForwardingHops      byte
 	maxConnectionIdleTime  time.Duration
 	allowedPeers           []string
+	workCommands           []string
 	epoch                  uint64
 	sequence               uint64
 	connLock               *sync.RWMutex
@@ -199,11 +200,12 @@ const (
 
 // ServiceAdvertisement is the data associated with a service advertisement
 type ServiceAdvertisement struct {
-	NodeID   string
-	Service  string
-	Time     time.Time
-	ConnType byte
-	Tags     map[string]string
+	NodeID       string
+	Service      string
+	Time         time.Time
+	ConnType     byte
+	Tags         map[string]string
+	WorkCommands []string
 }
 
 // serviceAdvertisementFull is the whole message from the network
@@ -497,11 +499,12 @@ func (s *Netceptor) addLocalServiceAdvertisement(service string, connType byte, 
 		s.serviceAdsReceived[s.nodeID] = n
 	}
 	n[service] = &ServiceAdvertisement{
-		NodeID:   s.nodeID,
-		Service:  service,
-		Time:     time.Now(),
-		ConnType: connType,
-		Tags:     tags,
+		NodeID:       s.nodeID,
+		Service:      service,
+		Time:         time.Now(),
+		ConnType:     connType,
+		Tags:         tags,
+		WorkCommands: s.workCommands,
 	}
 	s.sendServiceAdsChan <- 0
 }
@@ -555,11 +558,12 @@ func (s *Netceptor) sendServiceAds() {
 	for sn := range s.listenerRegistry {
 		if s.listenerRegistry[sn].advertise {
 			sa := ServiceAdvertisement{
-				NodeID:   s.nodeID,
-				Service:  sn,
-				Time:     time.Now(),
-				ConnType: s.listenerRegistry[sn].connType,
-				Tags:     s.listenerRegistry[sn].adTags,
+				NodeID:       s.nodeID,
+				Service:      sn,
+				Time:         time.Now(),
+				ConnType:     s.listenerRegistry[sn].connType,
+				Tags:         s.listenerRegistry[sn].adTags,
+				WorkCommands: s.workCommands,
 			}
 			ads = append(ads, sa)
 		}
@@ -728,6 +732,25 @@ func (s *Netceptor) GetServerTLSConfig(name string) (*tls.Config, error) {
 		return nil, fmt.Errorf("unknown TLS config %s", name)
 	}
 	return sc.Clone(), nil
+}
+
+// AddWorkCommand records a work command so it can be included in service announcements
+func (s *Netceptor) AddWorkCommand(command string) error {
+	if command == "" {
+		return fmt.Errorf("must provide a name")
+	}
+	s.serviceAdsLock.Lock()
+	defer s.serviceAdsLock.Unlock()
+	if n, ok := s.serviceAdsReceived[s.NodeID()]; ok {
+		// if it's the local node, just update the local service advertisement
+		// structs directly
+		for _, ad := range n {
+			ad.WorkCommands = append(ad.WorkCommands, command)
+		}
+	} else {
+		s.workCommands = append(s.workCommands, command)
+	}
+	return nil
 }
 
 // SetServerTLSConfig stores a server TLS config by name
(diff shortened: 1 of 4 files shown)


pkg/netceptor/netceptor.go
```diff
@@ -730,6 +734,15 @@ func (s *Netceptor) GetServerTLSConfig(name string) (*tls.Config, error) {
 	return sc.Clone(), nil
 }
 
+// AddWorkCommand records a work command so it can be included in service announcements
+func (s *Netceptor) AddWorkCommand(command string) error {
+	if command == "" {
```
ryanpetrello, line 739: (WIP note)

One thing I'm unclear on @fosterseth @squidboylan is if we need a mutex around this.  I don't totally grok receptor internals yet, so I'm unsure if the registration of the work commands in `RegisterWorker` happens in parallel goroutines.  I see other fields on the `Netceptor` struct in this file seem to follow the pattern of locking access w/ a field-specific mutex, so I may need to do so here as well for safety.

receptorctl/receptorctl/cli.py
```diff
@@ -104,7 +104,7 @@ def status(ctx):
     ads = status.pop('Advertisements', None)
     if ads:
         print()
-        print(f"{'Node':<{longest_node}} Service   Type       Last Seen            Tags")
+        print(f"{'Node':<{longest_node}} Service   Type       Last Seen            Work Commands                Tags")
```
ryanpetrello: If you've got a ton of work commands defined, this table may not print the headers very pretty.

But I also don't feel like implementing complicated ascii table drawing (given that this is just a CLI tool for humans).

pkg/netceptor/netceptor.go
```diff
@@ -199,11 +200,12 @@ const (
...
-	Service  string
-	Time     time.Time
-	ConnType byte
-	Tags     map[string]string
+	NodeID       string
+	Service      string
+	Time         time.Time
+	ConnType     byte
+	Tags         map[string]string
+	WorkCommands []string
```
ryanpetrello, line 208: I may expand this PR further to include some metadata about capacity and receptor version in this struct as part of the advertisements.  We already track this data per receptor and make it available via the status command, but I suppose we could just as easily attach it here.

pkg/netceptor/netceptor.go
squidboylan: Yeah I don't know enough about receptor internals anymore to know either

squidboylan (commented):
It would be good to have a test for this but we probably dont need to block this PR on that.

pkg/netceptor/netceptor.go
fosterseth: I think this is safe. RegisterWorker is called one at a time on each `work-command` in the yaml.

fosterseth:
I can see the work commands for node bar, but not for foo. I am connected to `/tmp/foo.sock`
```
bar          control   StreamTLS  2021-02-08 16:31:25  sleepcat,100cat,echosleepshortNone
foo          control   StreamTLS  2021-02-08 16:31:49  None
foo          echo      Stream     2021-02-08 16:31:49  {'type': 'Command Service'}
foo          udp       Datagram   2021-02-08 16:31:49  {'address': 'localhost:8989', 'type': 'UDP Proxy'}
```

receptorctl/receptorctl/cli.py
fosterseth: maybe we can just put Work Commands to the far right, so if it is really long it won't crash in to the Tags column

ryanpetrello:
@fosterseth are all the nodes running the receptor from this PR?

fosterseth:
@ryanpetrello yep, and if I connect to /tmp/bar.sock I can see foo's work commands, but not bar's

```
Node         Service   Type       Last Seen            Work Commands                Tags
bar          control   StreamTLS  2021-02-08 17:02:47  None
foo          control   StreamTLS  2021-02-08 17:02:33  sleepcat,100cat,testparam,idk,echosleepshort,kubeit2None
```
is it just me? I can't think of why this would be happening in the first place

ryanpetrello:
No, I can reproduce what you're describing - good find.

fosterseth:
@ryanpetrello my suspicion is that we are calling this https://github.com/project-receptor/receptor/blob/03f02eab9b490a3381e7cdc9f291083c083843cc/pkg/netceptor/netceptor.go#L491 before we are registering the local work commands

pkg/netceptor/netceptor.go
ryanpetrello: Well, I wrapped it w/ a service ad lock just to be sure.

pkg/netceptor/netceptor.go
```diff
@@ -730,6 +734,25 @@ func (s *Netceptor) GetServerTLSConfig(name string) (*tls.Config, error) {
...
 
+// AddWorkCommand records a work command so it can be included in service announcements
+func (s *Netceptor) AddWorkCommand(command string) error {
+	if command == "" {
+		return fmt.Errorf("must provide a name")
+	}
+	s.serviceAdsLock.RLock()
+	defer s.serviceAdsLock.RUnlock()
+	if n, ok := s.serviceAdsReceived[s.NodeID()]; ok {
+		// if it's the local node, just update the local service advertisement
```
ryanpetrello, line 745: @fosterseth this should address the problem you noticed; if the netceptor in question is **local** node, I just modify the local advertisement object explicitly (since it'll never get changed again)

Otherwise for *remote* nodes, I store the record on the netceptor object so that every subsequent remote advertisement that comes in gets updated.

receptorctl/receptorctl/cli.py
ryanpetrello: image

pkg/netceptor/netceptor.go
ryanpetrello: For posterity:

image

fosterseth (approved):

pkg/netceptor/netceptor.go
```diff
@@ -730,6 +734,25 @@ func (s *Netceptor) GetServerTLSConfig(name string) (*tls.Config, error) {
...
+	}
+	s.serviceAdsLock.Lock()
+	defer s.serviceAdsLock.Unlock()
+	if n, ok := s.serviceAdsReceived[s.NodeID()]; ok {
+		// if it's the local node, just update the local service advertisement
+		// structs directly
+		for _, ad := range n {
+			ad.WorkCommands = append(ad.WorkCommands, command)
+		}
+	} else {
```
fosterseth, line 750: trying to wrap my head around this

This function will only update the workcommands that are defined _locally_. Therefore I don't think the distinction between local / remote make much sense (i.e. I don't think this `else` ever hits)

If that is the case, we don't need a` workCommands [] string` on the netceptor struct -- it doesn't actually do anything anymore. We only need it on the `ServiceAdvertisement` struct since we are updating it directly via this `AddWorkCommand` function.

ryanpetrello: This is a little complicated because each receptor itself generates a "local" advertisement for the control service itself in-process at startup.

Let's say we have (via `tcp-peer`):

`A <------- B`

When receptor `A` runs this `AddWorkCommand` function, it'll fall into the local part of the `if` block and update the "local advertisement" object that was auto-created when the process started up as part of the `listen()` call.

When receptor `B` runs this code, each work command will append to the list of work commands on the private `workCommands` field of the netceptor object.  *This* field is what gets looked at when `B` composes the advertisement it will *send* to `A` (momentarily):

https://github.com/project-receptor/receptor/pull/301/files#diff-23f6be7c946bcb25e244cc53843ea9dfff1ba921dfb1b2645a619c060aa6bd62R566

So `netceptor.workCommands` is a slice of known work types that will be appended to all announcements that are sent _out_.

`netceptor.serviceAdsReceived[NODE].WorkCommands` are the actual values *in* the announcement object (regardless of which process it actually originated in).

ryanpetrello:
@squidboylan 👋 I have added a test, please go easy on me.

squidboylan:
> @squidboylan wave I have added a test, please go easy on me.

This is great! Thanks!

squidboylan (approved):

pkg/netceptor/netceptor.go
ryanpetrello: After chatting w/ @matburt, I think capacity and versioning don't belong here, because they're not really concepts below the mesh/at the netceptor level.  I'll revisit how we want to deal with this as we continue to experiment with integration w/ AWX in the coming weeks.

PR #306, opened 2021-02-24
add vendor/ from go mod vendor

fosterseth:
see https://github.com/project-receptor/receptor/pull/303


Spredzy:
Can be abandoned. https://github.com/project-receptor/receptor/pull/303 is now passing tests.

PR #308, opened 2021-03-02
gitignore: Only ignore top level net folder

Spredzy:
Currently all net folders are ignored by git. This leads to unexpected
issues. The expected intent was to only ignore the root net file, so
let's do that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@ receptor
 receptor.exe
 receptor.app
 recepcert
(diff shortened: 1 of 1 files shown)


ghjm (approved):
Does my review still mean anything?

rooftopcellist:
It means something to me @ghjm :smile:

rooftopcellist (approved):

PR #309, opened 2021-03-20
Move cmdline to an external dependency

ghjm:
Hello, Receptor people!  I have an upcoming project where I want to use the cmdline library, so I have copied it to a new repo at `github.com/ghjm/cmdline`.  This PR updates Receptor to import cmdline from the new repo.

A few changes to the cmdline library have been taken into account:

* Instead of accessing global variables in the module, you now allocate a struct with `NewCmdline()`.  But you can get a singleton global instance with `GlobalInstance()`.
* Instead of presuming cmdline has permission to print to stdout and exiting `os.Exit(1)`, the library now allows configuring the output stream, and returns errors in an `error` variable like a properly behaved library.
* Instead of `ParseAndRun()` returning a string identifying the exclusive command that ran, it now returns the error, and you can call `WhatRan()` to find out if an exclusive command ran.


cmd/receptor.go
```diff
@@ -83,11 +83,16 @@ func (cfg nullBackendCfg) Run() error {
...
-	cmdline.AddConfigType("version", "Show the Receptor version", version.CmdlineCfg{}, cmdline.Exclusive)
-	whatRan := cmdline.ParseAndRun(os.Args[1:], []string{"Init", "Prepare", "Run"})
-	if whatRan != "" {
+	cl := cmdline.GlobalInstance()
+	cl.AddConfigType("node", "Node configuration of this instance", nodeCfg{}, cmdline.Required, cmdline.Singleton)
+	cl.AddConfigType("local-only", "Run a self-contained node with no backends", nullBackendCfg{}, cmdline.Singleton)
+	cl.AddConfigType("version", "Show the Receptor version", version.CmdlineCfg{}, cmdline.Exclusive)
+	err := cl.ParseAndRun(os.Args[1:], []string{"Init", "Prepare", "Run"}, cmdline.ShowHelpIfNoArgs)
+	if err != nil {
+		fmt.Printf("Error: %s\n", err)
```
ghjm, line 92: I'm trying not to break existing behavior too much, but this should arguably be:
```
_, _ = fmt.Fprintf(os.Stderr, "Error: %s\n", err)
```
so that errors go to stderr.

ghjm:
If y'all decide to merge this PR, I am happy to add a `dont_break_receptor` CI check to the cmdline repo.

shanemcd:
@ghjm I am good with this, but would very much appreciate that CI check you mentioned. Can you set that up?

ghjm:
> @ghjm I am good with this, but would very much appreciate that CI check you mentioned. Can you set that up?

Yes, as I said, I'm happy to set that up.  What I'm thinking of is a GitHub Action on the ghjm/cmdline repo that checks out Receptor, updates Receptor's vendored cmdline, and then runs some builds and tests.  It would be hard to do this without first merging this PR, because current Receptor imports its own pkg/cmdline, not ghjm/cmdline.

shanemcd (approved):

PR #310, opened 2021-03-20
Use explicit version to avoid warning about mutable ref

ghjm:
This will avoid the warning
```
[WARNING] The 'rev' field of repo 'https://github.com/dnephin/pre-commit-golang' appears to be a mutable reference
```
during the pre-commit CI run.


ghjm:
Looks like someone else did this

PR #316, opened 2021-05-04
Adding a Context Timeout, for pods stuck in 'pending' State

### LINKED ISSUE #315: k8s image pull error pod stays in pending
#### To reproduce
- start a minikube server locally
- run receptor with including the `work-kubernetes` entry (below)
- submit work to receptor that points to a pod yaml with incorrect image name
  - `receptorctl --socket /tmp/foo.sock work submit kubeit2 --param secret_kube_pod=@/home/sbf/sockceptor/pod.yml --param secret_kube_config=@/home/sbf/.kube/config --no-payload`
- run `work list` command, you will see job stuck in pending state indefinitely

#### Where the problem occurs

The hang occurs here https://github.com/project-receptor/receptor/blob/e8a789ee8c2b684f0368df1f770d338aaf7002a1/pkg/workceptor/kubernetes.go#L73

We stay in this method until either the pod fails, succeeds, or enters running. We currently don't have any conditions to exit if stuck in pending.

I believe there could be a variety of reasons a pod stays in pending, not just image pull error. Here is list of go events to give you an idea of the various failure reasons (some of these may error out the pod, some might not)
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/events/event.go

Therefore, we probably shouldn't just target Image pull error specifically, but a more general approach to any reason the pod might stay in pending.

#### Solution

We could pass a timeout context into `UntilWithSync`

Additionally, if we timeout, I am wondering if there is a way to get some relevant info back to awx controller. I don't think a generic "timed out" message is good enough. Ideally we'd say something about why the pod is not starting and stick it in the `Detail` field. This will be the tricky part.

=====================================================
in receptor configuration yaml
```yaml
- work-kubernetes:
    worktype: kubeit2
    authmethod: runtime
    allowruntimeauth: true
    allowruntimepod: true
```

pod.yml
```yaml
apiVersion: v1
kind: Pod
metadata:
    generateName: myapp-pod-
    labels:
        app: myapp
spec:
    containers:
    - name: worker
      image: busyboxwrongname
      command: ['sh', '-c', 'echo The Pod is running && sleep 6 && exit 0']
    restartPolicy: Never
```

amolgautam25:
- [x] Added a context with Timeout ```ctxWithTimeout```
- [x] If the context times out while the pod is still in pending state, the error is printed on the debug 
- [x] Need to integrate the timeout with ```receptor , receptorctl , AWX and tower UI``` as well . The user will input a timeout value in UI and that will be propagated to  receptor

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -13,6 +13,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/ghjm/cmdline"
 	"github.com/google/shlex"
@@ -51,6 +52,7 @@ type kubeUnit struct {
 	config              *rest.Config
 	clientset           *kubernetes.Clientset
 	pod                 *corev1.Pod
+	podPendingTimeout   time.Duration
 }
 
 // kubeExtraData is the content of the ExtraData JSON field for a Kubernetes worker
@@ -106,6 +108,7 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
 	if err != nil {
 		return err
 	}
+
 	pod := &corev1.Pod{}
 	spec := &corev1.PodSpec{}
 	objectMeta := &metav1.ObjectMeta{}
@@ -203,7 +206,12 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
 			return kw.clientset.CoreV1().Pods(ked.KubeNamespace).Watch(kw.ctx, options)
 		},
 	}
-	ev, err := watch2.UntilWithSync(kw.ctx, lw, &corev1.Pod{}, nil, podRunningAndReady)
+
+	ctxPodReady := kw.ctx
+	if kw.podPendingTimeout != time.Duration(0) {
+		ctxPodReady, _ = context.WithTimeout(kw.ctx, kw.podPendingTimeout)
+	}
+	ev, err := watch2.UntilWithSync(ctxPodReady, lw, &corev1.Pod{}, nil, podRunningAndReady)
 	var ok bool
 	kw.pod, ok = ev.Object.(*corev1.Pod)
 	if !ok {
@@ -219,6 +227,12 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
 		}
 		return err
 	} else if err != nil {
+		kw.Cancel()
+		if len(kw.pod.Status.ContainerStatuses) == 1 {
+			if kw.pod.Status.ContainerStatuses[0].State.Waiting != nil {
+				return fmt.Errorf("%s, %s", err.Error(), kw.pod.Status.ContainerStatuses[0].State.Waiting.Reason)
+			}
+		}
 		return err
 	}
 	if ev == nil {
@@ -661,6 +675,7 @@ func (kw *kubeUnit) SetFromParams(params map[string]string) error {
 		}
 		return ssf
 	}
+
 	var err error
 	ked.KubePod, err = readFileToString(ked.KubePod)
 	if err != nil {
@@ -674,6 +689,7 @@ func (kw *kubeUnit) SetFromParams(params map[string]string) error {
 	userCommand := ""
 	userImage := ""
 	userPod := ""
+	podPendingTimeoutString := ""
 	values := []value{
 		{name: "kube_command", permission: kw.allowRuntimeCommand, setter: setString(&userCommand)},
 		{name: "kube_image", permission: kw.allowRuntimeCommand, setter: setString(&userImage)},
@@ -683,6 +699,7 @@ func (kw *kubeUnit) SetFromParams(params map[string]string) error {
 		{name: "secret_kube_pod", permission: kw.allowRuntimePod, setter: setString(&userPod)},
 		{name: "kube_verify_tls", permission: kw.allowRuntimeTLS, setter: setBool(&ked.KubeVerifyTLS)},
 		{name: "kube_tls_ca", permission: kw.allowRuntimeTLS, setter: setString(&ked.KubeTLSCAData)},
+		{name: "pod_pending_timeout", permission: kw.allowRuntimeParams, setter: setString(&podPendingTimeoutString)},
 	}
 	for i := range values {
 		v := values[i]
@@ -703,6 +720,16 @@ func (kw *kubeUnit) SetFromParams(params map[string]string) error {
 	if userPod != "" && (userParams != "" || userCommand != "" || userImage != "") {
 		return fmt.Errorf("params kube_command, kube_image, kube_params not compatible with secret_kube_pod")
 	}
+
+	if podPendingTimeoutString != "" {
+		podPendingTimeout, err := time.ParseDuration(podPendingTimeoutString)
+		if err != nil {
+			logger.Error("Failed to parse pod_pending_timeout -- valid examples include '1.5h', '30m', '30m10s'")
+			return err
+		}
+		kw.podPendingTimeout = podPendingTimeout
+	}
+
 	if userCommand != "" {
 		ked.Command = userCommand
 	}


pkg/workceptor/kubernetes.go
```diff
@@ -683,6 +701,7 @@ func (kw *kubeUnit) SetFromParams(params map[string]string) error {
 		{name: "secret_kube_pod", permission: kw.allowRuntimePod, setter: setString(&userPod)},
 		{name: "kube_verify_tls", permission: kw.allowRuntimeTLS, setter: setBool(&ked.KubeVerifyTLS)},
 		{name: "kube_tls_ca", permission: kw.allowRuntimeTLS, setter: setString(&ked.KubeTLSCAData)},
+		{name: "pod_pending_timeout", permission: kw.allowRuntimeAuth, setter: setString(&ked.PodPendingTimeout)},
```
fosterseth: should be permission allowruntimeparams

pkg/workceptor/kubernetes.go
```diff
@@ -203,7 +206,15 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
 			return kw.clientset.CoreV1().Pods(ked.KubeNamespace).Watch(kw.ctx, options)
 		},
 	}
-	ev, err := watch2.UntilWithSync(kw.ctx, lw, &corev1.Pod{}, nil, podRunningAndReady)
+
+	timeout, err := strconv.Atoi(ked.PodPendingTimeout)
```
fosterseth: We should be using time.ParseDuration instead https://golang.org/pkg/time/#ParseDuration

it allows user the specify units of time to, e.g. `1.5h`, `20m`, etc

pkg/workceptor/kubernetes.go
```diff
@@ -203,7 +206,15 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
...
+	ctxWithTimeout := kw.ctx
+	if err != nil {
+		logger.Debug("Error parsing timeout, using infinte timeout")
```
fosterseth: if there is an error parsing the timeout param we should error and return

pkg/workceptor/kubernetes.go
```diff
@@ -55,15 +56,16 @@ type kubeUnit struct {
...
+	Image             string
+	Command           string
+	Params            string
+	KubeNamespace     string
+	KubeVerifyTLS     bool
+	KubeTLSCAData     string
+	KubeConfig        string
+	KubePod           string
+	PodName           string
+	PodPendingTimeout string
```
fosterseth: I think this should be on the kubeUnit struct rather than kubeExtraData

pkg/workceptor/kubernetes.go
fosterseth: we should probably parse the duration earlier, like in SetFromParams -- the earlier in the code we can return an error the better. We can parse into a time.Duration object and put that on the kubeUnit struct directly, not the string representation.

pkg/workceptor/kubernetes.go
amolgautam25: I have made the changes. I dont think that this line is required anymore.

pkg/workceptor/kubernetes.go
amolgautam25: I have made the required changes.

pkg/workceptor/kubernetes.go
amolgautam25: I have made the required changes. The error gets returned early now.

pkg/workceptor/kubernetes.go
amolgautam25: I have made the required changes.

pkg/workceptor/kubernetes.go
```diff
@@ -703,6 +718,14 @@ func (kw *kubeUnit) SetFromParams(params map[string]string) error {
 	if userPod != "" && (userParams != "" || userCommand != "" || userImage != "") {
 		return fmt.Errorf("params kube_command, kube_image, kube_params not compatible with secret_kube_pod")
 	}
+
+	if pod_pending_timeout_string, ok := params["pod_pending_timeout"]; ok {
```
fosterseth: we still need to go through the setter above, like all the other params
```golang
pod_pending_timeout_string := ""
.
.
{name: "pod_pending_timeout", permission: kw.allowRuntimeParams, setter: setString(&pod_pending_timeout_string)},

```
then you can set `kw.podPendingTimeout` to `pod_pending_timeout_string`

pkg/workceptor/kubernetes.go
```diff
@@ -703,6 +718,14 @@ func (kw *kubeUnit) SetFromParams(params map[string]string) error {
...
+		pod_pending_timeout, err := time.ParseDuration(pod_pending_timeout_string)
+		if err != nil {
+			return fmt.Errorf("Caould Not Parse pod_pending_timeout")
```
fosterseth: let's put a logger.Error message, then return the err directly

```golang
logger.Error("Failed to parse pod_pending_timeout -- valid examples include '1.5h', '30m', '30m10s'")
return err
```

pkg/workceptor/kubernetes.go
```diff
@@ -51,6 +52,7 @@ type kubeUnit struct {
 	config              *rest.Config
 	clientset           *kubernetes.Clientset
 	pod                 *corev1.Pod
+	PodPendingTimeout   time.Duration
```
fosterseth: let's use camelCase here and elsewhere, so instead of pod_pending_timeout_string it should be podPendingTimeoutString

amolgautam25: I have made the required changes.

pkg/workceptor/kubernetes.go
amolgautam25: I have made the required changes.

pkg/workceptor/kubernetes.go
amolgautam25: I have made the required changes.

pkg/workceptor/kubernetes.go
```diff
@@ -203,7 +206,12 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
 			return kw.clientset.CoreV1().Pods(ked.KubeNamespace).Watch(kw.ctx, options)
 		},
 	}
-	ev, err := watch2.UntilWithSync(kw.ctx, lw, &corev1.Pod{}, nil, podRunningAndReady)
+
+	ctxWithTimeout := kw.ctx
+	if kw.podPendingTimeout != time.Duration(time.Duration.Seconds(0)) {
```
fosterseth: I think this can just be `!= time.Duration(0)`

pkg/workceptor/kubernetes.go
```diff
@@ -203,7 +206,12 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
 			return kw.clientset.CoreV1().Pods(ked.KubeNamespace).Watch(kw.ctx, options)
 		},
 	}
-	ev, err := watch2.UntilWithSync(kw.ctx, lw, &corev1.Pod{}, nil, podRunningAndReady)
+
+	ctxWithTimeout := kw.ctx
```
fosterseth: sometimes this context is not a timeout, so let's just call it something else
`ctxPodReady` instead of `ctxWithTimeout`

pkg/workceptor/kubernetes.go
amolgautam25: I have made the required changes.

pkg/workceptor/kubernetes.go
amolgautam25: I have made the required changes.

fosterseth (approved):

amolgautam25:
recheck

PR #319, opened 2021-05-12
Obtain status file lock when updating full status

shanemcd:
We are randomly seeing errors that look like this:

```
2021-05-03 07:59:09,461 ERROR    [a40f6c03679040dc990455ce2422c80c] awx.main.tasks job 111 (running) Exception occurred while running task
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/tasks.py", line 1427, in run
    res = receptor_job.run()
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/awx/main/tasks.py", line 2969, in run
    receptor_ctl.simple_command(f"work release {self.unit_id}")
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/receptorctl/socket_interface.py", line 72, in simple_command
    return self.read_and_parse_json()
  File "/var/lib/awx/venv/awx/lib64/python3.8/site-packages/receptorctl/socket_interface.py", line 51, in read_and_parse_json
    raise RuntimeError(text[7:])
RuntimeError: unlinkat /tmp/receptor/example.com/S4kPB890: directory not empty
```

### DIFF
--- a/pkg/workceptor/workunitbase.go
+++ b/pkg/workceptor/workunitbase.go
@@ -251,6 +251,8 @@ func (sfd *StatusFileData) UpdateFullStatus(filename string, statusFunc func(*St
 // UpdateFullStatus atomically updates the whole status record.  Changes should be made in the callback function.
 // Errors are logged rather than returned.
 func (bwu *BaseWorkUnit) UpdateFullStatus(statusFunc func(*StatusFileData)) {
+	bwu.statusLock.Lock()
+	defer bwu.statusLock.Unlock()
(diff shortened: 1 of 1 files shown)


fosterseth (approved):

ghjm (commented):
Before merging this I would want to go through the codebase for the dozen or so callers of this function, and make sure none of them touch the status lock inside their update function.

shanemcd:
I was wrong. This does not fix the issue.

shanemcd:
I think we actually want / need this now.

PR #320, opened 2021-05-13
Ensure that status lockfile is closed before trying to release work

shanemcd:
For context, see:

- https://github.com/ansible/receptor/pull/319
- https://github.com/ansible/awx/issues/9961

### DIFF
--- a/pkg/workceptor/workunitbase.go
+++ b/pkg/workceptor/workunitbase.go
@@ -114,6 +114,7 @@ func (bwu *BaseWorkUnit) StdoutFileName() string {
 func (sfd *StatusFileData) lockStatusFile(filename string) (*lockedfile.File, error) {
 	lockFileName := filename + ".lock"
 	lockFile, err := lockedfile.OpenFile(lockFileName, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0600)
+	sfd.LockFile = lockFile
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


chrismeyersfsu:
I can't put my finger on it. But this feels like the wrong way to solve this.

pkg/workceptor/workunitbase.go
```diff
@@ -366,8 +368,14 @@ func (bwu *BaseWorkUnit) UnredactedStatus() *StatusFileData {
 func (bwu *BaseWorkUnit) Release(force bool) error {
 	bwu.statusLock.Lock()
 	defer bwu.statusLock.Unlock()
+	if bwu.status.LockFile != nil {
+		// There seems to be a race condition with the `defer`s that
```
ghjm, line 372: If we're right here, having already passed the `!= nil` check, and some other goroutine or thread gains control and sets `bwu.status.LockFile` to nil, then the following call to `Close()` will panic because it is trying to call a function on a nil pointer.

fosterseth: eek good point. My first instinct is that we ought to make more use of the statusLock mutex. There are some methods that are reading and writing to status file without getting a lock. `monitorLocalStatus` calls Load, which grabs status file lock, without getting the mutex lock, for example.

ghjm: I used way too many mutex locks in Receptor.  A better and more Go-idiomatic way to do this would be to have a single goroutine that does all the status file changes, with channels for other processes to send updates or receive current data.

ghjm: I remember there was a lot of subtlety to this, that I spent a lot of time on it, and that the absence of mutex locks in the status file updates is not just an accident - I had _something_ in mind.  But looking at it now, it seems pretty reasonable that you should hold a read lock on `BaseWorkUnit.statusLock` while performing a file update, or a full lock if you're reading back from the file to memory (ie, changing the in-memory data).

shanemcd:
Replaced by https://github.com/ansible/receptor/pull/321

PR #321, opened 2021-05-15
Add missing locks

shanemcd:
For context, see:

- https://github.com/ansible/receptor/pull/319
- https://github.com/ansible/awx/issues/9961


pkg/workceptor/workunitbase.go
```diff
@@ -318,6 +320,8 @@ func (bwu *BaseWorkUnit) monitorLocalStatus() {
 		select {
 		case event := <-watcherEvents:
 			if event.Op&fsnotify.Write == fsnotify.Write {
+				bwu.statusLock.RLock()
```
ghjm: Seems like this should be a full lock, not just a read lock, since we're writing to the in-memory data structure.

shanemcd: Oh? I assumed we only wanted a full lock when writing to the file. I must have misunderstood how these were being used.

ghjm: The lock object protects the in-memory data structure.  When we're loading from the file, we're writing to the copy in memory.

PR #322, opened 2021-05-18
Surface logs from command-runner processes

shanemcd:


pkg/logger/logger.go
```diff
@@ -50,6 +51,18 @@ func GetLogLevel() int {
...
 
+// LogLevelToName takes an int and returns the corresponding log level name
+func LogLevelToName(logLevel int) (string, error) {
+	var err error
+	for k, v := range logLevelMap {
+		if v == logLevel {
+			return k, nil
+		}
+	}
+	err = fmt.Errorf("%s is not a valid log level", strconv.Itoa(logLevel))
```
fosterseth: I think just %d instead of %s should work, without needing to use `strconv.Itoa`

fosterseth:
maybe we should call LogLevelToName here to reduce the duplicated code

https://github.com/ansible/receptor/blob/568db1c683c9949239768388c289ef7ea2ff9a8d/pkg/logger/logger.go#L78

fosterseth (approved):

PR #323, opened 2021-05-18
No need for a separate signal handler goroutine

ghjm:
Signal handling can be integrated into the main select loop, simplifying the logic


fosterseth (approved):

shanemcd (approved):

shanemcd:
Ugh, I clicked a button in the UI and GitHub created an ugly merge commit.

PR #329, opened 2021-05-24
Respect the kube worker context in a couple places

shanemcd:


pkg/workceptor/kubernetes.go
```diff
@@ -381,6 +383,11 @@ func (kw *kubeUnit) runWorkUsingLogger() {
 	streamWait.Wait()
 	close(finishedChan)
 	if errStdin != nil || errStdout != nil {
+		select {
+		case <-kw.ctx.Done():
+			return
```
fosterseth: it might be essential that `runWorkUsingLogger` returns only with UpdateBasicStatus has been set to "WorkStateFailed" or "WorkStateSucceeded". It looks like a return on this line could occur when the job status still says "Running"

```golang
	// Launch runner process
	if kw.streamMethod == "tcp" {
		go kw.runWorkUsingTCP()
	} else {
		go kw.runWorkUsingLogger()
	}
	go kw.monitorLocalStatus()
```

that `monitorLocalStatus`, which runs at the same time as `runWorkUsingLogger`, only terminates when it sees that the job is complete (failed or succeeded). We might be leaking that.

I think we can just do an UpdateBasicStatus with Failed state right before returning here (unless the whole point of this select return block is to prevent a status file write...)

fosterseth (approved):

pkg/workceptor/kubernetes.go
shanemcd: I don't think we actually need this `select` now that we have a retry in the `Release` function. Removed.

PR #336, opened 2021-06-02
[WIP] issue 317

### LINKED ISSUE #317: work release <unitid> results in panic
### Steps to reproduce
* Start a receptor worker node `cmd/cmd --config local.yml`
* Submit work through the local node. Note, we are going to pick a node that doesn't exist just so the work hangs out. `RECEPTORCTL_SOCKET=/tmp/receptor.sock receptorctl work submit --node foobar --no-payload work` Note the `Unit ID:` returned.
* Shutdown the receptor node `Ctrl+C`
* Start the receptor node back up `cmd/cmd --config local.yml`
* Release the work previously started `receptorctl work release <Unit ID>`

### Expected
`Released: ('z6m3Tqtl',)`

### Actual
`<Unit ID>: ERROR: Expecting value: line 1 column 1 (char 0)`

Panic
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x13d666e]

goroutine 46 [running]:
github.com/project-receptor/receptor/pkg/utils.(*JobContext).Cancel(...)
        /home/meyers/ansible/receptor/pkg/utils/job_context.go:92
github.com/project-receptor/receptor/pkg/workceptor.(*remoteUnit).cancelOrRelease(0xc0004968f0, 0x410001, 0xc0002655c0, 0x1750c88)
        /home/meyers/ansible/receptor/pkg/workceptor/remote_work.go:554 +0x1ee
github.com/project-receptor/receptor/pkg/workceptor.(*remoteUnit).Release(0xc0004968f0, 0xc000495400, 0x8, 0x18954c0)
        /home/meyers/ansible/receptor/pkg/workceptor/remote_work.go:580 +0x39
github.com/project-receptor/receptor/pkg/workceptor.(*workceptorCommand).ControlFunc(0xc0001117a0, 0xc0002f9b00, 0x187df60, 0xc0004a48f0, 0x0, 0x0, 0x0)
        /home/meyers/ansible/receptor/pkg/workceptor/controlsvc.go:276 +0x41b
github.com/project-receptor/receptor/pkg/controlsvc.(*Server).RunControlSession(0xc000375b30, 0x188be80, 0xc00038a090)
        /home/meyers/ansible/receptor/pkg/controlsvc/controlsvc.go:210 +0xaa2
github.com/project-receptor/receptor/pkg/controlsvc.(*Server).RunControlSvc.func2.1(0x188be80, 0xc00038a090, 0xc00038c020, 0xc000375b30)
        /home/meyers/ansible/receptor/pkg/controlsvc/controlsvc.go:343 +0x66
created by github.com/project-receptor/receptor/pkg/controlsvc.(*Server).RunControlSvc.func2
        /home/meyers/ansible/receptor/pkg/controlsvc/controlsvc.go:320 +0x5e
```

```
# local.yml
---
- log-level: debug
- control-service:
    service: control
    filename: /tmp/receptor.sock
- local-only:
- work-command:
    worktype: worker
    command: ansible-runner
    params: worker
- node:
    id: foo
```

### Extra Info
Interesting that calling work release on a directory that does NOT exist in `/tmp/receptor/foo/` is handled OK. 

Related changes https://github.com/ansible/receptor/pull/235 These changes also result in receptor failing to resume work for nodes it never saw, when restarted.

amolgautam25:
Proper Handling of Nil context being passed around

### DIFF
--- a/pkg/workceptor/remote_work.go
+++ b/pkg/workceptor/remote_work.go
@@ -7,8 +7,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/project-receptor/receptor/pkg/logger"
-	"github.com/project-receptor/receptor/pkg/utils"
 	"io"
 	"net"
 	"os"
@@ -17,6 +15,9 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/project-receptor/receptor/pkg/logger"
+	"github.com/project-receptor/receptor/pkg/utils"
 )
 
 // remoteUnit implements the WorkUnit interface for the Receptor remote worker plugin
@@ -535,6 +536,7 @@ func (rw *remoteUnit) Restart() error {
 	if red.RemoteStarted {
 		return rw.startOrRestart(false)
 	}
+	rw.topJC = nil
 	return fmt.Errorf("remote work had not previously started")
 }
 
@@ -551,8 +553,10 @@ func (rw *remoteUnit) cancelOrRelease(release bool, force bool) error {
 	})
 	// if remote work has not started, don't attempt to connect to remote
 	if !remoteStarted {
-		rw.topJC.Cancel()
-		rw.topJC.Wait()
+		if rw.topJC != nil {
+			rw.topJC.Cancel()
+			rw.topJC.Wait()
+		}
 		if release {
 			return rw.BaseWorkUnit.Release(true)
 		}


pkg/controlsvc/controlsvc.go
```diff
@@ -115,6 +116,10 @@ func (s *Server) AddControlFunc(name string, cType ControlCommandType) error {
 func (s *Server) RunControlSession(conn net.Conn) {
 	logger.Info("Client connected to control service\n")
 	defer func() {
+		if conn == nil {
+			logger.Debug("nil context")
```
shanemcd: I think `conn` stands for connection. Maybe something like "Connection already closed, continuing."

amolgautam25: I have changed it , I think that the problem lies somewhere else and have made changes accordingly.

pkg/workceptor/remote_work.go
```diff
@@ -535,6 +536,7 @@ func (rw *remoteUnit) Restart() error {
 	if red.RemoteStarted {
 		return rw.startOrRestart(false)
 	}
+	rw.topJC = nil
```
fosterseth, line 539: `return rw.startOrRestart(false)` may also return an error, and we might need to set topJC to nil in that case too

pkg/controlsvc/controlsvc.go
ghjm: I think this is only failing because we have already panicked elsewhere.  Note that `conn` is a local variable which is never assigned in the body of this function, so it can't become `nil` as a result of local code, and it would fail on the first write if it was passed in as a nil.

ghjm:
See https://github.com/ansible/receptor/pull/337.  This fixes the panic that @chrismeyersfsu reported.  I'm not sure I understand what @fosterseth is saying about `r.startOrRestart`.

shanemcd:
Going to close this in favor of #337.

PR #337, opened 2021-06-02
Fix issue where releasing a never-started remote job caused a panic

ghjm:
See https://github.com/ansible/receptor/issues/317.  What's going on here is:

- Receptor is trying to "restart" a remote job, but has never actually started it.
- A `JobContext` has been created, but `NewJob` has never been called on it.
- Therefore, the fields of the `JobContext` - `ctx`, `cancel`, `wg` etc - are all nil.
- When the user releases the job, it attempts to cancel the context.
- This results in calling a nil function.

The fix is:

- The `JobContext.Cancel()` and `JobContext.Wait()` functions should just be no-ops if `NewJob` was never cancelled.

### DIFF
--- a/pkg/utils/job_context.go
+++ b/pkg/utils/job_context.go
@@ -63,8 +63,11 @@ func (mw *JobContext) WorkerDone() {
 }
 
 // Wait waits for the current job to complete, like sync.WaitGroup.Wait().
+// If no job has been started, always just returns.
 func (mw *JobContext) Wait() {
-	mw.wg.Wait()
+	if mw.wg != nil {
+		mw.wg.Wait()
+	}
 }
 
 // Done implements Context.Done()
@@ -87,9 +90,11 @@ func (mw *JobContext) Value(key interface{}) interface{} {
 	return mw.ctx.Value(key)
 }
 
-// Cancel cancels the JobContext's context
+// Cancel cancels the JobContext's context.  If no job has been started, this does nothing.
 func (mw *JobContext) Cancel() {
-	mw.cancel()
+	if mw.cancel != nil {
+		mw.cancel()
+	}
 }
 
 // Running returns true if a job is currently running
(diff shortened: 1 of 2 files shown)


chrismeyersfsu:
My thoughts on a proper, long-term fix.

## Receptor Remote Work Protocol

The diagrams do not capture the current state of the code in Receptor. They are to start a conversation that moves towards desired behavior of Receptor. Think idempotent.

### Happy Path

Receptor work protocol for work submitted to Bob destined for Alice.

uncached image

### Crash Scenarios

1. The first `work unit 123` never makes it to Alice. Upon restart, Bob will resubmit `work unit 123` request to Alice.

uncached image

2. Bob submits work that is already known to Alice. In this case, Alice ignores the work and replies to Bob informing him of this case. Bob then records this for efficiency sake. If Bob were to crash any time after `RemoteStarted 123`, he would know that work has already been submitted to Alice. Note that even if Bob lost his recording of `RemoteStarted 123`; it would be fine for him to re-submit `work unit 123`. The `RemoteStarted` is a nicety.

uncached image

Note that the above 2 crash scenarios are really the same scenario; a crash of Bob or Alice occuring before `RemoteStarted 123` will result in re-submitting `work unit 123` to Alice. This is idempotent behavior.

shanemcd (approved):

fosterseth (approved):
looks good, thanks!

ghjm:
@chrismeyersfsu I think the approach you're describing has merit.  The current remote work unit keeps track of two different unit IDs, one on the local node that represents the remote work supervisor process, and a different one on the remote node that represents the actual work being done.  This is prone to race conditions and edge cases.  I never did a full swim lane diagram or state map of the whole protocol, but intuitively it makes sense to me to have the remote supervisor send idempotent commands about the state of affairs it thinks ought to exist on the controlled node.

There are also, I think, interesting opportunities to make Receptor itself aware of work capacity and able to make some decisions about what node a unit of work is actually executed on.  I always had a vague idea that there should be group addresses, so you can submit a unit of work in a way that means "run this on the least busy of this set of nodes."

PR #342, opened 2021-06-25
Switch container & CI images from Fedora to CentOS 8

ghjm:
The Receptor container and CI images are still using Fedora 32, which is EOL.  This PR updates to Fedora 34.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,16 @@
 # Calculate version number
 # - If we are on an exact Git tag, then this is official and gets a -1 release
 # - If we are not, then this is unofficial and gets a -0.date.gitref release
-OFFICIAL_VERSION = $(shell if VER=`git describe --exact-match --tags 2>/dev/null`; then echo $$VER; else echo ""; fi)
-VERSION = $(shell cd receptorctl && python3 setup.py --version)
+OFFICIAL_VERSION := $(shell if VER=`git describe --exact-match --tags 2>/dev/null`; then echo $$VER; else echo ""; fi)
+VERSION := $(shell cd receptorctl && python3 setup.py --version)
 ifeq ($(OFFICIAL_VERSION),)
-RELEASE = 0.git$(shell date +'%Y%m%d').$(shell git rev-parse --short HEAD)
-OFFICIAL =
-APPVER = $(VERSION)-$(RELEASE)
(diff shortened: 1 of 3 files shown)


fosterseth (approved):

ghjm:
@fosterseth You know I don't have permissions to merge this myself, right?  😃

fosterseth:
we copy the receptor binary directly into our tools_awx container

https://github.com/ansible/awx/blob/bc8e19b51dbb9066318ae49faac3edef4a60ca71/tools/ansible/roles/dockerfile/templates/Dockerfile.j2#L169

if built on fedora 34, it doesn't run on our centos:8 image
`tools_awx_1 | awx-receptor stdout | receptor: /lib64/libc.so.6: version GLIBC_2.32' not found (required by receptor)`

maybe we can use centos:8? but I guess EOL for that is nearing too

ghjm:
Ah, interesting.  Can you check if building with the Makefile update I just pushed fixes this?

fosterseth:
this does work! but there were still talks about how centos:8 might make more sense for our case. Would changing to centos:8 as a base image work for your needs? @ghjm

ghjm:
I have no objections to CentOS 8.  There was some reason I used Fedora in the first place rather than CentOS, but I can't remember what it was, and I'm sure it's not anything I care about now.  Of course, CentOS 8 is EOL in six months.

ghjm:
I added the Makefile change from https://github.com/ansible/receptor/pull/346 since it seems this is likely to get merged first.

Makefile
```diff
@@ -49,7 +49,7 @@ else
 endif
 
 receptor: $(shell find pkg -type f -name '*.go') cmd/receptor.go
-	go build -ldflags "-X 'github.com/project-receptor/receptor/pkg/version.Version=$(APPVER)'" $(TAGPARAM) cmd/receptor.go
+	CGO_ENABLED=0 go build -ldflags "-X 'github.com/project-receptor/receptor/pkg/version.Version=$(APPVER)'" $(TAGPARAM) cmd/receptor.go
```
fosterseth, line 52: is statically linking necessary still? I'm guessing it doesn't hurt either

ghjm: I don't see a reason why we _want_ a dependency on a specific version of glibc...

Issue #343, opened 2021-06-28
work cancel <unit_id> doesn't cancel if running on another node, gives exit code of 0


AlanCoding:
I have a receptor cluster of 3 machines, call them A, B, and C

I have a job running on node A, and I know the unit id.

I shell into node B and run `receptorctl work cancel <unit_id>`. Then I get the standard out of:

```
Cancelled: ('wFvhMc80',)
```

it also gives an exit code of 0.

From reading this, I would expect that the unit id `wFvhMc80` has been canceled, but it has not been.

If, instead, I shell into node A and run the same command, it does cancel it.

fosterseth:
how was work submitted initially? from node A?

fosterseth:
in general, work units are not "seen" across the mesh. If you connect to node A and submit work on node A, then node B and C don't know about work unit `wFvhMc80`. Therefore, the job can only be cancelled or released from the node you started the work from.**

That said, `Cancelled: ('wFvhMc80',)` is not the response I would expect, investigating...

** The exception to this if you start remote work (while on node A, start work on node B). In that case, you could cancel from either place, but node A and node B would have 2 different work IDs.

AlanCoding:
> how was work submitted initially? from node A?

Yes, it was submitted on node A and it runs on node A. For now, that's how jobs work in this deployment type.

> That said, Cancelled: ('wFvhMc80',) is not the response I would expect, investigating...

Perhaps a more simple reproducer:

```
[awx@ip-10-0-13-122 ~]$ RECEPTORCTL_SOCKET=/var/run/receptor/receptor.sock /var/lib/awx/venv/awx/bin/receptorctl work cancel alan_made_this_up
Cancelled: ('alan_made_this_up',)
```

This was a non-obvious debugging process because (from the output) it was assumed that the job was canceled, and then later we hit the effects from the process still running.

> ** The exception to this if you start remote work (while on node A, start work on node B). In that case, you could cancel from either place, but node A and node B would have 2 different work IDs.

Good to know. When we get there, I want to have a test case for canceling on both node A and B on the AWX side.

PR #344, opened 2021-06-28
Fix receptorctl cancel command response

fosterseth:
related https://github.com/ansible/receptor/issues/343

The cancel (and release) command was ignoring the actual command response. Now it will return the exact response for each unit ID that is canceled. E.g. could be "unit not found" or "cancelled" or "cancel pending"

example output
```
Released:
(hyr4QvbG, unit not found)
(uzk2Z9k0, unit not found)
(wWE3ZWHS, released)
```


ghjm (approved):
Tested & works for me

shanemcd (approved):

PR #348, opened 2021-07-07
New tests environments

thenets:
Doesn't exist yet a simple way to generated artifacts in a controlled environment.

This PR is a possible approach to have everything being builded for the current source code state.

### Benefits:
- Easy way to test any individual commit
- Easy way to integrate CI process for PR
- Fixed version for main dependencies, like golang and python

### TODO:
- [x] Add build artifacts command
- [x] Allow change golang version
- [x] Allow change python version
- [x] Add command to run all tests on current source code
- [x] Create documentation
- [x] Create GitHub Action workflow to build artifacts 

### How to use:

- Build artifacts
  - Install podman
  - Go to `tests/`
  - Run `make artifacts`
  - The artifacts will be generated at `tests/artifacts`


tests/container/base/Dockerfile
```diff
@@ -0,0 +1,53 @@
+FROM fedora:34
```
shanemcd: We just switched away from using Fedora as the base image in a recent PR. It may be better to keep the tooling consistent, especially since Fedora moves so fast.

tests/container/test/run-ci.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+export PATH=/home/test-user/receptor:/home/test-user/go/bin:$PATH
```
shanemcd: Where is this script used?

tests/container/base/Dockerfile
thenets: Oh, cool. I'll switch to CentOS 8.

tests/container/test/run-ci.sh
thenets: My mistake. I copied from GitHub Action workflow tests. That will be used when I finish the task called "Add command to run all tests on current source code".

tests/container/Makefile
```diff
@@ -0,0 +1,28 @@
+IMAGE_TAG_BASE=ansible/receptor:base
+IMAGE_TAG_TEST=ansible/receptor:test
+
+image: image-base
+
+image-base:
+	podman build -t $(IMAGE_TAG_BASE) -f ./base/Dockerfile ../..
```
kdelee: s/Dockerfile/Containerfile/g


in the interest of being consistent with our other projects

tests/container/base/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+        git wget make iproute openssl findutils virtualenv \
+        # Python build requirements
+        gcc zlib-devel bzip2 bzip2-devel readline-devel sqlite \
+        sqlite-devel openssl-devel tk-devel libffi-devel xz-devel \
+    && dnf clean all
+
+# Install specific python version
+RUN set -x \
+    && cd /tmp \
+    && wget -q https://www.python.org/ftp/python/${PYTHON_VERSION}/Python-${PYTHON_VERSION}.tgz \
```
kdelee: as discussed, we can install the dnf module instead of building from source
https://www.server-world.info/en/note?os=CentOS_8&p=python&f=4

```
dnf module -y install python38
```
will have to tweak a few other things to use the right python binary

tests/container/base/Dockerfile
thenets: Done :)

tests/container/base/Dockerfile
thenets: Done :)

tests/container/test/run-ci.sh
thenets: Done :)

tests/container/Makefile
thenets: Done ^^

tests/environments/container/Containerfile
```diff
@@ -0,0 +1,51 @@
...
+    && dnf module install -y ${PYTHON_PKG_NAME} \
+    && dnf clean all \
+    && alternatives --set python /usr/bin/python${PYTHON_VERSION} \
+    # Install specific golang version
+    && cd /tmp/ \
+    && wget -q https://golang.org/dl/go${GOLANG_VERSION}.linux-amd64.tar.gz \
+    && rm -rf /usr/local/go \
+    && tar -C /usr/local -xzf go${GOLANG_VERSION}.linux-amd64.tar.gz \
+    && rm -f go${GOLANG_VERSION}.linux-amd64.tar.gz \
+    && echo 'export PATH=${PATH}:/usr/local/go/bin' >> /etc/bashrc
```
thenets: I'll need help for Go installation on CentOS 8.

If I run the Receptor CLI tests using the golang installation method above, then I got this error message:

```shell
+ go test -v ./functional/cli/cli_test.go
# runtime/cgo
cgo: exec /missing-cc: fork/exec /missing-cc: no such file or directory
FAIL    command-line-arguments [build failed]
FAIL
```

How to solve:

If I install the golang package using `yum`, everything works fine, but I can't specify a golang version this way.

```shell
yum install -y golang
```

I need help to find a way to make the first method works fine or find some way to install any (maybe not any) golang versions via `yum`.

thenets: Was decided to use the stable version from CentOS 8 repo.

thenets:
I thing the latest pipeline break for random reasons. Should be fixed and triggered again.

.github/workflows/build.yml
```diff
@@ -0,0 +1,65 @@
+---
+
+name: Build Artifact
+
+on: 
+  push:
+    branches:
+      - '*'
+    # tags:        
+    #   - v*
```
thenets: I think should be better restrict this to some branches. Maybe only `devel` for now, and maybe we can add more later.

.github/workflows/build.yml
```diff
@@ -0,0 +1,65 @@
...
+    #     mkdir dist
+    #     mv receptor ./dist/receptor-darwin-x86_64
+    #   if: matrix.os == 'macOS-latest'
+      
+    # - name: Build <windows>
+    #   run: |
+    #     go build -v .
+    #     mkdir dist
+    #     mv receptor.exe ./dist/receptor-windows-x86_64.exe
+    #   if: matrix.os == 'windows-latest'
```
thenets: I don't know if I should build artifacts for other OS too or not.

tests/receptor-tester.sh
```diff
@@ -0,0 +1,161 @@
+#!/bin/bash
```
shanemcd, line 1: I'm not a fan of the hand-rolled bash scripts. 

We should use a testing library, either Go or Python.

thenets: But it's just a way to invoke commands. This is not `pytest` wrap or something like this, it's just a way to run the existent tests inside containers. Should I use `Python` CLI to invoke commands instead of `bash`, already designed for CLI invoking?

The `README.md` that I created maybe can give you a better idea why I created this script/tool.

The reason I created this is to avoid `Makefile`. `Makefile` are cool for a lot of stuff, but it lacks documentation. At least with this `bash` script I'm able to invoke regular commands like `Makefile` (which also uses `sh` or `bash`) with a documentation embedded.

Go applications implies a dedicated dir, build processes, and dependencies installed. Python also requires those stuff and even an "requirements.txt" file if I need to use some dependency to create a CLI.

I can recreate this using Go or Python, but the consequence will be a lot of complexity been introduced. I also can convert this to regular Makefile commands, but it lacks context and documentation. Should I do this anyway?

.github/workflows/build.yml
thenets: CI removed. Maybe can be added in another PR.

.github/workflows/build.yml
thenets: CI removed. Maybe can be added in another PR.

shanemcd (approved):

PR #350, opened 2021-07-09
Live reload Backends

### LINKED ISSUE #354: Implement Live Reload feature in Receptor
Complexity estimate : Highly complex
Time Estimate: Pod dev work is done, just need other pods to review the code

Acceptance criteria:
Ability for a user to add or drop backend connection definitions to a running receptor server, and for the receptor server to respect those changes without disturbing ongoing receptor work, or without the need to restart the entire receptor process.

- [ ] https://github.com/ansible/receptor/pull/350

fosterseth:
Adds ability to add/remove backends from the config file without restarting Receptor

#### Context
We want to be able to add and remove nodes to the mesh without disturbing ongoing receptor functionality, e.g. sending payloads, or streaming playbook results

#### Details
Adds a "reload" service command, which does the following,
- Stop all backends
- Re-run `ParseAndRun()`, but with "Reload" as the input instead of "Init, Prepare, Run"
- `Reload()` is currently only defined for the 6 backend connections (peer and listeners for tcp, udp, websockets), so no other cfg objects will be re-ran on a reload. For now, `cfg.Reload()` simply calls `cfg.Prepare()`, and `cfg.Run()`

#### How to test this PR
save this as controlnode.yml
```
---
- node:
    id: controlnode
    
- log-level: Debug

- control-service:
    service: control
    filename: /tmp/receptor.sock

- tcp-peer:
    address: localhost:2222
```

save this as executionnode.yml
```
---
- node:
    id: executionnode

- log-level: Debug

- tcp-listener:
    port: 2223

- control-service:
    service: control

- work-command:
    workType: echosleep
    command: bash
    params: "-c \"while read -r line; do echo $line; sleep 5; done\""
```

1. `receptor -c controlnode.yml`
2. `receptor -c executionnode.yml`
3. You should see log statements about connections not working. This is because the ports don't match (2222 and 2223)
4. Keep nodes running, and modify controlnode.yml to make the port numbers match
5. `nc -U /tmp/receptor.sock` and type `reload`
6. You should now see the nodes successfully connect and send routing information to each other

More testing,
1. You can use receptorctl to submit work on the executionnode by typing `seq 100 | receptorctl --socket /tmp/receptor.sock work submit echosleep --node executionnode --payload - -f`
2. This will stream the results, a sequence of numbers.
3. While this is running, change the port numbers in the config to "break" the connection and issue reload command.
4. You'll see the work continues to stream, but no results will pop up.
5. Fix the config, and reload once more. The results should resume streaming once more.

### DIFF
--- a/pkg/backends/utils.go
+++ b/pkg/backends/utils.go
@@ -2,6 +2,7 @@ package backends
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"github.com/project-receptor/receptor/pkg/logger"
@@ -15,12 +16,16 @@ const (
 
 type dialerFunc func(chan struct{}) (netceptor.BackendSession, error)
 
-// dialerSession is a convenience function for backends that use dial/retry logic.
-func dialerSession(ctx context.Context, redial bool, redialDelay time.Duration,
+// dialerSession is a convenience function for backends that use dial/retry logic
+func dialerSession(ctx context.Context, wg *sync.WaitGroup, redial bool, redialDelay time.Duration,
 	df dialerFunc) (chan netceptor.BackendSession, error) {
 	sessChan := make(chan netceptor.BackendSession)
+	wg.Add(1)
 	go func() {
-		defer close(sessChan)
+		defer func() {
+			wg.Done()
+			close(sessChan)
+		}()
 		redialDelayInc := utils.NewIncrementalDuration(redialDelay, maxRedialDelay, 1.5)
 		for {
 			closeChan := make(chan struct{})
@@ -75,14 +80,17 @@ type (
 	listenerCancelFunc func()
 )
 
-// listenerSession is a convenience function for backends that use listen/accept logic.
-func listenerSession(ctx context.Context, lf listenFunc, af acceptFunc, lcf listenerCancelFunc) (chan netceptor.BackendSession, error) {
-	if err := lf(); err != nil {
+// listenerSession is a convenience function for backends that use listen/accept logic
+func listenerSession(ctx context.Context, wg *sync.WaitGroup, lf listenFunc, af acceptFunc, lcf listenerCancelFunc) (chan netceptor.BackendSession, error) {
+	err := lf()
+	if err != nil {
 		return nil, err
 	}
 	sessChan := make(chan netceptor.BackendSession)
+	wg.Add(1)
 	go func() {
 		defer func() {
+			wg.Done()
 			lcf()
 			close(sessChan)
 		}()
@@ -107,4 +115,16 @@ func listenerSession(ctx context.Context, lf listenFunc, af acceptFunc, lcf list
 	}()
 
 	return sessChan, nil
+}
+
+func runFuncs(f []func() error) error {
+	// convenience for running an slice of functions and returning any
+	// errors along the way
+	for _, toRun := range f {
+		err := toRun()
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
--- a/pkg/backends/utils.go
+++ b/pkg/backends/utils.go
@@ -16,7 +16,7 @@ const (
 
 type dialerFunc func(chan struct{}) (netceptor.BackendSession, error)
 
-// dialerSession is a convenience function for backends that use dial/retry logic
+// dialerSession is a convenience function for backends that use dial/retry logic.
 func dialerSession(ctx context.Context, wg *sync.WaitGroup, redial bool, redialDelay time.Duration,
 	df dialerFunc) (chan netceptor.BackendSession, error) {
 	sessChan := make(chan netceptor.BackendSession)
@@ -80,10 +80,9 @@ type (
 	listenerCancelFunc func()
 )
 
-// listenerSession is a convenience function for backends that use listen/accept logic
+// listenerSession is a convenience function for backends that use listen/accept logic.
 func listenerSession(ctx context.Context, wg *sync.WaitGroup, lf listenFunc, af acceptFunc, lcf listenerCancelFunc) (chan netceptor.BackendSession, error) {
-	err := lf()
-	if err != nil {
+	if err := lf(); err != nil {
 		return nil, err
 	}
 	sessChan := make(chan netceptor.BackendSession)
@@ -126,5 +125,6 @@ func runFuncs(f []func() error) error {
 			return err
 		}
 	}
+
 	return nil
 }
--- a/cmd/receptor.go
+++ b/cmd/receptor.go
@@ -69,8 +69,8 @@ func (cfg nodeCfg) Run() error {
 
 type nullBackendCfg struct{}
 
-// make the nullBackendCfg object be usable as a do-nothing Backend.
-func (cfg nullBackendCfg) Start(ctx context.Context) (chan netceptor.BackendSession, error) {
+// make the nullBackendCfg object be usable as a do-nothing Backend
+func (cfg nullBackendCfg) Start(ctx context.Context, wg *sync.WaitGroup) (chan netceptor.BackendSession, error) {
 	return make(chan netceptor.BackendSession), nil
 }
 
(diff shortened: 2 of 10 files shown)


fosterseth:
reloading a config and determining which backends are running is a little clunky.

Presumably there will be some external operation going on that knows precisely which nodes are being added and removed. So as an alternative, we *could* implement a way to pass in tcp-peer cfg, and a way to cancel a running backend as control service commands. That way, receptor doesn't need to infer from the config what has changed.

I guess it depends on how we intend to modify these receptor configs.

ghjm:
It seems pretty fundamentally impossible to associate a config entry in the old file with a config entry in the new file, after someone has edited it.  I see what you're doing with using the address as a key, but as you mention, this doesn't allow you to reconfigure backend settings.

Perhaps an alternative way of doing this would be to recreate all backends.  This could work like:
* Shut down all dialers
* Close all listeners and exit all accept loops, but allow already-established sessions to continue
* Start all the backends in the config file, without making any attempt to correlate them with the old backends

This means that when you do a reload on the dialer side, its connection will be broken and re-established.  (Perhaps it should be an error to do a reload if there are active dialers with redial:false.)  There may also be some fancy footwork needed to make sure ports are available for the new listeners to bind to.

Another alternative is to just do a restart of the Receptor process.  Workceptor is robust about retrying, resuming streams, etc.  The main problem is breaking the local Unix socket connection.  You would need a handler that retransmits work that was being sent, and re-establishes streaming from results that were being received.  This is certainly effort, but perhaps less effort than trying to implement kill -HUP functionality in Receptor.

Also, cluster architecture can help a lot with this.  I noticed that in the latest AWX builds there seem to be short-lived receptor instances inside execution environment pods.  I think it would be better to have a smaller number of long-lived receptor instances, which orchestrate EE pods from outside.  If there's a desire to have an agent running inside a short-lived EE pod, then netceptor, as it currently exists, might not be the right protocol.

shanemcd:
Hi Graham! Thanks for taking a look at this. Just wanted to clarify one thing for now:

> I noticed that in the latest AWX builds there seem to be short-lived receptor instances inside execution environment pods.

This is incorrect. There is 1 long-lived receptor process in each control plane pod. I am curious what you saw / what the confusion was.

ghjm:
For the control plane jobs there's a long-lived Receptor process in the awx-ee container, but the execution plane jobs in a container group seem to launch a separate instance of Receptor that only runs as long as the job does.  Unless I misunderstood something.

fosterseth:
another alternative, use a sidecar receptor process for handling local domain sockets to the main receptor process

e.g. foo.yml, foosidecar.yml

If we need to add a backend to foo, we modify foo.yml, then restart the entire foo process. Meanwhile, foosidecar keeps running. AWX only interacts with foosidecar, so domain sockets are never broken.

Pros:

This leverages all the great reconnect functionality that is already in receptor. We probably don't even need to modify receptor at all to make this work.

Cons:

Every node would need 2 running receptor processes, making things even harder to debug, etc. Probably other cons I can't immediately think of...

fosterseth:
some mores cons to running a sidecar receptor for handling local traffic

- stdin and stdout will be duplicated on disk, one for the sidecar and one for the main receptor process. Passing and storing this data twice could have performance ramifications
- these 2 receptor processes will need to be connected over a valid backend connection, which means at least one of these would need a tcp-listener, which might not be desirable

ghjm:
One point I would make here is that the reload problem is entirely an issue of the Receptor CLI and YAML config files.  The underlying library has no problem at all with adding and removing back-ends at runtime.  So I don't see anything stopping you from adding control service commands to start and stop backends at runtime, that aren't part of the YAML at all.

fosterseth:
Yes, and if we go this route, we might have to concede having the config file represent the full state of the receptor node, unless we take extra care to also add and remove backends to the config when issuing said commands to receptor.

if a node crashes, on boot up we'd need a way to re-add or re-remove the backends we did during runtime. This _could_ just be an external tool to receptor that maintains mesh topology and adds/removes backends as needed.

ghjm:
Or you could go the other way, and not have _any_ backends defined in the YAML config, and have the Tower dispatcher (or whatever other process) set up backends as needed.  The basic question is what the source of truth should be about the configuration of the mesh.

pkg/backends/utils.go
```diff
@@ -76,8 +76,10 @@ func listenerSession(ctx context.Context, lf listenFunc, af acceptFunc, lcf list
 		return nil, err
 	}
 	sessChan := make(chan netceptor.BackendSession)
+	netceptor.MainInstance.BackendAdd()
```
fosterseth: these calls to MainInstance are problematic in the case where netceptor is imported. I will look into this

fosterseth:
one change is that backendWaitGroup has been extended to include the go-routines associated with the Start() methods (i.e. the dialerSession and listenerSession go routines). Therefore, `netceptor.BackendWait()` waits for runProtocol goroutines to exit, as it did before this PR, and for all dialers, listeners to exit.

fosterseth:
@ghjm I've made some recent changes. I'm interested in your thoughts on this. Basically, the "reload" service command will shutdown all goroutines associated with all the backends (runprotocol, and session dialers, listeners), wait for the backendWaitGroup to hit 0, then reload the backends from the config file.

Will adding the session dialers and listeners to the backendwaitgroup be problematic for you? If so, I think I can create and track those goroutines in a separate wait group.

ghjm:
It makes sense to me.  Of course as you know, this means that calling Reload will break all existing connections, so after a Reload it will take several seconds for routing to be re-established.  So hopefully we're very careful with only calling Reload when we know something has actually changed.

Also, I'm not super keen on passing the whole netceptor object into the backend functions.  This is the kind of thing that leads to dependency loops.  If the only reason we're passing this in is so that a backend can call `BackendWaitGroupAdd` and `BackendDone`, then maybe it would be worth creating an interface that has those two functions?  Or just passing the waitgroup object itself?

I'm also wondering how external backends work with all this.  If an external service is running, does its `runProtocol` loop get shut down as part of Reload?  Are we just assuming it will somehow eventually restore itself?  Or does Reload just not really support external services?

shanemcd (approved):
I pulled this down and kicked the tires a bit. Worked as advertised. Nice work!

yagomarques:
@fosterseth following the scenario that you share like an example, check this situation:

**Scenario: Run Reload with invalid yml**

1. a receptor mesh with node "A" with the Control process running on it.
the `control.yml` file below
```
---
- node:
    id: controlnode
    
- log-level: Debug

- control-service:
    service: control
    filename: /tmp/receptor.sock

- tcp-peer:
    address: fedora-localhost:2223
```

2. run the work type: 
`seq 100 | receptorctl --socket /tmp/receptor.sock work submit echosleep --node executionnode --payload - -f`

3. When we edit the control config file adding an invalid item like:
```
- control-service:test
    service: control
    filename: /tmp/receptor.sock
```
4. And we run the Reload command
5. Then on the logs we should see that the Reload command failing:
`Error: error loading config file: yaml: line 8: mapping values are not allowed in this context`
6. And the receptor should be working with the old configuration

But right now the step 6 is not happening. After the step 5 the control_node stop to work.
```
INFO2021/07/30 12:56:06 Client connected to control service
DEBUG2021/07/30 12:56:06 Reloading
DEBUG2021/07/30 12:56:06 Canceling backends
INFO2021/07/30 12:56:06 Control service closed
INFO2021/07/30 12:56:06 Client disconnected from control service
DEBUG2021/07/30 12:56:07 Read error reading from executionnode: INTERNAL_ERROR: no connection to next hop
WARNING2021/07/30 12:56:07 Connection to executionnode failed with error: no connection to next hop
WARNING2021/07/30 12:56:08 Connection to executionnode failed with error: no connection to next hop
WARNING2021/07/30 12:56:10 Connection to executionnode failed with error: no connection to next hop
WARNING2021/07/30 12:56:12 Error copying to stdout file /tmp/receptor/controlnode/28lx3NNW/stdout: INTERNAL_ERROR: no connection to next hop
WARNING2021/07/30 12:56:12 Connection to executionnode failed with error: no connection to next hop
WARNING2021/07/30 12:56:15 Connection to executionnode failed with error: no connection to next hop
WARNING2021/07/30 12:56:20 Connection to executionnode failed with error: no connection to next hop
```

fosterseth:
@yagomarques thanks for bringing that up. I updated the PR to include error handling around that situation. If you give a bad-syntax yaml file, the reload will return an error before canceling any backends.

yagomarques:
hello @fosterseth, thanks for the update.

I retested, main scenarios are working. From the QE side we did a sanity check test, we could merge. In the future we gonna test more scenarios but in another layer(tower integration).

shanemcd (approved):

PR #352, opened 2021-07-13
Fixed receptorctl dictionary to String parsing

### LINKED ISSUE #345: Unsupported format string in receptorctl status
### Steps to reproduce : 

create a receptor node with following config  ```foo.yml```: 
```
---
- node:
    id: foo
    allowedpeers: bar, baz
- log-level: Debug
- trace
- control-service:
    service: control
    filename: /tmp/foo.sock
- tcp-listener:
    port: 2222
    cost: 1.0
    nodecost:
        bar: 1.5
        fish: 2.0
- command-service:
    service: echo
    command: "ls"
```

Then on ```receptorctl``` we can see this : 
```
(receptorctl-test-venv) [fedora@fedora-vm receptor]$ receptorctl --socket /tmp/foo.sock status
Node ID: foo
Version: 0.9.4.dev66-0.git20210629.fac08b5
System CPU Count: 6
System Memory MiB: 15971

Node         Service   Type       Last Seen           Tags            Work Types
Error: unsupported format string passed to dict.__format__

### PR DESCRIPTION
**Pull request by amolgautam25**
The cause of the issue was this line : 
https://github.com/ansible/receptor/blob/devel/receptorctl/receptorctl/cli.py#L117

```ad['Tags']``` was a dictionary in some cases , so i have explicitly changed it to a string. 

Looking for better way to do this.

### DIFF
--- a/receptorctl/receptorctl/cli.py
+++ b/receptorctl/receptorctl/cli.py
@@ -114,7 +114,7 @@ def status(ctx):
             elif ad['ConnType'] == 2:
                 conn_type = 'StreamTLS'
             print(
-                f"{ad['NodeID']:<{longest_node}} {ad['Service']:<9} {conn_type:<10} {time:%Y-%m-%d %H:%M:%S} {(ad['Tags'] or '-'):<16}",
+                f"{ad['NodeID']:<{longest_node}} {ad['Service']:<9} {conn_type:<10} {time:%Y-%m-%d %H:%M:%S} {'-' if (ad['Tags'] is None) else str(ad['Tags']):<16}",
                 end=""
             )
             commands = ad['WorkCommands']
--- a/receptorctl/receptorctl/cli.py
+++ b/receptorctl/receptorctl/cli.py
@@ -104,7 +104,7 @@ def status(ctx):
     ads = status.pop('Advertisements', None)
     if ads:
         print()
-        print(f"{'Node':<{longest_node}} Service   Type       Last Seen           Tags            Work Types")
+        print(f"{'Node':<{longest_node}} Service   Type       Last Seen           Tags                            Work Types")
         for ad in ads:
             time = dateutil.parser.parse(ad['Time'])
             if ad['ConnType'] == 0:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by fosterseth — APPROVED** [review]

File: receptorctl/receptorctl/cli.py [hunk 1]
```diff
@@ -114,7 +114,7 @@ def status(ctx):
             elif ad['ConnType'] == 2:
                 conn_type = 'StreamTLS'
             print(
-                f"{ad['NodeID']:<{longest_node}} {ad['Service']:<9} {conn_type:<10} {time:%Y-%m-%d %H:%M:%S} {(ad['Tags'] or '-'):<16}",
+                f"{ad['NodeID']:<{longest_node}} {ad['Service']:<9} {conn_type:<10} {time:%Y-%m-%d %H:%M:%S} {(str(ad['Tags'])  or '-'):<16}",
```
fosterseth: if ad['Tags'] is None, then str(None) is 'None'. But if ad['Tags'] is None, it should be '-'

amolgautam25: I have made some changes.

receptorctl/receptorctl/cli.py
```diff
@@ -114,7 +114,7 @@ def status(ctx):
             elif ad['ConnType'] == 2:
                 conn_type = 'StreamTLS'
             print(
-                f"{ad['NodeID']:<{longest_node}} {ad['Service']:<9} {conn_type:<10} {time:%Y-%m-%d %H:%M:%S} {(ad['Tags'] or '-'):<16}",
+                f"{ad['NodeID']:<{longest_node}} {ad['Service']:<9} {conn_type:<10} {time:%Y-%m-%d %H:%M:%S} {'-' if (ad['Tags'] is None) else str(ad['Tags']):<16}",
```
fosterseth: does this if conditional honor the :<16 at the end? 

best way to test is to put some more characters after like `{'-' if (ad['Tags'] is None) else str(ad['Tags']):<16} here is more"` and make sure you still get 16 spaces after Tags entry

PR #353, opened 2021-07-13
Fix add work types to service ads

fosterseth:
Somewhere along the way this broke, and we are no longer getting work command names in our `status` command output

- Only append work command names to a "control service". Before it was appending these names to every type of service, which doesn't make sense for things like udp-client, etc.

- control-service by default is called "control", but this isn't required. Therefore, we can use `.Tags` to keep track of which services are control services. Side note, there is probably a case to enforce control-service to be called "control". Remote work submission assumes target node has a "control" service running
https://github.com/fosterseth/sockceptor/blob/e0315e92a2ee99641f633d8a82a881d358d22836/pkg/workceptor/remote_work.go#L54

- the cli.py now prints work commands separately from the service information. Both the Tags and Work Types column can be arbitrarily long, so hard to keep the table columns fixed width. 

e.g. 

```
Node ID: foo
Version: 0.9.8.dev46-0.git20210713.e0315e9
System CPU Count: 8
System Memory MiB: 15876

Connection   Cost
bar          1

bar          {'foo': 1}
foo          {'bar': 1}

Route        Via
bar          bar

Node         Service   Type       Last Seen             Tags
foo          control2  Stream     2021-07-13 14:18:46   {'type': 'Control Service'}
foo          control3  Stream     2021-07-13 14:18:46   {'type': 'Control Service'}
bar          control   Stream     2021-07-13 14:18:38   {'type': 'Control Service'}

Node         Work Types
foo          sleepcat, 100cat, testparam, idk, echosleepshort, kubeit, kubeit2, kubernetes-runtime-auth
bar          sleepcat, 100cat, echosleepshort, idk
```

### DIFF
--- a/pkg/controlsvc/controlsvc.go
+++ b/pkg/controlsvc/controlsvc.go
@@ -290,7 +290,9 @@ func (s *Server) RunControlSvc(ctx context.Context, service string, tlscfg *tls.
 	}
 	var li *netceptor.Listener
 	if service != "" {
-		li, err = s.nc.ListenAndAdvertise(service, tlscfg, nil)
+		li, err = s.nc.ListenAndAdvertise(service, tlscfg, map[string]string{
+			"type": "Control Service",
+		})
 		if err != nil {
 			return fmt.Errorf("error opening Unix socket: %s", err)
 		}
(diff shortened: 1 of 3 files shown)


shanemcd:
Hmm... there seems to be a conflict with devel. You sure you were using up-to-date code?

shanemcd (approved):

PR #366, opened 2021-07-29
Enable Linters GigaPR

eqrx:
Extending #365 I wanted to enable golangci-lint linters one by one and see how far I can get without doing major code changes... Here we are, please forgive me for the gigantic PR. :D
I tried to group the changes into commits as much as possible. I also can split the PR into smaller ones if requested, but it feels to me like the change set belongs together.

Please do not hesitate to come back with change request to me, while I find that some fixes should be really applied, some are more debatable!


eqrx:
Splitting this PR into smaller ones

PR #381, opened 2021-08-16
Move version command to internal package

eqrx:
Keeping it in pkg could lead to confusion since it only gets
populated when the receptor binary is build with the Makefile,
not when importing it.


shanemcd:
@eqrx Can you rebase this plz?

eqrx:
@shanemcd done :)

shanemcd (approved):

PR #384, opened 2021-08-18
On reload, detect if modified config is reloadable

fosterseth:
If a user changes the configuration file that is not supported by a reload, then proper errors are returned.

reloadable config items include

```
tcp-peer
tcp-listener
ws-peer
ws-listener
udp-peer
udp-listener
local-only
```

Error message if a non-reloadable config item was added or modified,

```
$ receptorctl reload
Error: a non-reloabable config action 'work-command' was modified or added. Must restart receptor for these changes to take effect
```
Error message if a non-reloadable config item was removed,

```
$ receptorctl reload
Error: a non-reloadable config action 'work-command' was removed. Must restart receptor for changes to take effect
```

receptorctl will exit with a meaningful error code, can be view with `echo $?` 

exit code 0 -- receptor reload was successful
exit code 3 -- configuration file was fine, but cannot reload due to the non-reloadable changes mentioned above
exit code 4 -- new configuration file we are trying to reload is broken, even a restart is probably not going to work. Installer role should fail

### DIFF
--- a/receptorctl/receptorctl/cli.py
+++ b/receptorctl/receptorctl/cli.py
@@ -171,7 +171,12 @@ def reload(ctx):
         print(f"Reload successful")
     else:
         print(f"Error: {results['Error']}")
-
+        if "ERRORCODE 3" in results['Error']:
+            sys.exit(3)
+        elif "ERRORCODE 4" in results['Error']:
+            sys.exit(4)
+        else:
+            sys.exit(4)
 
 @cli.command(help="Do a traceroute to a Receptor node.")
 @click.pass_context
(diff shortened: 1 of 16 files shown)


receptorctl/receptorctl/cli.py
```diff
@@ -169,9 +169,15 @@ def reload(ctx):
     results = rc.simple_command(f"reload")
     if "Success" in results and results["Success"]:
         print(f"Reload successful")
+        sys.exit(0)
```
jbradberry: technically this isn't needed, running off of the end of the program will automatically give it a return code of 0.

fosterseth: good point, fixed

jbradberry (approved):

shanemcd (approved):

PR #387, opened 2021-08-19
Firewall Rules

shanemcd:
This PR adds the following security features to Receptor:

### Firewall rules

The firewall feature for adds a new method `netceptor.AddFirewallRules()`, which attaches a list of rules to this Netceptor instance.  Once attached, the rules are checked for every `MsgTypeData` packet, including both local and to-be-forwarded packets.  Firewall rules are processed just before dispatching the packet to a local service or router.

A firewall rule is simply a Go function that takes a `*MessageData` and returns a decision, which can be `FirewallResultAccept` (allow the packet), `FirewallResultReject` (deny the packet with an unreachable message), or `FirewallResultDrop` (deny the packet silently).

This PR changes the visibility of `MessageData` from private to public, so that it can be passed to firewall functions in other packages.

### Firewall CLI language & parser

A new option `firewallrules` is added to the `node` CLI struct.  This takes a list of hashes/dictionaries/objects, each describing a firewall rule to be created.  Rules can match simple strings or regular expressions.  Example rules

```yaml
# Accepts everything

---
node:
  firewallrules:
    - action: "accept"
```

```yaml
# Drops traffic from `foo` to `bar`'s control service

---
node:
  firewallrules:
    - action: "drop"
      fromnode: "foo"
      tonode: "bar"
      toservice: "control"
```

```yaml
# Rejects traffic originating from nodes like abcb, adfb, etc

---
node:
  firewallrules:
    - action: "reject"
      fromnode: "/a.*b/"
```

```yaml
# Rejects traffic destined for nodes like abcb, AdfB, etc

---
node:
  firewallrules:
    - action: "reject"
      tonode: "/(?i)a.*b/"
```

### Per-connection AllowedPeers

The existing AllowedPeers command applies to the whole Receptor node, with no regard to which connection a peer is connecting through.  In circumstances where control is desired over the node IDs allowed to connect, it is likely that the user will also want to control which connection they can connect through.

To accomplish this, the `AllowedPeers` parameter is removed from `netceptor.New()` and moved to `AddBackend()`.  **This is a breaking change** because we can't change the signature of these functions without breaking callers.  The `AddBackend()` function is changed to use the modifier pattern, which provides the ability to non-breakingly add more options in future.

#### TODOs

---

- [x] Add documentation

### DIFF
--- a/pkg/netceptor/firewall_rules.go
+++ b/pkg/netceptor/firewall_rules.go
@@ -2,10 +2,11 @@ package netceptor
 
 import (
 	"fmt"
-	"github.com/alecthomas/participle/v2"
-	"github.com/alecthomas/participle/v2/lexer/stateful"
 	"regexp"
 	"strings"
+
+	"github.com/alecthomas/participle/v2"
+	"github.com/alecthomas/participle/v2/lexer/stateful"
 )
 
 /*
@@ -39,7 +40,7 @@ TONODE = /(?i)a.*b/ : reject
 
 */
 
-// ParseFirewallRule takes a single string describing a firewall rule, and returns a FirewallRule function
+// ParseFirewallRule takes a single string describing a firewall rule, and returns a FirewallRule function.
 func ParseFirewallRule(rule string) (FirewallRule, error) {
 	parsedRule := &ruleString{}
 	err := ruleParser.ParseString("rule", rule, parsedRule)
@@ -48,19 +49,20 @@ func ParseFirewallRule(rule string) (FirewallRule, error) {
 	}
 	comps := make([]compareFunc, 0)
 	for _, pr := range parsedRule.RuleSpec.Rules {
-		if pr.Value.Value != "" {
+		switch rv := pr.Value; {
+		case rv.Value != "":
 			comp, err := stringCompare(pr.Field, pr.Value.Value)
 			if err != nil {
 				return nil, err
 			}
 			comps = append(comps, comp)
-		} else if pr.Value.Regex != "" {
+		case rv.Regex != "":
 			comp, err := regexCompare(pr.Field, pr.Value.Regex)
 			if err != nil {
 				return nil, err
 			}
 			comps = append(comps, comp)
-		} else {
+		default:
 			return nil, fmt.Errorf("no value or regex provided for field %s", pr.Field)
 		}
 	}
@@ -68,10 +70,11 @@ func ParseFirewallRule(rule string) (FirewallRule, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	return fwr, nil
 }
 
-// ParseFirewallRules takes a slice of string describing firewall rules, and returns a slice of FirewallRule functions
+// ParseFirewallRules takes a slice of string describing firewall rules, and returns a slice of FirewallRule functions.
 func ParseFirewallRules(rules []string) ([]FirewallRule, error) {
 	results := make([]FirewallRule, 0)
 	for i, rule := range rules {
@@ -81,6 +84,7 @@ func ParseFirewallRules(rules []string) ([]FirewallRule, error) {
 		}
 		results = append(results, result)
 	}
+
 	return results, nil
 }
 
@@ -103,6 +107,7 @@ func firewallRule(comparers []compareFunc, action string) (FirewallRule, error) 
 			return result
 		}, nil
 	}
+
 	return func(md *MessageData) FirewallResult {
 		matched := true
 		for _, comp := range comparers {
@@ -111,6 +116,7 @@ func firewallRule(comparers []compareFunc, action string) (FirewallRule, error) 
 		if matched {
 			return result
 		}
+
 		return FirewallResultContinue
 	}, nil
 }
@@ -134,6 +140,7 @@ func stringCompare(field string, value string) (compareFunc, error) {
 			return md.ToService == value
 		}, nil
 	}
+
 	return nil, fmt.Errorf("unknown field: %s", field)
 }
 
@@ -164,6 +171,7 @@ func regexCompare(field string, value string) (compareFunc, error) {
 			return re.MatchString(md.ToService)
 		}, nil
 	}
+
 	return nil, fmt.Errorf("unknown field: %s", field)
 }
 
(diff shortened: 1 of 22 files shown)


ghjm:
I'd recommend dropping Participle and writing a quick all-in-one state machine parser for the firewall rules language.  The problem is that the rules aren't a context-free grammar due to the embedded regexps, so a separate lexer is always going to be hacky at best.  Also, Participle is just something I used to prove the concept - it's a far bigger dependency than is actually needed here.

ghjm:
Also, you should copy and paste my carefully written description over to the new PR.  😁

shanemcd:
> I'd recommend dropping Participle and writing a quick all-in-one state machine parser for the firewall rules language.

@ghjm you trying to flex on me? 😅  As much as I wish I had the intelligence required to do this... you know I don't.

I can however think of a different way of achieving the same result that _might_ end up being faster than using a parser:

```yaml
---
node: 
  id: foo
  firewallrules:
    - action: drop
      fromnode: "foo"
      tonode: "/control-*/"
```

Other than stylistic preference, do you see any problem with this?

ghjm:
> Other than stylistic preference, do you see any problem with this?

No, I guess that will work, I just find it inelegant.

What I'm talking about is just reading through the string character by character, keeping track of what you're expecting next (the state).  It shouldn't be hard.  But if you prefer the YAML syntax, then that's your choice.

shanemcd:
@ghjm I think this PR is going to merge soon. Giving you a heads up in case you wanted a chance to review and provide any more feedback. I've updated the PR description to reflect the current state.

pkg/netceptor/firewall_rules.go
```diff
@@ -2,83 +2,117 @@ package netceptor
...
-
-	"github.com/alecthomas/participle/v2"
-	"github.com/alecthomas/participle/v2/lexer/stateful"
 )
 
 /*
 
-This file provides a parser for firewall rules provided by the CLI.  It takes strings
-and returns rule functions suitable for passing to AddFirewallRules().
+TODO: Rewrite docstring for firewall rules
```
shanemcd: Note to self: Don't forget to do this before merge.

ghjm:
In `receptor --help`, this is pretty ugly:

```
   --node: Node configuration of this instance (required)
      id=<string>: Node ID. Defaults to local hostname.
      datadir=<string>: Directory in which to store node data
      firewallrules=<JSON list of JSON dict of JSON data to JSON data>: Firewall Rules (see documentation for syntax)
```

But apparently y'all are dropping my cmdline library?  So maybe it doesn't matter?

Do we care about setting firewall rules from the command line, or is the assumption that anyone with a complicated enough configuration to need firewall rules will always be using a YAML config file?

shanemcd:
```
<JSON list of JSON dict of JSON data to JSON data>
```

LOL, I hadn't seen that yet.

> But apparently y'all are dropping my cmdline library?

Yes, I think we are heading in that direction, for all of the reasons described in the description of https://github.com/ansible/receptor/pull/385.

> Do we care about setting firewall rules from the command line, or is the assumption that anyone with a complicated enough configuration to need firewall rules will always be using a YAML config file?

I think we are currently leaning towards forcing the usage of a config file for everything. The new cobra/viper code currently doesn't allow for passing everything on the CLI in the way your cmdline package does, and it I am not sure we're going to spend the time reimplementing that functionality. Reason being that any amount of configuration quickly gets unwieldy when trying to represent as command line args. Maybe we could add that back later if enough people ask for it.

shanemcd:
Need to investigate https://github.com/ansible/receptor/issues/393#issuecomment-905884043

shanemcd:
I applied the patch from @ghjm. @yagomarques Can you give this another spin?

yagomarques:
@shanemcd i execute the same steps again, the error didn't occurs.

execution_node
```
DEBUG 2021/08/26 17:23:19 Sending routing update bfG1JwZZ. Connections: controlnode(1.00)
DEBUG 2021/08/26 17:23:19 Received routing update gu8Pz2Y6 from controlnode via controlnode
DEBUG 2021/08/26 17:23:22 Received service advertisement &{0xc000708b40 false}
DEBUG 2021/08/26 17:23:28 Sending service advertisement: &{executionnode control 2021-08-26 17:23:28.132952139 -0300 -03 m=+545.215051057 1 map[type:Control Service] [echosleep]}
DEBUG 2021/08/26 17:23:29 Sending routing update w0YAWqis. Connections: controlnode(1.00)
DEBUG 2021/08/26 17:23:29 Received routing update as20Fivs from controlnode via controlnode
DEBUG 2021/08/26 17:23:39 Sending routing update gTZ9BOtG. Connections: controlnode(1.00)

```

controlnode
```
DEBUG 2021/08/26 17:23:19 Received routing update bfG1JwZZ from executionnode via executionnode
DEBUG 2021/08/26 17:23:22 Sending service advertisement: &{controlnode control 2021-08-26 17:23:22.045094449 -0300 -03 m=+545.110177267 1 map[type:Control Service] []}
WARNING 2021/08/26 17:23:25 Connection to executionnode failed with error: NO_ERROR: Handshake did not complete in time
DEBUG 2021/08/26 17:23:28 Received service advertisement &{0xc0001d94a0 false}
DEBUG 2021/08/26 17:23:29 Sending routing update as20Fivs. Connections: executionnode(1.00)
DEBUG 2021/08/26 17:23:29 Received routing update w0YAWqis from executionnode via executionnode
DEBUG 2021/08/26 17:23:39 Sending routing update B2rOIJdm. Connections: executionnode(1.00)

```
Just an insight point, would be good to have a log related to this feature at least inside the executionnode?

For this scenario, something like:
`work rejected by firewall` or something like that

ghjm:
> Just an insight point, would be good to have a log related to this feature at least inside the executionnode?

Sure, y'all can add a log entry if you want.  It would probably be a good idea to be able to turn on logging of all firewall drops/rejections.  Or just log then by default at debug severity.

I don't know that it's particularly worth logging when you don't send an unreachable message from another unreachable message.

fosterseth (approved):

fosterseth:
In a separate PR, we should probably fail work units if service is unreachable. Unreachable could mean control service is not running on target node, or firewall has rejected traffic somewhere.

as it stands, receptor work units will remain in pending indefinitely (by design).

ghjm:
The concept of Workceptor was that you could launch a remote job against a node that isn't online yet, and the job will eventually succeed when the node comes on line.  The same thinking could apply to service reconfigurations.  A service being unreachable isn't necessarily a permanent condition - for all we know, the remote node is about to be reconfigured and might accept the job on the next retry.

fosterseth:
@ghjm yeah good point. Maybe instead of failing the job, we can put an "unreachable" message into the work Unit Detail, but keep it pending. Just some way to let the user know that the job is not going to start unless some configuration file is changed somewhere

PR #407, opened 2021-08-31
[DNM] Add basic sosreport plugin

kdelee:
Still need to do all the magic in packaging/rpm/receptor.spec.j2 to make it work

### DIFF
--- a/tools/sosreport/receptor.py
+++ b/tools/sosreport/receptor.py
@@ -1,0 +1,36 @@
+# Copyright (c) 2016 Ansible, Inc.
+# All Rights Reserved.
+
+try:
+    from sos.plugins import Plugin, RedHatPlugin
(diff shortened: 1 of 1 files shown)


kdelee:
Sounds like maybe all the RPM stuff is moving and this will move with it

rooftopcellist:
@kdelee if we are following the Controller pattern, then the sosreport plugin should live in the upstream codebase (https://github.com/ansible/awx/blob/devel/tools/sosreport/controller.py), so even when we move the `packaging/rpm/` dir, this would stay.  It could live in something like `packaging/sosreport/receptor.py`.  

This would need to be merged before we could test/build the downstream changes.

kdelee:
Closing as this is getting implemented elsewhere

Issue #416, opened 2021-09-09
Command to remove a receptor node


thenets:
For some use cases make sense to remove an older receptor node without the need to delete everything and restart the mesh state.

### Motivation

During my test cases for `execution_nodes` with the receptor, I'm facing some challenges trying to add and remove nodes to check the AWX auto-discovery behavior.

I always need to destroy all the mesh and start again. Would be great if I could delete the nodes without the need to destroy everything.

### Suggestion

I think that would be nice to have different ways to remove a receptor node from the mesh. If this is not possible to add those additional options, no problem, I can handle it from the Python code. 

Examples:
- Remove based on the name:
   `receptorctl remove <name>`
- Remove based on the prefix:
   `receptorctl remove --prefix <prefix_name>`
- Remove based on the age:
   `receptorctl remove --older_than <age_in_seconds>`
- Remove based on last seen:
   `receptorctl remove --last_seen <seconds>`

Time Sprint: 1 sprint

chrismeyersfsu:
related to https://github.com/ansible/tower-packaging/issues/1429 If/when we add uninstalling a mesh node from the mesh in AWX.

chrismeyersfsu:
@fosterseth does receptor have an notion of expiring a node from it's list of known node(s) ?

fosterseth:
Potentially a broadcast message that originates from the node being removed

PR #424, opened 2021-09-24
purge command to remove internal data related to disconnected nodes

fosterseth:
`receptorctl purge --all`
or
`receptorctl purge --node baz`

cleans up the following internal data structures in netceptor

```
knownConnectionCosts
routingPathCosts
knownNodeInfo
serviceAdsReceived
```

Useful to cleanup the status output

```
bar          {'fish': 1, 'foo': 1}
baz          {}
fish         {'bar': 1}
foo          {'bar': 1}
```

In this case `baz` is a disconnected node. After running purge

```
$ receptorctlfoo purge
Purged baz
```

in status output
```
bar          {'fish': 1, 'foo': 1}
fish         {'bar': 1}
foo          {'bar': 1}
```

Note: only affects the receptor node you run the purge command on.

### DIFF
--- a/pkg/netceptor/netceptor.go
+++ b/pkg/netceptor/netceptor.go
@@ -1670,6 +1670,56 @@ func (s *Netceptor) sendAndLogConnectionRejection(remoteNodeID string, ci *connI
 	return fmt.Errorf("rejected connection with node %s because %s", remoteNodeID, reason)
 }
 
+// PurgeUnreachableNodes cleans up internal data of nodes that are no longer on mesh.
+func (s *Netceptor) PurgeUnreachableNodes(node string) ([]string, error) {
+	s.knownNodeLock.Lock()
+	s.routingTableLock.Lock()
+	s.serviceAdsLock.Lock()
+	defer func() {
+		s.routingTableLock.Unlock()
+		s.knownNodeLock.Unlock()
+		s.serviceAdsLock.Unlock()
+	}()
+
+	// Find which nodes currently have routes on the mesh. These cannot be purged.
+	nodesReachable := make(map[string]struct{})
+	for k, v := range s.routingTable {
+		nodesReachable[k] = struct{}{}
+		nodesReachable[v] = struct{}{}
+	}
+	nodesReachable[s.nodeID] = struct{}{}
+
+	if _, ok := nodesReachable[node]; ok && node != "" {
+		return nil, fmt.Errorf("target node is still reachable on mesh")
+	}
+
+	// knownNodeInfo has a history of all seen nodes, whether they are reachable or not.
+	// Find nodes in this map that are not in nodesReachable. These will be purged.
+	nodesToPurge := []string{}
+	for k := range s.knownNodeInfo {
+		if _, ok := nodesReachable[k]; !ok && (node == "" || node == k) {
+			nodesToPurge = append(nodesToPurge, k)
+		}
+	}
+
+	if node != "" && len(nodesToPurge) == 0 {
+		return nil, fmt.Errorf("target node is not part of mesh")
+	}
+
+	// Delete the data.
+	for _, p := range nodesToPurge {
+		delete(s.knownNodeInfo, p)
+		delete(s.knownConnectionCosts, p)
+		for g := range s.knownConnectionCosts {
+			delete(s.knownConnectionCosts[g], p)
+		}
+		delete(s.routingPathCosts, p)
+		delete(s.serviceAdsReceived, p)
+	}
+
+	return nodesToPurge, nil
+}
+
 // Main Netceptor protocol loop.
 func (s *Netceptor) runProtocol(ctx context.Context, sess BackendSession, bi *backendInfo) error {
 	if bi.connectionCost <= 0.0 {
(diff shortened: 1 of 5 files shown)


receptorctl/receptorctl/cli.py
```diff
@@ -195,6 +195,30 @@ def reload(ctx):
         else:
             sys.exit(4)
 
+@cli.command(help="Purge runtime data related to disconnected nodes. Useful for cleaning up status output.")
+@click.option('--all', 'purgeall', help="All disconnected nodes will be removed", is_flag=True)
+@click.option('--unit_id', type=str, default="", help="Remote specific disconnected node")
+@click.pass_context
+def purge(ctx, purgeall, unit_id):
+    if purgeall is False and not unit_id:
```
fosterseth: @john-westcott-iv 

users must specify --all or --unit_id

great suggestion

fosterseth:
Testing this PR:

The following will setup this mesh `foo <---> bar <---> fish`

foo.yml
``` yaml
---
- node:
    id: foo
- tcp-listener:
    port: 2222
- control-service:
    service: control
    filename: /tmp/foo.sock
```

bar.yml
``` yaml
---
- node:
    id: bar
- tcp-listener:
    port: 2223
- tcp-peer:
    address: localhost:2222
    redial: true
```

fish.yml
``` yaml
---
- node:
    id: fish
- tcp-peer:
    address: localhost:2223
    redial: true
```

1. Open 3 terminals, and start each node, e.g. `receptor -c foo.yml`
2. Run `receptorctl --socket /tmp/foo.sock status` to see the mesh status. The Routing Table should display information related to each of the nodes and how they are connected.
3. To test, Control-C either bar or fish, and then run `receptorctl --socket /tmp/foo.sock purge --all`. This should remove whatever nodes are disconnected from the mesh. Keep in mind, since bar is the middle node in this set up, if you shutdown bar, foo and fish are disconnected even if fish is still running. A purge in this case would remove both bar and fish.
4. Also use the --node to specify a particular node to purge.

pkg/netceptor/netceptor.go
```diff
@@ -1656,6 +1656,86 @@ func (s *Netceptor) sendAndLogConnectionRejection(remoteNodeID string, ci *connI
...
+	nodesWithRoutes := make(map[string]struct{})
+	for k, v := range s.routingTable {
+		nodesWithRoutes[k] = struct{}{}
+		nodesWithRoutes[v] = struct{}{}
+	}
+	nodesWithRoutes[s.nodeID] = struct{}{}
+
+	if target != "" {
+		if _, ok := nodesWithRoutes[target]; ok {
+			return nil, fmt.Errorf("target node is not disconnected")
```
john-westcott-iv: To be consistent with the params do we want to say "target unit_id is not disconnected"? If so, this is in a couple other strings above and below as well.

receptorctl/receptorctl/cli.py
```diff
@@ -187,18 +187,45 @@ def ping(ctx, node, count, delay):
...
         elif "ERRORCODE 4" in results['Error']:
+            # problem loading the config file, probably a syntax error
             sys.exit(4)
         else:
+            # problem loading the config file, probably a syntax error
             sys.exit(4)
 
+@cli.command(help="Purge runtime data related to disconnected nodes. Useful for cleaning up status output.")
+@click.option('--all', 'purgeall', help="All disconnected nodes will be removed", is_flag=True)
+@click.option('--unit_id', type=str, default="", help="Remote specific disconnected node")
```
john-westcott-iv: Should this be "Remove specific...."

receptorctl/receptorctl/cli.py
```diff
@@ -187,18 +187,45 @@ def ping(ctx, node, count, delay):
...
     else:
         print(f"Error: {results['Error']}")
         if "ERRORCODE 3" in results['Error']:
+            # config file was fine, but some changes were non-reloadable
             sys.exit(3)
         elif "ERRORCODE 4" in results['Error']:
+            # problem loading the config file, probably a syntax error
             sys.exit(4)
         else:
+            # problem loading the config file, probably a syntax error
```
john-westcott-iv: Should these changes be in this PR or a separate one? It looks like its reload related and not purge related.

pkg/netceptor/netceptor.go
fosterseth: calling the parameter `unit_id` was a mistake, it should have been `--node`. I've changed it

receptorctl/receptorctl/cli.py
fosterseth: fixed

receptorctl/receptorctl/cli.py
fosterseth: yeah not related, I just wanted to tidy up the code.

yagomarques (approved):

john-westcott-iv:
Addresses #416

kdelee (approved):

kdelee:
@jbradberry is this of interest again, especially considering what you were working on with registering/de-registering peers in awx?

jbradberry:
@kdelee not a huge amount?  The auto-discovery mechanism that relied on getting data from Receptor has been chopped out of AWX.  For now, the only entry point for addition and removal of node data in AWX is via the `provision_instance` and `register_peers` commands.

fosterseth:
closing for now, we may need to return to this at some point.

PR #429, opened 2021-09-28
Remove unneeded packaging directory

rooftopcellist:
The packaging bits are now handled elsewhere, so this is no longer needed.


shanemcd:
Can you please also update the Makefile and remove anything related to this directory?

rooftopcellist:
:+1: I just removed the relevant make targets and ignore entries.  I don't see anything else related to `packaging/rpm`.

shanemcd (approved):
Thank you!

PR #430, opened 2021-09-30
Verify signature for work cancel, release, and results

fosterseth:
This extends the work signing and verification to `work cancel`, `work release`, and `work results`

In the case of `work submit`, clients are required to issue a "signwork=true" as part of the payload (if the target work expects signed work). Since the cancel, release, and results commands operate on an _existing_ work unit, the clients don't need to include a "signwork=true" as part of the payload. Instead, receptor will track which units should expect signed work, and automatically sign the commands.

Another big change -- we don't enforce verification if the commands are coming from a unix socket. This should be helpful for debugging purposes.

Additionally, work will now fail if the client sends a signature and the receiver was not expecting one.

Testing this PR:

1. Start 3 node mesh: e.g. foo, baz, fish
2. foo should have a work-signing set up with a private key
3. fish should have a work-verification set up with the public key. verifySignature set to true on a work-command, `echosleep`
4. On foo, submit work on fish
5. Now, on bar, use the connect command, `connect fish control`
6. Before this PR, we could do a `work cancel`, `work results`, or `work release` here and succeed. This is bad. With this PR, work cancel, results, or release should fail.

### DIFF
--- a/pkg/workceptor/controlsvc.go
+++ b/pkg/workceptor/controlsvc.go
@@ -168,6 +168,10 @@ func (t *workceptorCommandType) InitFromJSON(config map[string]interface{}) (con
 		if err != nil {
 			return nil, err
 		}
+		signature, err := strFromMap(config, "signature")
+		if err == nil {
+			c.params["signature"] = signature
+		}
 	case "list":
 		unitID, err := strFromMap(config, "unitid")
 		if err == nil {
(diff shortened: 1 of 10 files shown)


pkg/workceptor/controlsvc.go
```diff
@@ -182,13 +186,37 @@ func (t *workceptorCommandType) InitFromJSON(config map[string]interface{}) (con
...
+	}
+
+	return nil
+}
+
 // Worker function called by the control service to process a "work" command.
 func (c *workceptorCommand) ControlFunc(nc *netceptor.Netceptor, cfo controlsvc.ControlFuncOperations) (map[string]interface{}, error) {
+	addr := cfo.RemoteAddr()
+	connIsUnix := false
+	if addr.Network() == "unix" {
```
fosterseth, line 217: @shanemcd we can do this to determine the origin of the connection to this control service.

if it's from a remote node on the mesh, Network will be "netceptor-foo". If it's from a tcp connection, it will be "tcp", and if from the local unix domain socket, "unix".

It should be safe to exempt signature verification in the case of unix socket. That is, we inherently trust any commands coming from the local domain socket.

shanemcd (approved):

PR #433, opened 2021-10-01
sleep longer before killing cmd process

jladdjr:
We are seeing cases where receptor is sent a `cancel` command and the `ansible-runner` process is exiting but is leaving behind a running container. Increasing the amount of time we sleep from one second to ten to give ansible-runner (and work units in general) longer to cleanup before forcefully killing the process.

### DIFF
--- a/pkg/workceptor/command.go
+++ b/pkg/workceptor/command.go
@@ -38,7 +38,7 @@ func termThenKill(cmd *exec.Cmd) {
 		return
 	}
 	_ = cmd.Process.Signal(os.Interrupt)
-	time.Sleep(1 * time.Second)
+	time.Sleep(10 * time.Second)
 	if cmd.Process != nil {
 		_ = cmd.Process.Kill()
 	}


jladdjr:
@fosterseth - this seems like it might be a reasonable change as is. Would you be opposed to us going ahead and merging this?

beeankha (approved):
:+1:

jladdjr:
Noting that @fosterseth mentioned that increasing the sleep here could have side effects on other parts of receptor:

> @jladdjr @chrismeyersfsu  waiting 10 seconds for a cancel command may or may not introduce problems. A lot of our tests have time limits to asserting things are cancelled etc. and not sure adding 9 seconds is going to work well. I just explored adding this as a parameter we can pass in, but for the reasons I outlined above, this is not trivial probably the most straight forward thing would be to pass this param in during work submission. That way the subprocess has this information when it comes time to issue cancel (sigINT, sigKILL)

chrismeyersfsu:
The spirit of this change is good. We want to wait some time before sending a `SIGKILL`. @fosterseth  is going to follow up this PR with another one.

jladdjr:
superseded by https://github.com/ansible/receptor/pull/440

Issue #435, opened 2021-10-04
[Tech Debt] Make changing log level in receptor config a reloadable attribute


kdelee:
Make changing log level in receptor a re-loadable (no restart required) config file attribute

This is important for debugging receptor mesh in production that may be in a bad state, because if we must restart receptor to get more verbose logs, this may eliminate the bad state, which makes it mysterious...and unless we leave it in debug (which is not reccommended for production), we may never get to find out what is going on.

kdelee:
https://github.com/ansible/receptor/pull/437 never made it in -- any chance this could get revived?

PR #437, opened 2021-10-07
Adding Log Level to reloadable config

amolgautam25:
This PR is for the issue : https://github.com/ansible/receptor/issues/435

### DIFF
--- a/pkg/backends/websockets.go
+++ b/pkg/backends/websockets.go
@@ -355,11 +355,11 @@ func (cfg websocketListenerCfg) PreReload() error {
 	return cfg.Prepare()
 }
 
-func (cfg websocketDialerCfg) Reload() error {
+func (cfg websocketDialerCfg) ReloadBackend() error {
 	return cfg.Run()
 }
 
-func (cfg websocketListenerCfg) Reload() error {
+func (cfg websocketListenerCfg) ReloadBackend() error {
 	return cfg.Run()
 }
 
--- a/pkg/controlsvc/reload.go
+++ b/pkg/controlsvc/reload.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"strings"
+	"sync"
 
 	"github.com/ansible/receptor/pkg/logger"
 	"github.com/ansible/receptor/pkg/netceptor"
@@ -16,34 +17,29 @@ type (
 	reloadCommand     struct{}
 )
 
-var configPath = ""
+var (
+	configPath       = ""
+	mu               sync.Mutex
+	cfgPrevious      = make(map[string]struct{})
+	cfgNext          = make(map[string]struct{})
+	backendModified  = false
+	loglevelModified = false
+	loglevelPresent  = false
+)
 
 var reloadParseAndRun = func(toRun []string) error {
-	return fmt.Errorf("no configuration file was provided, reload function not set")
+	return fmt.Errorf("no configuration file was provided")
 }
 
-var cfgNotReloadable = make(map[string]bool)
-
-var reloadableActions = []string{
-	"tcp-peer",
-	"tcp-listener",
-	"ws-peer",
-	"ws-listener",
-	"udp-peer",
-	"udp-listener",
-	"local-only",
-}
-
-func isReloadable(cfg string) bool {
-	// checks if top-level keys (e.g. tcp-peer) are in the list of reloadable
-	// actions
-	for _, a := range reloadableActions {
-		if strings.HasPrefix(cfg, a) {
-			return true
-		}
-	}
-
-	return false
+var reloadableActions = map[string]*bool{
+	"tcp-peer":     &backendModified,
+	"tcp-listener": &backendModified,
+	"ws-peer":      &backendModified,
+	"ws-listener":  &backendModified,
+	"udp-peer":     &backendModified,
+	"udp-listener": &backendModified,
+	"local-only":   &backendModified,
+	"log-level":    &loglevelModified,
 }
 
 func getActionKeyword(cfg string) string {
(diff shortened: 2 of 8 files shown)


chrismeyersfsu:
Does this sort of "just work" or does more work need to be done?

pkg/logger/logger.go
```diff
@@ -143,6 +143,19 @@ func (cfg traceCfg) Prepare() error {
...
 }
 
+func (cfg loglevelCfg) ReloadLogger() error {
+
+	return cfg.Init()
+}
+
+func (cfg loglevelCfg) InitLogger() error {
+
+	fmt.Println("We are now resetting the logger")
```
amolgautam25: The flow of execution is not able to go inside this function. 


```DEBUG 2021/11/29 10:43:10 Reloading
toRun map[InitLogger:{}]
> github.com/ansible/receptor/pkg/controlsvc.(*reloadCommand).ControlFunc() ./pkg/controlsvc/reload.go:197 (hits goroutine(75):1 total:4) (PC: 0xbb6117)
   192:		for callableStr := range toRun {
   193:			toRunStr = append(toRunStr, callableStr)
   194:		}
   195:		// reloadParseAndRun is a ParseAndRun closure, set in receptor.go/main()
   196:		fmt.Printf("toRun %v\n", toRun)
=> 197:		err = reloadParseAndRun(toRunStr)
   198:		if err != nil {
   199:			return handleError(err, 4)
   200:		}
   201:
   202:		// set old config to new config, only if successful
(dlv) p toRunStr
[]string len: 1, cap: 1, [
	"InitLogger",
]
(dlv)```

File: pkg/controlsvc/reload.go [hunk 2]
```diff
@@ -159,37 +152,65 @@ func handleError(err error, errorcode int) (map[string]interface{}, error) {
...
+	}
+
+	if !loglevelPresent {
+		logger.InitLogger()
 	}
 
-	nc.CancelBackends()
 	// reloadParseAndRun is a ParseAndRun closure, set in receptor.go/main()
-	err = reloadParseAndRun([]string{"PreReload", "Reload"})
+	fmt.Printf("toRun %v\n", toRun)
```
fosterseth: let's go ahead and drop this print

amolgautam25: done.

eqrx (changes requested):
Could we get function documentation in there? I find it hard to figure out which function does what and docs would really help out here

eqrx:
also ... are all the changes only to change the log level or has this PR something piggybacked?

amolgautam25:
Docs PR : https://github.com/ansible/receptor/pull/523

pkg/backends/websockets.go
```diff
@@ -355,11 +355,11 @@ func (cfg websocketListenerCfg) PreReload() error {
 	return cfg.Prepare()
 }
 
-func (cfg websocketDialerCfg) Reload() error {
+func (cfg websocketDialerCfg) ReloadBackend() error {
```
eqrx, line 358: Isn't the name redundant since the method is part of backend config?

pkg/controlsvc/reload.go
```diff
@@ -16,34 +17,29 @@ type (
 	reloadCommand     struct{}
 )
 
-var configPath = ""
+var (
+	configPath       = ""
+	mu               sync.Mutex
```
eqrx, line 22: A global mutex is allowed but I would suggest to bundle it up with all the other vars here and create a struct to allow grouping and encapsulation.

pkg/controlsvc/reload.go
```diff
@@ -159,37 +152,67 @@ func handleError(err error, errorcode int) (map[string]interface{}, error) {
...
+	defer resetAfterReload()
 
-	// Do a quick check to catch any yaml errors before canceling backends
+	cfr := make(map[string]interface{})
+	cfr["Success"] = true
+
+	// do a quick check to catch any yaml errors before canceling backends
 	err := reloadParseAndRun([]string{"PreReload"})
 	if err != nil {
 		return handleError(err, 4)
```
eqrx, line 168: what is error code #4? aka: how about making error codes constants?

pkg/controlsvc/reload.go
```diff
@@ -16,34 +17,29 @@ type (
 	reloadCommand     struct{}
 )
 
-var configPath = ""
+var (
+	configPath       = ""
+	mu               sync.Mutex
+	cfgPrevious      = make(map[string]struct{})
```
eqrx, line 23: Why are configuration items empty structs?

pkg/controlsvc/reload.go
```diff
@@ -88,53 +71,63 @@ func parseConfigForReload(filename string, checkReload bool) error {
...
-	// checks to see if any item in cfgNotReloadable has a value of false,
-	// if so, that means an unreloadable item has been removed from the config
-	defer func() {
-		for k := range cfgNotReloadable {
-			cfgNotReloadable[k] = false
+func checkReload() error {
+	// Determine which items from the old config have been added or modified
+	for cfg := range cfgNext {
+		action := getActionKeyword(cfg)
+		_, isReloadable := reloadableActions[action]
```
eqrx, line 84: This checks if `action` is a key in `reloadableActions`, not if it is reloadable. Other name perhaps?

pkg/controlsvc/reload.go
```diff
@@ -88,53 +71,63 @@ func parseConfigForReload(filename string, checkReload bool) error {
...
+		_, inPrevious := cfgPrevious[cfg]
+		if !isReloadable && !inPrevious {
+			return fmt.Errorf("a non-reloadable config action '%s' was added or modified. Must restart receptor for these changes to take effect", action)
+		}
+		if isReloadable && !inPrevious {
+			*reloadableActions[action] = true
```
eqrx, line 90: In line #34 you assign the same bool via pointer to multiple actions. Setting one of those to true will set all of them to true. Is this intended?

eqrx (changes requested):

pkg/backends/websockets.go
amolgautam25: Well, I thought, in the future, we might add more types of Reload. Like ```ReloadLogger()```. It would be better if we specify what kind of reload this is.

pkg/controlsvc/reload.go
amolgautam25: Initially, the configurations are empty structs. While a reload is being invoked, it saves the current configuration( in cfgPrevious) and parses the new configuration. Then a the new configuration is checked against the old configuration to figure out what all things have changed and accordingly decide what reload method to call.

pkg/controlsvc/reload.go
amolgautam25: yes, this is intended.

pkg/controlsvc/reload.go
eqrx: Sorry I was unclear. The type definition of the map value is a struct with no fields. Why is that not a concrete type ... or `interface{}` if it needs to be purely dynamic?

PR #440, opened 2021-10-13
extend time between sig INT and sig KILL

fosterseth:


pkg/workceptor/command.go
```diff
@@ -33,13 +33,19 @@ type commandExtraData struct {
...
+func termThenKill(cmd *exec.Cmd, doneChan chan bool) {
 	if cmd.Process == nil {
 		return
 	}
 	_ = cmd.Process.Signal(os.Interrupt)
-	time.Sleep(1 * time.Second)
+	select {
+	case <-doneChan:
+		return
+	case <-time.After(10 * time.Second):
```
jladdjr, line 44: will this poll or just wait 10 seconds?

shanemcd (approved):

PR #445, opened 2021-10-15
Receptorctl tests refactor

thenets:
1. The original code:
- Move fixture to `conftest.py `
- Move original tests to `test_connection.py`

---

2. Code quality improvement:
- Complete new fixtures structure
- Create test files with the `test_` prefix
- Use `subprocess.check_output()` instead of `os.system()`. This force a raise exception when the `rc` is not 0
- Allow selecting the `receptor` binary from different places
- Test CLI with `click` library

---

Tests executed locally with `make receptorctl-tests`:

image


shanemcd:
Please see my feedback from #447

Makefile
```diff
@@ -171,7 +171,9 @@ receptorctl-test-venv/bin/pytest:
 	receptorctl-test-venv/bin/pip install -r receptorctl/test-requirements.txt
 
 receptorctl-tests: receptor receptorctl-test-venv/bin/pytest
-	cd receptorctl && ../receptorctl-test-venv/bin/pytest tests/tests.py
+	cd receptorctl && \
+		PATH=$(PATH):$(PWD) \
+			../receptorctl-test-venv/bin/pytest tests/
```
kdelee: nitpick: this whitespace looks a bit wacky. I think you are using tabs, maybe make it all align with `receptorctl` or both the lines after receptorctl line up with eachother.

thenets: Hmm, I thought this would make more sense this way, cause the `175` and `176` are the same command.
I'll update and change this and leave `176` at the same level as `175`, but I believe that they should not be at the same level as `174`, cause it will look like different commands.

The result will be:

```
cd receptorctl && ../receptorctl-test-venv/bin/pytest tests/tests.py
cd receptorctl && \
	PATH=$(PATH):$(PWD) \
	../receptorctl-test-venv/bin/pytest tests/
```

thenets:
@shanemcd problem solved. I fixed the problem invoking the command in a different way.

shanemcd (approved):

PR #456, opened 2021-10-26
Update dependencies

eqrx:
Module dependencies are outdated. Updating them to avoid fighting solved problems


.github/workflows/pull_request.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.15
+          go-version: 1.17
```
shanemcd, line 27: Honest question, would it be better to take the approach I took in https://github.com/ansible/receptor/pull/491 (match the version in RHEL)? Does it not matter? Would appreciate your insight.

shanemcd:
Can you rebase this to pull in the CI changes?

PR #458, opened 2021-10-26
Fix goroutine leak

fosterseth:


kdelee (approved):
I tested this with local jobs (project updates in awx/tower) and saw significant improvement, in that a workload that previously caused receptor to fall over ceased to exhibit the goroutine leak and receptor stayed stable under the workload.

fosterseth:
https://github.com/ansible/receptor/pull/459

PR #463, opened 2021-11-03
always read from broker, and let broker close monitorUnreachable

fosterseth:
fixes a variety of spots in the code where we are expecting to write a channel that is never read from, because the goroutine that reads has already exited due to a cancelled context.


fosterseth:
@Zokormazo 

I think this patch seems promising for fixing the nodes connection issue after 10 minutes of running jobs

pkg/netceptor/conn.go
```diff
@@ -372,17 +372,14 @@ func (s *Netceptor) DialContext(ctx context.Context, node string, service string
...
-			return
-		case <-doneChan:
-			return
-		case msg := <-msgCh:
-			if msg.Problem == ProblemServiceUnknown && msg.ToNode == remoteAddr.node && msg.ToService == remoteAddr.service {
-				logger.Error("remote service unreachable")
-				cancel()
-			}
+	if msgCh == nil {
+		return
```
chrismeyersfsu, line 378: Should `cancel()` be called here too?

fosterseth: good call, we probably should

chrismeyersfsu (approved):

beeankha (approved):

thenets (approved):

PR #465, opened 2021-11-03
Upload binary to S3 instead of archiving with Actions

shanemcd:


eqrx (changes requested):
This will break all actions running on forks, could you split this workflow into push and pull_request triggered ones?

.github/workflows/pull_request.yml
```diff
@@ -41,11 +41,14 @@ jobs:
...
-          name: receptor
-          path: receptor
+      - name: Upload binary
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        run: |
+          pip install boto3
+          ansible -i localhost, -c local all -m aws_s3 \
+            -a "bucket=receptor-nightlies object=PRs/${{github.event.number}}/receptor src=./receptor mode=put"
```
eqrx, line 51: Doesn't this also allow any github user to upload arbitrary source code to be build by actions and uploaded into the bucket?

eqrx (changes requested):

fosterseth:
we have this solution instead https://github.com/ansible/receptor/pull/466
probably can close this PR

Issue #428, opened 2021-09-28
add verification to work cancel and work release


fosterseth:
if verifySignature is set to true, only authorized nodes are able to start that work unit. However, it is still the case that any node on the mesh could cancel and release that work unit.

We should extend the work signing feature to only allow authorized nodes to cancel those work units.

chrismeyersfsu:
@fosterseth should this be in `needs_test` ?

fosterseth:
https://github.com/ansible/receptor/pull/430

thenets:
`receptorctl` is failing. I added in the PR the error message:
- https://github.com/ansible/receptor/pull/468

shanemcd:
Why was this moved back to needs_devel?

thenets:
@shanemcd the test that I created `test_work_signed_expect_block` is failing:
- https://github.com/ansible/receptor/pull/468

We need to fix the code. That's why I moved back to `needs_devel`.

thenets:
@shanemcd I need to have that PR merged. I talked to @fosterseth and we decided to comment the test that was failing until the patch that fixes the issue.

Can you help us to review and merge it? We have many PR already open.

fosterseth:
implemented here https://github.com/ansible/receptor/pull/430

PR #468, opened 2021-11-04
Add test `work cancel` and `work release`

thenets:
DO NOT Fixes anymore https://github.com/ansible/receptor/issues/428

Many more fixtures were included or modified in order to support multiple meshes deployments.

Added two tests:
- 🟢 test_work_unsigned 
- [disabled/commented] 🔴 test_work_signed_expect_block
  - This is failing, cause when the `work submit` fails, it doesn't return the `work_unit_id`.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -106,10 +106,6 @@ jobs:
       - name: Run receptor tests
         run: make test
 
-      - name: Remove sockets for artifacting
-        if: ${{ failure() }}
-        run: "find /tmp/receptor-testing -type s -exec /bin/rm -f {} \\;"
-
       - name: get k8s logs
         if: ${{ failure() }}
         run: .github/workflows/artifact-k8s-logs.sh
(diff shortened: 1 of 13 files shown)


thenets:
@shanemcd @fosterseth hello! 😄 

- I removed the `test_work_signed_expect_block` until the fix be pushed to the `receptor` code base. I believe that we need to have a discussion in the future about merging broken tests. I can't understand why not merge a test that will break the pipeline, because the tests are made to raise those problems as soon as possible.

- I also removed the `receptorctl-lint`. We can't have a blocker step in the pipeline without even having a discussion before that. Things like length are not configured for `black` and require manual "fixes", for example. We can enable this lint but in a dedicated PR, because the current state is causing more harm than helping.

Please, help me to review and merge this. I need this PR merge in order to add more test that requires multiple meshes.

shanemcd:
@thenets The whole point of adding black is to remove the need for discussing code styles. To fix your code, just run `black receptorctl`.

thenets:
@shanemcd doesn't work. You can clone the branch and try.

I agree that we should add this, but in a dedicated PR, without blocking existent test code.

shanemcd (approved):

PR #469, opened 2021-11-05
Fix get results returning when unit failed

fosterseth:
Ansible-runner will return a non-zero code if the tasks failed. This results in a receptor unit with a Failed status State. We still want to get the full stdout from this ansible playbook, thus we shouldn't early-exit from GetResults.

This early exit was mostly put in place to prevent a situation where we exited out of monitorRemoteStdout in remote_work.go, but hung around in GetResults indefinitely because Status reported a stdout size greater than what we had on disk. This scenario is probably less of an issue for AWX, as it will close out of the socket when a job is cancelled. This closing of the socket should trigger the GetResults to return.

### DIFF
--- a/pkg/workceptor/workceptor.go
+++ b/pkg/workceptor/workceptor.go
@@ -524,9 +524,6 @@ func (w *Workceptor) GetResults(ctx context.Context, unitID string, startPos int
 				return
 			}
 			for {
-				if unit.Status().State == WorkStateFailed {
-					return
-				}
 				select {
 				case <-ctx.Done():
 					return


jladdjr (approved):

beeankha (approved):

AlanCoding:
Added @fosterseth 's remote to the repo. You tripped me up naming it sockceptor! With this branch:

```
tests/api/inventories/test_smart_inventory.py::TestSmartInventory::test_source_inventory_variables_ignored PASSED [100%]
```

with `release_1.1`

```
tests/api/inventories/test_smart_inventory.py::TestSmartInventory::test_source_inventory_variables_ignored FAILED [100%]

============================================= FAILURES =============================================
____________________ TestSmartInventory.test_source_inventory_variables_ignored ____________________
  File "/home/alancoding/repos/tower-qa/tests/api/inventories/test_smart_inventory.py", line 390, in test_source_inventory_variables_ignored
    assert job.status == 'failed'
AssertionError: assert 'error' == 'failed'
  - failed
  + error
-----------
```

It's pretty well beyond dispute that the removed code here was closing the socket, which lead to the empty reads.

AlanCoding (approved):

chrismeyersfsu (approved):

chrismeyersfsu (approved):

Issue #423, opened 2021-09-22
append conntype string representation in status output


fosterseth:
conntype is 0,1,2 depending on if service expects datagram, stream, or streamTLS connection

For keeping serviceAds as small as possible, we should still keep these at bytes 0,1,2

However, I think the output in of a status command can append the semantic representations

I am finding that we are hardcoding in a various receptor clients (receptorcontrol, awx) the mapping between conntype and meaning, so we should do that in a single place only, for better maintainability.

this would look similar to how we append work unit State name (e.g. succeeded, failed, etc) to the output

https://github.com/ansible/receptor/blob/156e6e24a49fbf868734507f9943ac96208ed8f5/pkg/workceptor/workceptor.go#L380

if we do this, we should then make use of it in awx, receptorcontrol

PR #471, opened 2021-11-09
Modify Status command to return a ConnTypeLabel in addition to the ConnType byte

john-westcott-iv:
Addresses #423 

Includes modification of receptorctl to show the label instead of the status.

### DIFF
--- a/pkg/netceptor/netceptor.go
+++ b/pkg/netceptor/netceptor.go
@@ -218,16 +218,23 @@ type routingUpdate struct {
 }
 
 const (
-	// If adding/changing an ConnType, make sure to change ConnTypeStrings as well. These are friendly strings printed out in a status command
+	// If adding/changing an ConnType, make sure to change ConnTypeStrings as well.
+	//    These are friendly strings printed out in a status command.
 	// ConnTypeDatagram indicates a packetconn (datagram) service listener.
 	ConnTypeDatagram = 0
 	// ConnTypeStream indicates a conn (stream) service listener, without a user-defined TLS.
 	ConnTypeStream = 1
 	// ConnTypeStreamTLS indicates the service listens on a packetconn connection, with a user-defined TLS.
 	ConnTypeStreamTLS = 2
+	// Default Label for an unknown connection type.
+	UnknownConnTypeStr = "Unknown"
 )
 
-var ConnTypeStrings = [...]string{"Datagram", "Stream", "StreamTLS"}
+var ConnTypeStrings = map[byte]string{
+	ConnTypeDatagram:  "Datagram",
+	ConnTypeStream:    "Stream",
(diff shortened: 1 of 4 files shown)


pkg/netceptor/netceptor.go
```diff
@@ -221,6 +222,8 @@ const (
 	ConnTypeStreamTLS = 2
 )
 
+var ConnTypeStrings = [...]string{"Datagram", "Stream", "StreamTLS"}
```
fosterseth: what if this was a map instead?

`connTypeStrings := map[byte] string {ConnTypeDatagram: "DataGram" ... }`

then in GetConnectionTypeAsString() we do

```
conntypestr, ok := connTypeString[conntype]
if !ok {
conntypestr = "unknown"
}
```

john-westcott-iv: I like the idea of the map because then the labels are not dependent on the order of the array.

pkg/netceptor/netceptor.go
```diff
@@ -432,6 +443,17 @@ func (s *Netceptor) MaxConnectionIdleTime() time.Duration {
 	return s.maxConnectionIdleTime
 }
 
+// Convert the connection type to a string.
+func (s *Netceptor) GetConnectionTypeAsString(connectionType byte) string {
+	// A byte can't be < 0 so we don't need to check the lower bounds
+	connTypeString, ok := ConnTypeStrings[connectionType]
+	if !ok {
+		connTypeString = UnknownConnTypeStr
```
shanemcd, line 456: Curious, under which conditions would this happen?

john-westcott-iv: If someone added a new connection type constant but forgot to add an entry into the map. I was also thinking of some weird upgrade scenario where a newer version of receptor with a new connection type was introduced to an existing mesh.

pkg/netceptor/netceptor.go
```diff
@@ -235,12 +245,13 @@ type WorkCommand struct {
...
-	Service      string
-	Time         time.Time
-	ConnType     byte
-	Tags         map[string]string
-	WorkCommands []WorkCommand
+	NodeID        string
+	Service       string
+	Time          time.Time
+	ConnType      byte
+	ConnTypeLabel string
```
ghjm, line 252: This probably shouldn't go here.  This is the struct for the low-level advertisements that get sent around the network, so its size matters.  If we want another field in the Status response, we should define another struct that inherits this one and adds the field, and populate it at the time the status is being asked for.

AaronH88:
Hey John,

Thanks for the PR man, Im closing up old ones and see this hasnt been touched in a while, 
Feel free to reopen if you ever get time to address the comments 😄

PR #473, opened 2021-11-10
k8s createPod panic

fosterseth:
Under some conditions the event returned from UnitWithSync from the k8s pod watcher is nil, so deferencing it leads to a panic.

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -209,6 +209,9 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
 		ctxPodReady, _ = context.WithTimeout(kw.ctx, kw.podPendingTimeout)
 	}
 	ev, err := watch2.UntilWithSync(ctxPodReady, lw, &corev1.Pod{}, nil, podRunningAndReady)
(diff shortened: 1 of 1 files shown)


shanemcd (approved):

john-westcott-iv (approved):
I was going to ask about just checking err but it looks like there is a possibility in the function for it to return (nil, nil).

chrismeyersfsu:
I think ideally we would check `err` being returned. I could imagine the k8s library that we are using changing it's implementation to maybe where an `err` could be returned along `ev`, pointing to an invalid memory address. They could reason this is OK by saying "Well, we returned an `err`, so why are you using `ev`?" 

That all being said, they probably won't because it's just bad best practice. More importantly, we've tested THIS change so don't touch it. And with that, I approve this merge.

chrismeyersfsu (approved):

fosterseth:
the error handling below the UntWithSync looks like it assumes we got some valid ev object back, and that we were able to do a type assertion for kw.pod (i.e. not sure what `kw.pod.Status.ContainerStatuses` will do if we didn't get a good ev object back).

that is why I duck out early if we don't get a good looking ev object back.

PR #474, opened 2021-11-11
Change how receptor handles an undefined host id

### LINKED ISSUE #470: remove "localhost" as a reserved target
related https://github.com/ansible/receptor/pull/146

Receptor doesn't work if your `os.hostname` begins with "localhost", but it's very common for VMs and whatnot to be setup with localhost* as the hostname.

We should either strip this out, or rethink it to avoid that problem

john-westcott-iv:
Fixes #470
This implementation will create a change in the existing behavior.

 Previously it attempted to take the hostname of a machine as the ID and would fail if the hostname matched /^localhost.?.*$/.
 
Now it will generate a node name like `node-<8 character string>` followed by an optional @<identifier>. The first identifier attempted will be the hostname. However, if the hostname matches the regex above it will scan the interfaces looking for a non-loopback IP and use @<ip> instead of @<hostname>. If we fail for any reason (can't get the host name or cant get an address of an interface) it will drop the @ notation and just use `node-<8 character string>`.
 
The 8 character string itself should have been enough to fix the issue but I was attempting to get some kind of machine info so that you know where the node is coming from. i.e. in `receptorctl status` it will show something like:
```
Known Node                  Known Connections
node-UuUTDwBH@10.0.0.34     {'node-dCdDbCqd@ansible-devel': 1}
node-dCdDbCqd@ansible-devel {'node-UuUTDwBH@10.0.0.34': 1}
```
This would hopefully give an admin an idea of where those nodes connected from.

This fact the node ID is missing and one had to be auto-generated is indicated in the logs as :
`WARNING 2021/11/11 09:47:13 node id is not set in the serve config generated an id of node-dCdDbCqd@ansible-devel`

Finally, this enables you to launch multiple receptor processes on the same node without an ID as the likely hood of the 8 character string being duplicated is very low. Previously, both processes would attempt to use the hostname as their identifier causing the second process will be kicked out with the error:
`ERROR 2021/11/11 10:31:33 Backend error: rejected connection with node ansible-devel because it tried to connect using our own node ID`

### DIFF
--- a/cmd/receptor-cl/receptor.go
+++ b/cmd/receptor-cl/receptor.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ansible/receptor/pkg/logger"
 	"github.com/ansible/receptor/pkg/netceptor"
 	_ "github.com/ansible/receptor/pkg/services"
+	"github.com/ansible/receptor/pkg/utils"
 	"github.com/ansible/receptor/pkg/workceptor"
 	"github.com/ghjm/cmdline"
 )
@@ -27,15 +28,7 @@ type nodeCfg struct {
 func (cfg nodeCfg) Init() error {
 	var err error
 	if cfg.ID == "" {
-		host, err := os.Hostname()
-		if err != nil {
-			return err
-		}
-		lchost := strings.ToLower(host)
-		if lchost == "localhost" || strings.HasPrefix(lchost, "localhost.") {
-			return fmt.Errorf("no node ID specified and local host name is localhost")
-		}
-		cfg.ID = host
+		cfg.ID = utils.GenerateHostID()
 	}
 	if strings.ToLower(cfg.ID) == "localhost" {
 		return fmt.Errorf("node ID \"localhost\" is reserved")
--- a/pkg/pkg.go
+++ b/pkg/pkg.go
@@ -4,14 +4,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os"
-	"strings"
 
 	"github.com/ansible/receptor/pkg/backends"
 	"github.com/ansible/receptor/pkg/controlsvc"
 	"github.com/ansible/receptor/pkg/logger"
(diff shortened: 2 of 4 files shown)


john-westcott-iv:
recheck

shanemcd:
If I'm reading this code correctly, under circumstances, Receptor will generate a random node ID each time it's started?

john-westcott-iv:
> If I'm reading this code correctly, under circumstances, Receptor will generate a random node ID each time it's started?

That is correct, if your hostname is localhost and you don't pass in a node ID in the config each time receptor starts it will get a random hostname.

shanemcd:
I'm sorry you wasted time on this, I don't think that's something we want to do.

shanemcd:
Reopened this after reading Seth's comment here: https://github.com/ansible/receptor/issues/470#issuecomment-971101974

john-westcott-iv:
@shanemcd per the comments on #470 did we want to change the random part of the generated ID in this PR or do we want to go the way @fosterseth was describing and just change how the workceptor processes get an ID and not use any part of this PR or some other combination? I'm asking because it feels like there are multiple layers to this problem and each layer may have multiple fixes. The solution in this PR solves the scenario of allowing one or more processes to start without a node ID on a single node regardless of the hostname (localhost or otherwise). The random part of the hostname allows for the "or more" part of that statement. It does have the drawback that if a single process is supposed to be on a machine and that particular process is restarted over and over it will have a new node ID every time. But, if you were trying to start multiple processes and we had a fixed name subsequent processes would end up running but unable to join a mesh because they had duplicate names.

shanemcd:
The main thing I want to avoid is generating a random node id for nodes that _might_ connect to a mesh.

I am fine with either of Seth's suggestion of passing through the parent node id to the `--command-runner` process, or generating a random node id in just that specific location.

shanemcd:
This issue was addressed by https://github.com/ansible/receptor/pull/501

PR #478, opened 2021-11-11
Adding warning to all commands in receptorctl

### LINKED ISSUE #477: Add version warning to all of receptorctl
If you run `receptorctl --socket /tmp/foo.sock version` you can get an error indicating if the version of receptor and receptorctl are different such as:
```
	receptorctl  0.9.8.dev194
	receptor     
```

We should make this a general check/warning for all commands in receptorctl.

john-westcott-iv:
Fixes #477

This moves the version check into the code that initialized the ReceptorControl so that all commands will warn if the versions are out of sync. Also, move the warning to std.err instead of std.out. Adding checking to make sure socket option is passed before actually making the connection.

### DIFF
--- a/receptorctl/receptorctl/cli.py
+++ b/receptorctl/receptorctl/cli.py
@@ -85,12 +85,12 @@ def print_error(message, nl=True):
 )
 def cli(ctx, socket, config, tlsclient, rootcas, key, cert, insecureskipverify):
     ctx.obj = {
-        'rc': None,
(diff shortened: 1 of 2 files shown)


shanemcd (approved):

shanemcd:
@john-westcott-iv Looks like there is a conflict here. Once you fix that we'll get this merged.

john-westcott-iv:
@shanemcd rebased.

PR #481, opened 2021-11-15
Updating printing

john-westcott-iv:
Changing from `print` to `click.echo`.
Adding printWarning method (prints magenta) and printError method (prints red) to distinguish messages. 
Implemented our own pprint method to use click.echo w/ json.dumps.
Added a few try/catch blocks to grab some errors found during testing.
Added two spaces to fix column headers in status command.

### DIFF
--- a/receptorctl/receptorctl/cli.py
+++ b/receptorctl/receptorctl/cli.py
@@ -7,7 +7,6 @@ import tty
 import termios
 import click
 import json
-from pprint import pprint
 from functools import partial
 import dateutil.parser
 import pkg_resources
@@ -27,6 +26,17 @@ class IgnoreRequiredWithHelp(click.Group):
             for param in self.params:
                 param.required = False
             return super(IgnoreRequiredWithHelp, self).parse_args(ctx, args)
+
+def pprint(json_data):
+    click.echo(json.dumps(json_data, indent=4, sort_keys=True))
+
+def printWarning(message):
+    click.echo(click.style(f"Warning: {message}", fg='magenta'), err=True)
+
+def printError(message, exitCode=None):
+    click.echo(click.style(f"ERROR: {message}", fg='red'), err=True)
+    if exitCode:
+        sys.exit(exitCode)
 
 
 @click.group(cls=IgnoreRequiredWithHelp)
@@ -93,7 +103,7 @@ def status(ctx, printjson):
     rc = get_rc(ctx)
     status = rc.simple_command("status")
     if printjson:
-        print(json.dumps(status))
+        pprint(status)
         return
     node_id = status.pop("NodeID")
     print(f"Node ID: {node_id}")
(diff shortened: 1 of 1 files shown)


receptorctl/receptorctl/cli.py
```diff
@@ -28,6 +27,17 @@ def parse_args(self, ctx, args):
                 param.required = False
             return super(IgnoreRequiredWithHelp, self).parse_args(ctx, args)
 
+def pprint(json_data):
+    click.echo(json.dumps(json_data, indent=4, sort_keys=True))
+
+def printWarning(message):
```
shanemcd: Python function names should be `snake_case`. https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

receptorctl/receptorctl/cli.py
```diff
@@ -401,10 +409,10 @@ def cancel(ctx, unit_ids):
 @click.pass_context
 def release(ctx, force, unit_ids):
     if len(unit_ids) == 0:
-        print("No unit IDs supplied: Not doing anything")
+        printWarning("No unit IDs supplied: Not doing anything")
         return
     op = "release" if not force else "force-release"
-    print("Released:")
+    click.echo("Released:")
```
shanemcd: I like that you added helper functions for errors and warnings, but dislike that we `click.echo` repeated everywhere. I don't care what you call the new function (`print_info`?), but we should apply a similar pattern here to future-proof this code a bit.

receptorctl/receptorctl/cli.py
```diff
@@ -28,6 +27,17 @@ def parse_args(self, ctx, args):
...
 
+def pprint(json_data):
+    click.echo(json.dumps(json_data, indent=4, sort_keys=True))
+
+def printWarning(message):
+    click.echo(click.style(f"Warning: {message}", fg='magenta'), err=True)
+
+def printError(message, exitCode=None):
+    click.echo(click.style(f"ERROR: {message}", fg='red'), err=True)
+    if exitCode:
```
shanemcd: I don't like the idea of exiting the program as a side effect of a print function, it makes the code difficult to follow. I had to go through the rest of your diff and find when you were passing `1` as an (unnamed) argument to this function.

shanemcd (changes requested):

receptorctl/receptorctl/cli.py
john-westcott-iv: Fixed.

receptorctl/receptorctl/cli.py
john-westcott-iv: Initially I only added error and warning to make sure we got a consistent color and prefix. But added a generic print_message function as well.

receptorctl/receptorctl/cli.py
john-westcott-iv: I probably should have made a print_error_and_exit function to be more explicit but went back to just a print_error followed by sys.exit.

shanemcd:
The test failure here is strange. I don't see how what you did caused that, but we aren't seeing it on other runs. I don't have time to investigate today, curious if you can figure out what's going on.

shanemcd:
Got curious and tried this locally. I was able to reproduce, so should be easy to track down when you have time. Just make sure you have `tox` installed and run `make receptorctl-test`.

john-westcott-iv:
@shanemcd I found the issue. There were two print commands in the format `print("string", variable)`. I missed them when converting to print_message so they were not properly displaying their variables which were parsed in the tests.

shanemcd (approved):

PR #483, opened 2021-11-15
Cherrypick release 1.1 into devel

fosterseth:
this pulls in all go-routine leak fixes and the k8s createpod panic fix


shanemcd (approved):

AlanCoding:
:+1: this would be really good in order to keep things synced up

Issue #108, opened 2020-08-24
go test exec.Command('receptor') command not found on PATH


fosterseth:
e.g. https://github.com/project-receptor/receptor/blob/615c1eaa01054ee8f0304a0675063ced9a49d9da/tests/functional/cli/cli_test.go#L52

Maybe we can use relative paths, or add project folder to PATH automatically

fosterseth:
basically, running tests requires "receptor" in your PATH. I don't think this is big deal and hasn't caused problems. closing.

PR #486, opened 2021-11-16
Add make targets recetprotctl_install, test_skip_kube and modify test

john-westcott-iv:
recetprotctl_install - builds and installs receptorctl into your python environment in one command
test_skuo_kube - sets SKIP_KUBE env var = 1 before running test
test - now add `pwd` to your path (to pickup a built but not installed receptor binary) and tells you what binary its choosing for the test:
```
make test
Testing with receptor binary from: /home/jowestco/GIT/github/john-westcott-iv/receptor/receptor
?   	github.com/ansible/receptor/cmd/receptor-cb	[no test files]
?   	github.com/ansible/receptor/cmd/receptor-cl	[no test files]
...
```

Addresses #108

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ RELEASE := 1
 OFFICIAL := yes
 APPVER := $(VERSION)
 endif
+SKIP_KUBE=0
 
 # Container command can be docker or podman
 CONTAINERCMD ?= podman
@@ -84,8 +85,14 @@ else
 TESTCMD = -run $(RUNTEST)
 endif
 
+test_skip_kube: SKIP_KUBE=1
+test_skip_kube: test
+
 test:
-	@go test ./... -p 1 -parallel=16 $(TESTCMD) -count=1
+	@export PATH=$(PATH):$(PWD); \
+		export SKIP_KUBE=$(SKIP_KUBE); \
+		echo "Testing with receptor binary from: `which receptor 2>/dev/null`"; \
+		go test ./... -p 1 -parallel=16 $(TESTCMD) -count=1
 
 testloop: receptor
 	@i=1; while echo "------ $$i" && \
@@ -108,6 +115,11 @@ $(RECEPTORCTL_WHEEL): receptorctl/README.md receptorctl/setup.py $(shell find re
 	@cd receptorctl && python3 setup.py bdist_wheel
 
 receptorctl_wheel: $(RECEPTORCTL_WHEEL)
+
+# Build and install receptorctl into your python environment
+receptorctl_install: $(RECEPTORCTL_WHEEL)
+	pip uninstall -y receptorctl
+	pip install $(RECEPTORCTL_WHEEL)
 
 RECEPTORCTL_SDIST = receptorctl/dist/receptorctl-$(VERSION).tar.gz
 $(RECEPTORCTL_SDIST): receptorctl/README.md receptorctl/setup.py $(shell find receptorctl/receptorctl -type f -name '*.py')


fosterseth (approved):

Makefile
```diff
@@ -12,6 +12,7 @@ RELEASE := 1
 OFFICIAL := yes
 APPVER := $(VERSION)
 endif
+SKIP_KUBE=0
```
shanemcd, line 15: The way this is written breaks the ability to override this variable via an environment variable. Let's use this stripped-down Makefile as an example:

```make
SKIP_KUBE=0

test:
	@echo $(SKIP_KUBE)
```

The output is:

```
$ SKIP_KUBE=1 make test
0
```

If written it like this:


```make
SKIP_KUBE ?= 0

test:
	@echo $(SKIP_KUBE)
```

It works as expected:

```
$ SKIP_KUBE=1 make test
1
```

Makefile
```diff
@@ -84,8 +85,14 @@ else
 TESTCMD = -run $(RUNTEST)
 endif
 
+test_skip_kube: SKIP_KUBE=1
```
shanemcd, line 88: One of the primary design aspects of Make is to augment its behavior by overriding variables. If we were to follow this pattern you've started here, things would get unwieldy rather quickly. This new target is unnecessary as you could achieve the same thing by running:

```
SKIP_KUBE=1 make test
```

Makefile
```diff
@@ -84,8 +85,14 @@ else
...
+test_skip_kube: test
+
 test:
-	@go test ./... -p 1 -parallel=16 $(TESTCMD) -count=1
+	@export PATH=$(PATH):$(PWD); \
+		export SKIP_KUBE=$(SKIP_KUBE); \
```
shanemcd, line 93: This is unnecessary if you set the variable like I show in my comment above.

Another issue here is the indentation. There should be only 1 tab character here, followed by the number of spaces you would like to indent the code.

Makefile
```diff
@@ -84,8 +85,14 @@ else
...
+		echo "Testing with receptor binary from: `which receptor 2>/dev/null`"; \
```
shanemcd, line 94: Not all environments will have `which`, and this hides the error so it will be hard to tell what's going on without coming to look at this code.

Makefile
```diff
@@ -109,6 +116,11 @@ $(RECEPTORCTL_WHEEL): receptorctl/README.md receptorctl/setup.py $(shell find re
 
 receptorctl_wheel: $(RECEPTORCTL_WHEEL)
 
+# Build and install receptorctl into your python environment
+receptorctl_install: $(RECEPTORCTL_WHEEL)
```
shanemcd, line 120: I question the general usefulness of these types of targets. Makefiles are for building software and executing commands. I would consider it an anti-pattern to have targets that alter the user's environment. I would like to see this removed.

Makefile
```diff
@@ -84,8 +85,14 @@ else
 TESTCMD = -run $(RUNTEST)
 endif
 
+test_skip_kube: SKIP_KUBE=1
+test_skip_kube: test
```
shanemcd, line 89: These targets are unnecessary so they should be removed, but in the future, stuff like this should be added under `.PHONY`.

Makefile
```diff
@@ -84,8 +85,14 @@ else
...
+
 test:
-	@go test ./... -p 1 -parallel=16 $(TESTCMD) -count=1
+	@export PATH=$(PATH):$(PWD); \
```
shanemcd, line 92: Wouldn't we want to invert the precedence here to always prefer the build artifact?

shanemcd (commented):
I really wish we didn't merge most of this code. I would like to see everything reverted except for lines 92 and 94, but both of those lines also have issues that I've noted in-line.

shanemcd:
The commit message also has a typo... 😓 

> Add make targets recetprotctl_install, test_skip_kube and modify test

Going forward we should _slow down_ and make sure there is adequate time for feedback and discussion. I don't see anything in here that would be considered urgent.

PR #487, opened 2021-11-16
Make tests more stable by extending context timeout

fosterseth:
- some contexts were too short (i.e. 10 seconds), not giving enough time for receptor to re-establish a proper state after shutting down and starting up a node.
- Failed tests now print mesh directory, which will help debugging in the future.


tests/functional/lib/mesh/climesh.go
```diff
@@ -654,7 +654,7 @@ func (m *CLIMesh) CheckControlSockets() bool {
 
 // WaitForReady Waits for connections and routes to converge.
 func (m *CLIMesh) WaitForReady(ctx context.Context) error {
-	sleepInterval := 100 * time.Millisecond
+	sleepInterval := 500 * time.Millisecond
```
fosterseth, line 657: 100ms seemed aggressive and unnecessary

amolgautam25 (approved):

chrismeyersfsu (approved):

PR #490, opened 2021-11-17
Enforce Python code standards with black and flake8

shanemcd:
Code standards are important, but I hate spending any amount of time actually thinking about it. So, let's automate that activity out of our lives.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -6,8 +6,8 @@ on:
   push:
 
 jobs:
-  lint:
-    name: lint
+  lint-receptor:
+    name: lint-receptor
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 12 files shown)


thenets:
@shanemcd I believe that we need to talk about this before enforcing a code style. The `black` library is not even configured to follow the `tox` asserts.

For Go it's easy because the `go fmt` does the job. For Python, not so much. Libraries like `black` and `tox` do not follow the same defaults.

I decided to disable at least the `receptorctl-lint` until we fix these issues and decide on one specific code style here:
- https://github.com/ansible/receptor/pull/468

shanemcd:
@thenets We already use black in AWX, and we had a team discussion around its use when it was adopted. I'm not sure what you mean by "not even configured to follow the tox asserts."

thenets:
Already solved while talking to @shanemcd, but to give context to other people:

My mistake about the libraries' names. The issue was about `black` and `flake8`.
The current problem is about the line length. By default, `black` doesn't follow the same size as `flake8`, but even passing this size as a param, the comments don't have any formating process, which will fail during the `lint` step.

The current solution needs some manual changes in some cases. We can search later about a better way to run `black` automatically.

PR #517, opened 2022-01-11
Use shlex instead of strings.split for command

ghjm:
Currently, `--command-service` uses `strings.Split()` to split the provided command, and then executes the first component with the remaining components as parameters.  This means that parameters cannot be grouped using quotes.  For example, you cannot do things like:

```
- command-service:
        service: test
        command: sh -c "/my/noisy/script.sh 2> /dev/null"
```

This PR switches to using `shlex.Split()`, which splits the string into tokens using shell semantics, respecting quotes.

### DIFF
--- a/pkg/services/command.go
+++ b/pkg/services/command.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"net"
 	"os/exec"
-	"strings"
 
 	"github.com/ansible/receptor/pkg/logger"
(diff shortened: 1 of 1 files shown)


rebeccahhh (approved):
I like this, especially because it has error handling

ghjm:
@shanemcd @fosterseth merge me please

fosterseth (approved):

PR #522, opened 2022-01-18
produce error if encounter ImagePullBackoff

### LINKED ISSUE #521: Fail kubernetes work units if image pull error
In AWX i'm experimenting with longer image pending timeout times, because I want to be able to use resource requests and let the jobs get scheduled as resources are available.

But that means that I want Pending jobs to really be Pending.

We have a test case where we try and start a job with a image we know is not pullable because it requires a pull secret we don't configure correctly.

The expecation is that this job fails. Currently, it only does this if we exceed the pod pending timeout, but it should fail sooner.

Example:
This work unit has the status Pending:

```
'wyclp8oe': {'Detail': 'Pod created',                                                                                                                                                                             
              'ExtraData': {'Command': '',                                                                                                                                                                         
                            'Image': '',                                                                                                                                                                           
                            'KubeConfig': '',                                                                                                                                                                      
                            'KubeNamespace': 'dmesseahl2',
                            'KubePod': '',
                            'Params': '',
                            'PodName': 'automation-job-1542-6f6sp'},
              'State': 0,
              'StateName': 'Pending',
              'StdoutSize': 0,
              'WorkType': 'kubernetes-runtime-auth'},
```

But the status in kubernetes is:
```
NAME                        READY   STATUS             RESTARTS   AGE
automation-job-1542-6f6sp   0/1     ImagePullBackOff   0          25m
```
This type of error generally does not resolve itself, and we should move the work unit status to Error with the detail of "Image pull backoff"

kdelee:
fixes: https://github.com/ansible/receptor/issues/521

This correctly has the work unit fail if the container has ImagePullBackOff error
```
{'DVxy89tF': {'Detail': 'Error creating pod: container failed to start, ImagePullBackOff',
              'ExtraData': {'Command': '',
                            'Image': '',
                            'KubeConfig': '',
                            'KubeNamespace': 'controller-integration-1641911565',
                            'KubePod': '',
                            'Params': '',
                            'PodName': 'kube-6tw5v'},
              'State': 3,
              'StateName': 'Failed',
              'StdoutSize': 0,
              'WorkType': 'kube'}}
```

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -65,31 +65,51 @@ type kubeExtraData struct {
 // ErrPodCompleted is returned when pod has already completed before we could attach.
 var ErrPodCompleted = fmt.Errorf("pod ran to completion")
 
-// podRunningAndReady is a completion criterion for pod ready to be attached to.
-func podRunningAndReady(event watch.Event) (bool, error) {
-	if event.Type == watch.Deleted {
-		return false, errors.NewNotFound(schema.GroupResource{Resource: "pods"}, "")
-	}
+// ErrImagePullBackOff is returned when the image for the container in the Pod cannot be pulled.
+var ErrImagePullBackOff = fmt.Errorf("container failed to start")
 
-	if t, ok := event.Object.(*corev1.Pod); ok {
-		switch t.Status.Phase {
-		case corev1.PodFailed, corev1.PodSucceeded:
-			return false, ErrPodCompleted
-		case corev1.PodRunning:
-			conditions := t.Status.Conditions
-			if conditions == nil {
-				return false, nil
-			}
-			for i := range conditions {
-				if conditions[i].Type == corev1.PodReady &&
-					conditions[i].Status == corev1.ConditionTrue {
-					return true, nil
+
+// podRunningAndReady is a completion criterion for pod ready to be attached to.
+func podRunningAndReady() func(event watch.Event) (bool, error){
+	imagePullBackOffRetries := 3
+	inner := func(event watch.Event) (bool, error) {
+		if event.Type == watch.Deleted {
+			return false, errors.NewNotFound(schema.GroupResource{Resource: "pods"}, "")
+		}
+		if t, ok := event.Object.(*corev1.Pod); ok {
+			switch t.Status.Phase {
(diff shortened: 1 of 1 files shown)


fosterseth (approved):

fosterseth:
@shanemcd I'd like to get  your thoughts on this one before merging. Basically the idea is that if the pod cannot enter "running" state due to ImagePullError (incorrect image name in podspec, incorrect credentials are supplied, registry is down, etc), then fail the work unit. The system will retry the image pull a number of times, currently set by `imagePullBackOffRetries := 3` in the code.

This feature is similar to the pod_pending_timeout option we currently have in receptor. However, there is a good use case where users will want to set pod_pending_timeout to something much longer, say 1 hour, but still duck out early if images cannot be pulled. 

For example, k8s might not be starting pods due to resources not being available (if cpu/memory requests are set in the podspec), and thus pods will stay in pending until resources are freed up. In that situation, users will want to bump the pod_pending_timeout from 5 minutes to something much longer, but having to wait 1 hour to know that an image cannot be pulled because you typed the image name incorrectly is a bad experience.

So this feature targets the ImagePullError case specifically, whereas the pod_pending_timeout is a general catch-all for any situation where the pod cannot be started.

pkg/workceptor/kubernetes.go
```diff
@@ -65,31 +65,51 @@ type kubeExtraData struct {
...
+									}
+									imagePullBackOffRetries--
+								}
+						}
+					}
 				}
 			}
 		}
+
+		return false, nil
```
shanemcd, line 109: Under which conditions would we hit this? Should the error really be nil if the pod can't start?

fosterseth: PodRunningandReady is being passed into a watcher function,

`ev, err := watch2.UntilWithSync(ctxPodReady, lw, &corev1.Pod{}, nil, podRunningAndReady())`

This will call PodRunningAndReady in a loop, one for each "event" it pulls from the k8s api. It only terminates with PodRunningAndReady returns True, or returns an error

so returning `false, nil` means, "no the pod isn't ready, but there wasn't an error, so keep polling events from the k8s api"

shanemcd (approved):

PR #524, opened 2022-01-24
[DNM] receptor branch testing

AlanCoding:
Obviously do not merge this. I'm only here to get a receptor binary.


AlanCoding:
Looking for where this goes

https://github.com/ansible/receptor/blob/94672cdce4b61fd7ff8f35e065b5da25aafa45e2/.github/workflows/build_binary_from_ref.yml#L47-L48

PR #526, opened 2022-01-27
Support pinned certificates

ghjm:
Receptor currently allows a user to specify a CA, or list of CAs, that must sign a certificate in order to trust it for a connection.  However, there are situations where you want to be able to identify a single certificate, or one of a list of certificates, that should be allowed to connect, and it's not sufficient to just say "any certificate signed by such-and-so CA."  The solution to this is certificate pinning, where you list specific certificate fingerprints that should be the only ones to be trusted.

Here are example configuration files exercising this feature:

```
---
- node:
    id: foo

- log-level: debug

- tls-server:
    name: foo_tls
    cert: certs/testcert-foo.crt
    key: certs/testcert-foo.key
    requireclientcert: true
    clientcas: certs/ca.crt
    pinnedclientcert: E6:9B:98:A7:A5:DB:17:D6:E4:2C:DE:76:45:42:A8:79:A3:0A:C5:6D:10:42:7A:6A:C4:54:57:83:F1:0F:E
2:95

- tcp-listener:
    port: 2222
    tls: foo_tls

- control-service:
    service: control
    filename: receptor.sock
```

```
---
- node:
    id: bar

- log-level: debug

- tls-client:
    name: bar_tls
    cert: certs/testcert-bar.crt
    key: certs/testcert-bar.key
    rootcas: certs/ca.crt
    pinnedservercert: 91:4B:90:C4:D2:33:94:62:37:2F:0C:83:7A:20:C1:E3:61:FB:0D:EE:9F:6B:E7:1C:2B:76:A2:97:63:21:4
E:F4

- control-service:
    service: control
    filename: bar.sock

- tcp-peer:
    address: localhost:2222
    redial: true
    tls: bar_tls

```

You can find the fingerprint of a given certificate using `openssl x509 -in certs/testcert-foo.crt -noout -fingerprint -sha256`.  This PR supports sha256 and sha512, but not sha1, because sha1 is now considered insecure.  If sha1 support is desired it can easily be added.

### DIFF
--- a/pkg/netceptor/conn.go
+++ b/pkg/netceptor/conn.go
@@ -152,7 +152,6 @@ func (li *Listener) sendResult(conn net.Conn, err error) {
 		err:  err,
 	}:
 	case <-li.doneChan:
-	case <-li.doneChan:
 	}
 }
 
--- a/pkg/netceptor/tlsconfig.go
+++ b/pkg/netceptor/tlsconfig.go
@@ -4,10 +4,15 @@
 package netceptor
 
 import (
+	"bytes"
+	"crypto/sha256"
+	"crypto/sha512"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/hex"
 	"fmt"
 	"io/ioutil"
+	"strings"
 
 	"github.com/ghjm/cmdline"
 )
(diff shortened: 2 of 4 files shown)


pkg/netceptor/tlsconfig.go
```diff
@@ -53,15 +59,22 @@ func (cfg tlsServerCfg) Prepare() error {
...
 			return fmt.Errorf("error reading client CAs file: %s", err)
 		}
 		clientCAs := x509.NewCertPool()
-		clientCAs.AppendCertsFromPEM(bytes)
+		clientCAs.AppendCertsFromPEM(caBytes)
 		tlscfg.ClientCAs = clientCAs
 	}
 
+	if len(cfg.PinnedClientCert) > 0 {
+		tlscfg.VerifyPeerCertificate, err = GetPeerCertificatePinVerifier(cfg.PinnedClientCert)
```
fosterseth, line 72: for the quic connections, will this conflict with https://github.com/ansible/receptor/blob/2fb942e162a89be0a8793eb80691d0fa10249d08/pkg/netceptor/conn.go#L74

ghjm: You're right, this will indeed conflict.

ghjm:
@fosterseth I fixed the conflict by having the netceptor verify function wrap the pinning verify function, if it exists.  Let me know what you think of this.  I guess the other solution would be to add a map to the netceptor struct that stores per-config pin fingerprints and have the netceptor-internal verify function do the pinning checks, but that seems like a lot of work.

ghjm:
@fosterseth @eqrx @shanemcd Can I get a :heavy_check_mark: on this?

fosterseth:
@ghjm any chance we could get you to put a small paragraph / snippet describing this feature in the docs, maybe after this Generate Certs paragraph here

https://github.com/ansible/receptor/blob/1dc72e019c12eddc41ac92ca8fb60a99fdb9828e/docs/source/tls.rst?plain=1#L61

maybe include that openssl command to get the fingerprint too

that would be very valuable for us

ghjm:
@fosterseth Docs added as requested

fosterseth:
pulled this down and tested, works well!

Users of course will need to know that if certs are ever refreshed, then this fingerprint will need to be manually updated in the config files. I think that is okay, we don't really have an automatic way to manage certs within receptor mesh as of right now anyways.

@shanemcd I'm good with this PR if you are

pkg/netceptor/conn.go
```diff
@@ -152,7 +152,6 @@ func (li *Listener) sendResult(conn net.Conn, err error) {
 		err:  err,
 	}:
 	case <-li.doneChan:
-	case <-li.doneChan:
```
shanemcd, line 155: What was going on here? Nice find.

shanemcd (approved):
Thanks @ghjm, hope you are doing well!

PR #529, opened 2022-01-31
adding workflow_dispatch trigger to github CI

sarabrajsingh:
this small change will allow us to re-run jobs in github actions

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -4,6 +4,7 @@ name: CI
 on:
   pull_request:
   push:
+  workflow_dispatch:
 
 jobs:
   lint-receptor:


shanemcd:
There's already a way to do this:

receptor@ca1bc69 2022-02-04 10-12-28

shanemcd:
I also think the way you have set this up would always clone devel.

sarabrajsingh:
> I also think the way you have set this up would always clone devel.

@shanemcd looks like the problems are occuring because i'm working on an out-of-date version of my fork. I'll try re-baseing the fork on my end to see if it helps

sarabrajsingh:
> I also think the way you have set this up would always clone devel.

@shanemcd looks like the problems are occuring because i'm working on an out-of-date version of my fork. I'll try re-baseing the fork on my end to see if it helps

PR #530, opened 2022-01-31
Move ping and traceroute to Netceptor package

ghjm:
Currently, the ping and traceroute functionality is locked inside the control service implementation.  This PR moves the functionality to the Netceptor module, with the control service handlers calling it.  The motivation is to allow the use of ping and traceroute from Go code that imports Netceptor.


fosterseth (approved):

ghjm:
@fosterseth I can't actually merge this because I don't have write access.  Can you do the necessary?  :slightly_smiling_face:

PR #532, opened 2022-02-01
Fix receptorctl status output

fosterseth:
before

```
bar          {
                "fish": 1,
                "foo": 1
}
fish         {
                "bar": 1
}
foo          {
                "bar": 1
}
```

after

```
bar          fish: 1 foo: 1 
fish         bar: 1 
foo          bar: 1 
```

### DIFF
--- a/receptorctl/receptorctl/cli.py
+++ b/receptorctl/receptorctl/cli.py
@@ -149,7 +149,9 @@ def status(ctx, printjson):
         print_message(f"{'Known Node':<{longest_node}} Known Connections")
         for node in costs:
             print_message(f"{node:<{longest_node}} ", nl=False)
-            print_json(costs[node])
+            for peer, cost in costs[node].items():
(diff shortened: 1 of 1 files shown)


shanemcd (approved):

AlanCoding:
Won't this have an impact on AWX? I don't think we use the `--json` flag (we could), but I don't see how this is conditional on requesting JSON.

fosterseth:
@AlanCoding 

there are 2 parts to receptorctl, `ReceptorSocket` which is the class that handles all socket IO, and the click library which is just the human-friendly command library which uses ReceptorSocket. 

AWX interacts with receptor via the ReceptorSocket only, not via the click library.

in receptor.py
```python3
from receptorctl.socket_interface import ReceptorControl
```

This fix just affects click library usage, so it shouldn't have any impact on awx.

PR #533, opened 2022-02-01
fix error message to show what node id was expected vs found in cert

fosterseth:
Node IDs are embedded in the TLS certificate. If a client attempts to connect with a cert with an invalid name (the node ID in the cert doesn't match the node ID of the connection), then the server properly emits an error.

Currently the error message is bugged, so this commit fixes that

before

`ERROR 2022/02/01 12:47:10 RVF ReceptorNameError: **%!s(<nil>)`

after

`ERROR 2022/02/01 12:54:15 RVF ReceptorNameError: expected foo but found baz`

### DIFF
--- a/pkg/netceptor/netceptor.go
+++ b/pkg/netceptor/netceptor.go
@@ -990,7 +990,7 @@ func (s *Netceptor) receptorVerifyFunc(tlscfg *tls.Config, expectedNodeID string
 			}
 		}
 		if !found {
-			logger.Error("RVF ReceptorNameError: %s", err)
+			logger.Error("RVF ReceptorNameError: expected %s but found %s", expectedNodeID, strings.Join(receptorNames, ", "))
 
 			return ReceptorCertNameError{ValidNodes: receptorNames, ExpectedNode: expectedNodeID}
 		}


pkg/workceptor/kubernetes.go
```diff
@@ -68,9 +68,8 @@ var ErrPodCompleted = fmt.Errorf("pod ran to completion")
 // ErrImagePullBackOff is returned when the image for the container in the Pod cannot be pulled.
 var ErrImagePullBackOff = fmt.Errorf("container failed to start")
 
-
 // podRunningAndReady is a completion criterion for pod ready to be attached to.
-func podRunningAndReady() func(event watch.Event) (bool, error){
+func podRunningAndReady() func(event watch.Event) (bool, error) {
```
fosterseth: these are all changes from linter

shanemcd: It would be best to have stuff like this live in a separate commit, to make it easier to figure out what's going on when looking at blames in the future.

shanemcd (approved):

sarabrajsingh (approved):

PR #536, opened 2022-02-03
Make node ID verifier factory public

ghjm:
Someone writing Go code importing Netceptor may want to establish an above-the-mesh mutual TLS connection.  To do this, they need to build a tls.Config that verifies node IDs.  Because `receptorVerifyFunc` is currently private, the user will have to write their own verifier that understands the Receptor otherName OID - or more likely, just copy-and-paste this code.  Making it public allows a library user to easily create verifier functions as needed.


ghjm:
Looks like you don't need this if you use `ClientConf` and `ServerConf`.  Maybe I wasn't having my very best day when I thought up this API.

PR #545, opened 2022-02-14
Fix hanging channel writes

shanemcd:
Tests were sometimes getting stuck waiting for backends to shutdown and I noticed that this goroutine was stuck trying to write to these channels.

Not sure if this is the correct way to fix this, but `make testloop` seems happy.

### DIFF
--- a/pkg/netceptor/netceptor.go
+++ b/pkg/netceptor/netceptor.go
@@ -1165,7 +1165,12 @@ func (s *Netceptor) forwardMessage(md *MessageData) error {
 	// decrement HopsToLive
 	message[1]--
 	logger.Trace("    Forwarding data length %d via %s\n", len(md.Data), nextHop)
-	c.WriteChan <- message
+
+	select {
+	case c.WriteChan <- message:
+	case <-time.After(1 * time.Second):
+		logger.Error("Unable to send message to WriteChan for %s.", nextHop)
+	}
 
 	return nil
 }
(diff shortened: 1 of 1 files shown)


pkg/netceptor/netceptor.go
```diff
@@ -1165,7 +1165,12 @@ func (s *Netceptor) forwardMessage(md *MessageData) error {
 	// decrement HopsToLive
 	message[1]--
 	logger.Trace("    Forwarding data length %d via %s\n", len(md.Data), nextHop)
-	c.WriteChan <- message
+
+	select {
+	case c.WriteChan <- message:
+	case <-time.After(1 * time.Second):
```
shanemcd, line 1171: @ghjm Mind giving me your thoughts on this? I found it when debugging this flaky test: 

https://github.com/ansible/receptor/blob/3000b6175bc31851017cd6004ed80bb1485ccd49/tests/functional/mesh/work_test.go#L327-L331

Notice how we shut down node2 and then immediately try to submit work.

We get hung up here on this write to `WriteChan` via: https://github.com/ansible/receptor/blob/3000b6175bc31851017cd6004ed80bb1485ccd49/pkg/netceptor/conn.go#L324

I was unable to use a context here like I did in the blob below because the node unexpected goes away. I thought maybe we could use a write deadline, but noticed this:

https://github.com/ansible/receptor/blob/3000b6175bc31851017cd6004ed80bb1485ccd49/pkg/netceptor/packetconn.go#L234

shanemcd: See my latest commit. The context didn't seem to be enough there either.

shanemcd:
Closing in favor of https://github.com/ansible/receptor/pull/547

ghjm:
I know this is closed now  but FWIW, I would have been concerned about the one second time limit introducing unexpected behaviors under heavy load.  What if everything's working fine but we're on an overloaded VM host that just didn't happen to schedule us for a whole second?

PR #547, opened 2022-02-15
Fix several race conditions and sources of test flake

fosterseth:
Fixed up a few places in the code where we do blocking send and receive calls, but don't select on the overall context being finished.

This should help mitigate some race conditions we've discovered in testing when starting up and shutting down nodes.

### DIFF
--- a/pkg/netceptor/netceptor.go
+++ b/pkg/netceptor/netceptor.go
@@ -106,6 +106,7 @@ type Netceptor struct {
 	maxForwardingHops      byte
 	maxConnectionIdleTime  time.Duration
 	workCommands           []WorkCommand
+	workCommandsLock       *sync.RWMutex
 	epoch                  uint64
 	sequence               uint64
 	connLock               *sync.RWMutex
@@ -195,6 +196,7 @@ type connInfo struct {
 	CancelFunc       context.CancelFunc
 	Cost             float64
 	lastReceivedData time.Time
+	lastReceivedLock *sync.RWMutex
 }
 
 type nodeInfo struct {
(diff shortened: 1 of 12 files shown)


pkg/netceptor/netceptor.go
```diff
@@ -852,16 +852,28 @@ func (s *Netceptor) SubscribeRoutingUpdates() chan map[string]string {
...
+		go func() {
+			defer wg.Done()
+			select {
+			case writeChans[i] <- message:
+			case <-contexts[i].Done():
+				logger.Debug("connInfo cancelled during flood write")
+			}
+		}()
 	}
+	wg.Wait()
```
fosterseth: this wait makes me a little nervous because it blocks the entire flood call until all of the writes are done, but at the same time not waiting on these go-routines also makes me nervous (i.e. go-routine leaks).

@ghjm any opinions on this? I could be convinced to just drop the wg.Wait() here

ghjm: I don't see the need for the WaitGroup at all.  We're not holding any locks and there's no cleanup code at the end.  I don't immediately recall any cases where a caller of flood() would depend on the messages being sent before the function returns, though it might be worthwhile to look at the places it's called and docker check this.

Also, it would be cleaner to just launch the goroutine inside the first loop, which would eliminate the need for the temporary data structures and the second loop.

fosterseth: @ghjm okay I'm going to remove the waitgroup. I didn't have any evidence of this being a problem before, so no need to introduce a waitgroup here. thanks

shanemcd:
It looks like this is leaking a whoooole lot of goroutines. Going to clone this locally and see if I can find what's going on.

shanemcd (approved):

Issue #489, opened 2021-11-17
k8s work unit fails because container exited before attaching stdin stream


fosterseth:
This bug occurs occasionally while running make test.

Here is the work unit status from the following functional test `TestWork/kube_worker/work_submit_while_remote_node_is_down`

```
{
  "State": 3,
  "Detail": "Stream error running pod: stdin: Internal error occurred: error attaching to container: container is in CONTAINER_EXITED state, stdout: http2: response body closed",
  "StdoutSize": 10,
  "WorkType": "echosleepshort",
  "ExtraData": {
    "Image": "alpine",
    "Command": "sh -c \"for i in `seq 1 5`; do echo $i;done\"",
    "Params": ""
  }
  etc.
}
```

Interestingly the pod ran and we got back the full stdout (10 bytes). This work unit doesn't even have stdin.

Seems the pod is running and finishing before we can attach a stdin stream, but this stream isn't needed in this case. This work unit should have succeeded.

fosterseth:
add this to your receptor node conf

```yaml
- work-kubernetes:
    worktype: kubeit2
    authmethod: runtime
    allowruntimeauth: true
    allowruntimepod: true
    allowruntimeparams: true
```

create pod.yml
```yaml
apiVersion: v1
kind: Pod
metadata:
    generateName: myapp-pod-
    labels:
        app: myapp
spec:
    containers:
    - name: worker
      image: busybox
      command: ['sh', '-c', 'echo The Pod is running && sleep 0 && exit 0']
      # resources:
      #     limits:
      #         memory: "1000000Mi"
    restartPolicy: Never
```

`minikube start`, start receptor process, then submit work

`receptorctlfoo work submit kubeit2 --param secret_kube_pod=@/home/sbf/sockceptor/pod.yml --param secret_kube_config=@/home/sbf/.kube/config --no-payload`

thenets:
@fosterseth @shanemcd @AlanCoding is this still a problem?

fosterseth:
@thenets this should no longer be an issue, fixed in https://github.com/ansible/receptor/pull/547

PR #549, opened 2022-02-15
k8s skip stdin if stdin file is 0 bytes

fosterseth:
related https://github.com/ansible/receptor/issues/489

Check if stdin file is 0 bytes. If empty, do not attempt to attach stdin stream to container. If the container already exited, this can lead to an unintended error.

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -383,6 +383,8 @@ func (kw *kubeUnit) runWorkUsingLogger() {
 				}
 			}
 		}()
+	} else {
+		kw.UpdateBasicStatus(WorkStateRunning, "Pod Running", stdout.Size())
 	}
 
 	// Actually run the streams.  This blocks until the pod finishes.
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -339,7 +339,9 @@ func (kw *kubeUnit) runWorkUsingLogger() {
 	var stdin *stdinReader
 	if !skipStdin {
 		stdin, err = newStdinReader(kw.UnitDir())
-		if err != nil {
+		if err == errFileSizeZero {
+			skipStdin = true
+		} else if err != nil {
 			errMsg := fmt.Sprintf("Error opening stdin file: %s", err)
 			logger.Error(errMsg)
 			kw.UpdateBasicStatus(WorkStateFailed, errMsg, 0)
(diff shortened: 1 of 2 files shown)


fosterseth:
to test this I simulated the error by adding a sleep(3) right before the code that attaches the stdin stream, to ensure the container is done at this point. Without patch, I got the error, although a slightly different message than in https://github.com/ansible/receptor/issues/489

```
    "Q1ojiukD": {
        "Detail": "Stream error running pod: unable to upgrade connection: container worker not found in pod kubeit2-ffp5x_default",
```

It may just be the case that the test is hitting the problem when the pod is in a particular state, that I am not exactly capturing here.

My patch makes these pods and work units pass as expected.

pkg/workceptor/kubernetes.go
```diff
@@ -68,9 +68,8 @@ var ErrPodCompleted = fmt.Errorf("pod ran to completion")
 // ErrImagePullBackOff is returned when the image for the container in the Pod cannot be pulled.
 var ErrImagePullBackOff = fmt.Errorf("container failed to start")
 
-
 // podRunningAndReady is a completion criterion for pod ready to be attached to.
-func podRunningAndReady() func(event watch.Event) (bool, error){
+func podRunningAndReady() func(event watch.Event) (bool, error) {
```
fosterseth: all of this is linter shenanigans

pkg/workceptor/stdio_utils.go
```diff
@@ -68,9 +68,19 @@ type stdinReader struct {
 	doneOnce sync.Once
 }
 
+var errFileSizeZero error
```
eqrx, line 71: That is nil. Use errors.New to create an error.

pkg/workceptor/kubernetes.go
```diff
@@ -339,7 +339,9 @@ func (kw *kubeUnit) runWorkUsingLogger() {
 	var stdin *stdinReader
 	if !skipStdin {
 		stdin, err = newStdinReader(kw.UnitDir())
-		if err != nil {
+		if err == errFileSizeZero {
```
eqrx, line 342: Use errors.Is to compare errors or create a custom error type

eqrx, line 342: Use errors.Is to compare errors or create a custom error type

pkg/workceptor/stdio_utils.go
eqrx, line 71: That is nil. Use errors.New to create an error.

eqrx (changes requested):

shanemcd:
I included this commit + the PR feedback in https://github.com/ansible/receptor/pull/547

PR #550, opened 2022-02-16
fix lint

fosterseth:


eqrx:
How was this even possible to merge? GH actions should have prevented that?

eqrx:
~Oh right, someone pushed directly, never mind~ Sorry, my bad.

shanemcd:
I'm also confused about this. I dont see where anyone pushed anything directly. I just looked at the blames and these were in PRs where the checks passed. Is something not working correctly?

eqrx:
https://github.com/ansible/receptor/commit/d8ad8d6da5998c0b4f8c58f89d35b52b9cae8996 Introduced autofixing in the linter. Has anyone checked it that just causes the linter to fix errors in the checked out version of GH actions (without committing them afterwards)?

shanemcd:
@eqrx That was it, nice find. Thank you.

shanemcd (approved):

.golangci.yml
```diff
@@ -69,7 +69,9 @@ linters:
     # - wrapcheck # TODO: Errors passed upwards should be wrapped.
 
 issues:
-  fix: true
+  # Dont commit the following line. 
+  # It will make CI pass without telling you about errors.
+  # fix: true
```
ghjm, line 74: THANK YOU!!!!!!!

PR #560, opened 2022-02-24
Update base image to stream9

shanemcd:
Also moved around / combined some layers to optimize image size.

### DIFF
--- a/packaging/container/Dockerfile
+++ b/packaging/container/Dockerfile
@@ -1,12 +1,13 @@
-FROM quay.io/centos/centos:stream8 as builder
+FROM quay.io/centos/centos:stream9 as builder
 
-RUN dnf -y update && dnf install -y golang make python3 python3-pip python3-wheel git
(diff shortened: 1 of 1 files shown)


packaging/container/Dockerfile
```diff
@@ -1,35 +1,36 @@
...
 COPY receptor_python_worker-${VERSION}-py3-none-any.whl /tmp
-RUN pip3 install /tmp/*.whl
-RUN rm /tmp/*.whl
-
 COPY receptor.conf /etc/receptor/receptor.conf
+
+RUN dnf -y update && \
+    dnf -y install python3-pip && \
+    dnf clean all && \
+    pip install --no-cache-dir wheel dumb-init && \
```
shanemcd, line 25: Note that I removed `tini` in favor of `dumb-init`.

- `epel-release` doesn't seem to exist in stream9 yet.
- dumb-init is available via pip
- dumb-init is the more active project
- dumb-init is already used by every Execution Environment.

PR #572, opened 2022-03-02
pytest: Custom marker for GitHub

### LINKED ISSUE #495: Add plugin `pytest-github` to the `receptorctl` test suite
This plugin allows us to skip tests with the correspondent open issue.
- https://github.com/jlaska/pytest-github

Fixes:
- https://github.com/ansible/receptor/issues/508

Akasurde:
Created custom marker for skipping test if GitHub issue/pull is open.

Fixes: https://github.com/ansible/receptor/issues/495

### DIFF
--- a/receptorctl/tests/conftest.py
+++ b/receptorctl/tests/conftest.py
@@ -7,6 +7,7 @@ import shutil
 import time
 import json
 import yaml
+import urllib.request
 from click.testing import CliRunner
 
 from lib import create_certificate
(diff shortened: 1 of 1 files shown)


receptorctl/tests/conftest.py
```diff
@@ -389,3 +390,23 @@ def f_invoke_as_json(command, args: list = []):
...
+        "markers",
+        "github(): this mark skips the tests for the given open GitHub issues",
+    )
+
+
+def pytest_runtest_setup(item):
+    gh_markers = [mark.args[0] for mark in item.iter_markers(name="github")]
+    if gh_markers:
+        issue_url = gh_markers[0]
+        with urllib.request.urlopen(issue_url) as f:
```
shanemcd, line 406: Does this require using the API endpoint for an Issue?

Akasurde: Yes. You will need to specify the marker like this - 

```python
import pytest

@pytest.mark.github('https://api.github.com/repos/ansible/receptor/issues/441')
def test_c():
    print("This is Test C")
```

I choose this routine in order to avoid additional dependency on third-party library.

Akasurde: we can simplify the logic further to specify only the issue number or pull request number. But for starters, I think this OK.

shanemcd: I think this is pretty difficult to use. Normally you'd just visit an issue in your browser. We should be able to copy / paste that URL instead of having to construct this API endpoint. We use https://github.com/jlaska/pytest-github in other projects, so we're not maintaining anything extra by also using it here.

AaronH88:
Thanks for the PR! 

As this has been open for a long time I am going to close it,
If you feel this PR can be reopend/redone to address the comments, feel free to open 😄 

Thanks again for contributing!

Issue #420, opened 2021-09-14
Verification of nodeID to cert before joining mesh


fosterseth:
- Currently nodes can "lie" about their nodeID (nodeID in config doesn't have to match what is in cert)

- Verification takes place as control service level, but not to join mesh

- This feature will ensure nodes joining mesh have a nodeID that matches what is in their cert

Acceptance Criteria:

- Try to set up a receptor node with nodeID "foo2" but your cert says you are "foo". Should not be able to join, and this node should not be on the routing table when issuing a "status" command

Time estimate: 1 sprint

alisondudiak:
might be related - https://github.com/ansible/receptor/pull/553

sarabrajsingh:
reproducer:

1. use `tower-packaging@devel` to install at least a 2-node cluster, compromising of one control and one execution node (normal path forward). this will install a cluster with the changes from #553.

#### scenario 1 - bad cert (wrong CN, wrong SAN) given to a node that was in the mesh

2. after installation, leverage the receptor self-signed CA located in `/etc/receptor/tls/ca/mesh-CA.crt` to create a new `csr`. the idea here is to generate a cert with a bad CN, bad SAN, and incorrect originating IPs. 
```bash
receptor --cert-makereq commonname=foobar bits=4096 nodeid=foobar outreq=/etc/receptor/tls/foobar.csr outkey=/etc/receptor/tls/foobar.key ipaddress=127.0.0.1 ipaddress=127.0.0.1
```
3. leverage receptors' self-signed CA to sign the CSR and generate a certificate for consumption
```bash
receptor --cert-signreq verify=yes cacert=/etc/receptor/tls/ca/mesh-CA.crt cakey=/etc/receptor/tls/ca/mesh-CA.key req=/etc/receptor/tls/foobar.csr outcert=/etc/receptor/tls/foobar.crt notafter="2031-12-28T00:03:51-05:00"
```
4. copy over `/etc/receptor/tls/foobar.crt` and `/etc/receptor/tls/foobar.key` to an execution node. note the IP address of the execution node (in my case, it's `192.168.56.104`)
5. modify the `/etc/receptor/receptor.conf` file on the execution node to reference the new fudged cert. namely, edit the `tls-server` and `tls-client` sections
```yaml
# TLS
- tls-server:
    name: tls_server
    cert: /etc/receptor/tls/foobar.crt
    key: /etc/receptor/tls/foobar.key
[...]

PR #578, opened 2022-03-15
use receptorNames (oid=otherNames) in tls bootstrap; fixed associated unit tests for receptor and receptorctl

sarabrajsingh:
upstream issue #420 

added otherName (SAN) checks to `tlsconfig.go` using `utils.ReceptorNames()`

@fosterseth @thenets @shanemcd please re-review!

### DIFF
--- a/pkg/netceptor/tlsconfig.go
+++ b/pkg/netceptor/tlsconfig.go
@@ -7,10 +7,12 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/hex"
+	"encoding/pem"
 	"fmt"
 	"io/ioutil"
 	"strings"
 
+	"github.com/ansible/receptor/pkg/utils"
 	"github.com/ghjm/cmdline"
 )
 
@@ -67,6 +69,33 @@ func (cfg tlsServerCfg) Prepare() error {
 	cert, err := tls.X509KeyPair(certbytes, keybytes)
 	if err != nil {
 		return err
+	}
+
+	block, _ := pem.Decode(certbytes)
+	if block == nil {
+		return fmt.Errorf("failed to parse certificate PEM")
+	}
+
+	parsedCert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return err
+	}
+
+	s, err := utils.ReceptorNames(parsedCert.Extensions)
+	if err != nil {
+		return err
+	}
+
+	found := false
+
+	for _, name := range s {
+		if MainInstance.nodeID == name {
+			found = true
+		}
+	}
+
+	if !found {
+		return fmt.Errorf("nodeID=%s not found in certificate name(s); cfg section=%s; server cert=%s", MainInstance.nodeID, cfg.Name, cfg.Cert)
 	}
 
 	tlscfg.Certificates = []tls.Certificate{cert}
--- a/pkg/netceptor/tlsconfig.go
+++ b/pkg/netceptor/tlsconfig.go
@@ -73,7 +73,7 @@ type tlsServerCfg struct {
 	RequireClientCert      bool     `required:"false" description:"Require client certificates" default:"false"`
 	ClientCAs              string   `required:"false" description:"Filename of CA bundle to verify client certs with"`
 	PinnedClientCert       []string `required:"false" description:"Pinned fingerprint of required client certificate"`
-	SkipReceptorNamesCheck bool     `required:"false" description:"if enabled, validate cert using ReceptorNames OID in Certificate"`
+	SkipReceptorNamesCheck bool     `required:"false" description:"Skip verifying ReceptorNames OIDs in certificate at startup" default:"false"`
 }
 
 // Prepare creates the tls.config and stores it in the global map.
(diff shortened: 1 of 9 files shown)


thenets (approved):
Thanks for improving the certificate generation for the `receptorctl` test suite.

shanemcd:
I just tried this and I'm not sure if I'm doing something wrong, but I'm still able to join the mesh after modifying my receptor config to include a mismatching node name. If you are around today and have time, let's jump on a call and look at this together.

shanemcd:
I just tried this and I'm not sure if I'm doing something wrong, but I'm still able to join the mesh after modifying my receptor config to include a mismatching node name. If you are around today and have time, let's jump on a call and look at this together.

pkg/utils/common.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"crypto/x509"
+
+	"github.com/ansible/receptor/pkg/logger"
+)
+
+func ParseReceptorNamesFromCert(cert *x509.Certificate, expectedHostname string) (bool, []string, error) {
+	var receptorNames []string
+	receptorNames, err := ReceptorNames(cert.Extensions)
```
fosterseth, line 11: if ReceptorNames is no longer being called outside of the utils package, then we can "unexport" it -- make it lower case

pkg/netceptor/tlsconfig.go
```diff
@@ -39,14 +41,39 @@ func decodeFingerprints(fingerprints []string) ([][]byte, error) {
...
-	RequireClientCert bool     `description:"Require client certificates" default:"false"`
-	ClientCAs         string   `description:"Filename of CA bundle to verify client certs with"`
-	PinnedClientCert  []string `description:"Pinned fingerprint of required client certificate"`
+	Name                   string   `required:"true" description:"Name of this TLS server configuration"`
+	Cert                   string   `required:"true" description:"Server certificate filename"`
+	Key                    string   `required:"true" description:"Server private key filename"`
+	RequireClientCert      bool     `required:"false" description:"Require client certificates" default:"false"`
+	ClientCAs              string   `required:"false" description:"Filename of CA bundle to verify client certs with"`
+	PinnedClientCert       []string `required:"false" description:"Pinned fingerprint of required client certificate"`
+	SkipReceptorNamesCheck bool     `required:"false" description:"if enabled, validate cert using ReceptorNames OID in Certificate"`
```
fosterseth: the description for this is tricky, I think we want to emphasize this boolean only controls the check at startup, not during the quic TLS handshake (where the receptor names will always be checked, regardless of how this boolean is set)

maybe something like,

`Skip verifying Receptor node ID in certificate at startup`

sarabrajsingh: done @fosterseth

shanemcd:
Let's get the "fix lint" and "fix tests" commits rebased out of the history here. Those should get squashed into the commits that introduced whatever problems you were addressing.

sarabrajsingh:
@shanemcd done

shanemcd:
Awesome, thank you. @fosterseth have you pulled this down and played around?

pkg/netceptor/tlsconfig.go
```diff
@@ -39,14 +41,39 @@ func decodeFingerprints(fingerprints []string) ([][]byte, error) {
...
+	if err != nil {
+		return err
+	}
+
+	found, receptorNames, err := utils.ParseReceptorNamesFromCert(parsedCert, n.nodeID)
+	if err != nil {
+		return err
+	}
+
+	if !found {
```
fosterseth, line 61: we still need to return without error here if len(receptorNames) == 0

if I create a cert without a nodeID, and use that on a config with node ID foo, but I see this error

```
[sbf@fedora sockceptor]$ ./receptor -c foo.yml
Error: MainInstance.nodeID=foo not found in certificate name(s); names found=[]; cfg section=server2; server cert=/home/sbf/sockceptor/certs/bob.crt
```

we want this to be permissible

fosterseth, line 61: we still need to return without error here if len(receptorNames) == 0

if I create a cert without a nodeID, and use that on a config with node ID foo, but I see this error

```
[sbf@fedora sockceptor]$ ./receptor -c foo.yml
Error: MainInstance.nodeID=foo not found in certificate name(s); names found=[]; cfg section=server2; server cert=/home/sbf/sockceptor/certs/bob.crt
```

we want this to be permissible

sarabrajsingh: @fosterseth - just thinking about this before i write a patch. would it be more optimal for the consumer to just flip the bit for the `SkipReceptorNamesCheck` flag to `true` if they give a `ReceptorName/OID=""` and skip the check altogether?

fosterseth: I'm okay with that option

pkg/netceptor/tlsconfig.go
```diff
@@ -106,12 +139,13 @@ func (cfg tlsServerCfg) Prepare() error {
...
-	RootCAs            string   `required:"false" description:"Root CA bundle to use instead of system trust"`
-	InsecureSkipVerify bool     `required:"false" description:"Accept any server cert" default:"false"`
-	PinnedServerCert   []string `required:"false" description:"Pinned fingerprint of required server certificate"`
+	Name                   string   `required:"true" description:"Name of this TLS client configuration"`
+	Cert                   string   `required:"false" description:"Client certificate filename"`
+	Key                    string   `required:"false" description:"Client private key filename"`
+	RootCAs                string   `required:"false" description:"Root CA bundle to use instead of system trust"`
+	InsecureSkipVerify     bool     `required:"false" description:"Accept any server cert" default:"false"`
+	PinnedServerCert       []string `required:"false" description:"Pinned fingerprint of required server certificate"`
+	SkipReceptorNamesCheck bool     `required:"false" description:"if enabled, skip verifying ReceptorNames OIDs in certificate at startup"`
```
fosterseth, line 148: golang will default this to false, but for consistency and clarify I think we can add `default:"false"` , similar to InsecureSkipVerify

To make the description a little shorter, we could omit "if enabled," and just start with "Skip.."

fosterseth (approved):

fosterseth:
Note: if users that have set up tls-server and tls-client to be used for TCP connections only, and thus may not have the receptor name in the cert, they should add `SkipReceptorNamesCheck=true` to those tls-server/client entries in the config file.

PR #586, opened 2022-03-25
Eliminate UDP receive buffer warning

ghjm:
Quic-go v0.26.0 has added an environment variable to suppress the warning about not being able to increase the UDP receive buffer size.  This PR updates Receptor to 0.26.0 and sets the needed environment variable.


shanemcd (approved):
Appreciate it, @ghjm

Issue #592, opened 2022-04-19
add required docs for receptorNames change


sarabrajsingh:
introduce docs for this change the got merged in - https://github.com/ansible/receptor/pull/578

AlanCoding:
@sarabrajsingh you merged the docs PR, is this done?

PR #593, opened 2022-04-19
adding documentation around skipreceptornamescheck and new certificate validation in receptor

sarabrajsingh:
documentation around the receptorNames change.

original (code) issue - https://github.com/ansible/receptor/issues/420
github issue to track documentation piece - https://github.com/ansible/receptor/issues/592
PR - https://github.com/ansible/receptor/pull/578

### DIFF
--- a/docs/source/tls.rst
+++ b/docs/source/tls.rst
@@ -4,6 +4,8 @@ TLS support
 Receptor supports mutual TLS authentication and encryption for above and below the
 mesh connections.
 
+If you decide to consume TLS authentication in your mesh, the certificates OIDs (1.3.6.1.4.1.2312.19.1) will be verified against the `node.id` specified in the configuration file. If there is no match, the receptor binary will hard exit. You can `Skip Certificate Validation`_ as well.
+
 Configuring TLS
 ^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


docs/source/tls.rst
```diff
@@ -4,6 +4,8 @@ TLS support
 Receptor supports mutual TLS authentication and encryption for above and below the
 mesh connections.
 
+If you decide to consume TLS authentication in your mesh, the certificates OIDs (1.3.6.1.4.1.2312.19.1) will be verified against the `node.id` you specify in the configuration file. If there isn't a match, the receptor binary will hard exit. You can `Skip Certificate Validation`_ as well.
```
Akasurde:
```suggestion
If you decide to consume TLS authentication in your mesh, the certificates OIDs (1.3.6.1.4.1.2312.19.1) will be verified against the `node.id` specified in the configuration file. If there is no match, the receptor binary will hard exit. You can `Skip Certificate Validation`_ as well.
```

docs/source/tls.rst
```diff
@@ -113,6 +115,43 @@ To find the fingerprint of a given certificate, use the following OpenSSL comman
 
 SHA256 and SHA512 fingerprints are supported.  SHA1 fingerprints are not supported due to the insecurity of the SHA1 algorithm.
 
+Skip Certificate Validation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you are in a development environment and need to turn off certificate validation, you can add a `SkipReceptorNamesCheck` key value pair in your configuration file for `tls-server`, `tls-config` or both.
```
Akasurde:
```suggestion
If you are in a development environment and need to turn off certificate validation, you can add a `SkipReceptorNamesCheck` key-value pair in your configuration file for `tls-server`, `tls-config`, or both.
```

sarabrajsingh:
thanks for the review @Akasurde - i will commit your changes as they look good & logical.

docs/source/tls.rst
```diff
@@ -113,6 +115,43 @@ To find the fingerprint of a given certificate, use the following OpenSSL comman
 
 SHA256 and SHA512 fingerprints are supported.  SHA1 fingerprints are not supported due to the insecurity of the SHA1 algorithm.
 
+Skip Certificate Validation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you are in a development environment and need to turn off certificate validation, you can add a `SkipReceptorNamesCheck` key-value pair in your configuration file for `tls-server`, `tls-config`, or both.
```
fosterseth: there is a valid, production-level use case for having this set to True. If a user has separate certs for the TLS TCP backend connection and the control service, then having the receptor name in the TLS TCP cert is not required. If the receptor name is not present, the user must set `skipreceptornamescheck` to True for that cert definition.

sarabrajsingh: i made this block of code less ambiguous. please recheck @fosterseth !

docs/source/tls.rst
```diff
@@ -4,6 +4,8 @@ TLS support
 Receptor supports mutual TLS authentication and encryption for above and below the
 mesh connections.
 
+If you decide to consume TLS authentication in your mesh, the certificates OIDs (1.3.6.1.4.1.2312.19.1) will be verified against the `node.id` specified in the configuration file. If there is no match, the receptor binary will hard exit. You can `Skip Certificate Validation`_ as well.
```
fosterseth: This paragraph would fit nicely in the Above the mesh TLS section, as this is the first section that explains the concept of the above the mesh certification.

docs/source/tls.rst
```diff
@@ -113,6 +115,42 @@ To find the fingerprint of a given certificate, use the following OpenSSL comman
 
 SHA256 and SHA512 fingerprints are supported.  SHA1 fingerprints are not supported due to the insecurity of the SHA1 algorithm.
 
+Skip Certificate Validation
```
fosterseth, line 126: I suggest we move this section below `Above the mesh TLS`

docs/source/tls.rst
```diff
@@ -113,6 +115,42 @@ To find the fingerprint of a given certificate, use the following OpenSSL comman
...
+.. code-block:: yaml
+
+    ---
+    - node:
+        id: bar
+
+    - log-level:
+        level: Debug
+
+    - tls-client:
```
fosterseth, line 141: I think for clarity we can just have a truncated config snippet here, with just the tls-client and tls-server portions. We do this elsewhere in the docs.

sarabrajsingh: i'd like to keep the `node.id` section in the code snippet, just to highlight the new relationship we're introducing between the certificates and the `node.id`. what do you think?

fosterseth (approved):

PR #607, opened 2022-05-03
adding CLI for MaxConnectionIdleTime

sarabrajsingh:
upstream issue - https://github.com/ansible/tower/issues/5777

adding ability to provide a `MaxIdleConnectionTimeout` parameter in a configuration file to keep backend `below-the-mesh` TCP connections alive.

to-do:

- [x] unit tests


pkg/netceptor/netceptor.go
```diff
@@ -441,6 +441,27 @@ func (s *Netceptor) MaxConnectionIdleTime() time.Duration {
...
+	duration, err := time.ParseDuration(userDefinedMaxIdleConnectionTimeout)
+	if err != nil {
+		return fmt.Errorf("failed to parse MaxIdleConnectionTimeout from configuration file -- valid examples include '1.5h', '30m', '30m10s'")
+	}
+	// we don't want the user defined timeout to be less than the defaultMaxConnectionIdleTime constant
+	if duration < defaultMaxConnectionIdleTime {
+		return fmt.Errorf("user defined maxIdleConnectionTimeout [%d] is less than the default default timeout [%d]", duration, defaultMaxConnectionIdleTime)
+	}
+	// we also need to ensure that if defined, a user defined connection timeout is at least 2x longer than the defaultRouteUpdateTime value
+	if duration < (2*defaultRouteUpdateTime + 1*time.Second) {
```
fosterseth: I think this check is unnecessary, given that `defaultMaxConnectionIdleTime` is defined as 2x defaultRouteUpdateTime + 1 second, and we ensure that duration < defaultMaxConnectionIdleTime in the previous check.

cmd/receptor-cl/receptor.go
```diff
@@ -19,9 +19,10 @@ import (
 )
 
 type nodeCfg struct {
-	ID            string                       `description:"Node ID. Defaults to local hostname." barevalue:"yes"`
-	DataDir       string                       `description:"Directory in which to store node data"`
-	FirewallRules []netceptor.FirewallRuleData `description:"Firewall Rules (see documentation for syntax)"`
+	ID                       string                       `description:"Node ID. Defaults to local hostname." barevalue:"yes"`
+	DataDir                  string                       `description:"Directory in which to store node data"`
+	FirewallRules            []netceptor.FirewallRuleData `description:"Firewall Rules (see documentation for syntax)"`
+	MaxIdleConnectionTimeout string                       `description:"User defined maximum time a backend connection can go without data before we consider it failed." default:"21s"`
```
fosterseth: I'd prefer if we don't set a default here. if not defined in the config it will just be an empty string. That way we still only have one place where we define the default, the constant at the top of the netceptor.go file

fosterseth: if we remove this default, SetMaxConnectionIdleTime should return without error if given an empty string

sarabrajsingh: @fosterseth - ok good suggestion. not including a default here will resolve the previous code comment you made :)

cmd/receptor-cl/receptor.go
```diff
@@ -19,9 +19,10 @@ import (
 )
 
 type nodeCfg struct {
-	ID            string                       `description:"Node ID. Defaults to local hostname." barevalue:"yes"`
-	DataDir       string                       `description:"Directory in which to store node data"`
-	FirewallRules []netceptor.FirewallRuleData `description:"Firewall Rules (see documentation for syntax)"`
+	ID                       string                       `description:"Node ID. Defaults to local hostname." barevalue:"yes"`
+	DataDir                  string                       `description:"Directory in which to store node data"`
+	FirewallRules            []netceptor.FirewallRuleData `description:"Firewall Rules (see documentation for syntax)"`
+	MaxIdleConnectionTimeout string                       `description:"User defined maximum time a backend connection can go without data before we consider it failed."`
```
fosterseth: We may change this to,

"Max duration with no traffic before a backend connection is timed out and refreshed."

To emphasize that the backend will timeout, but then refresh automatically.

sarabrajsingh: done :-)

tests/functional/mesh/netceptor_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/ansible/receptor/tests/functional/lib/mesh"
+)
+
+func TestSetMaxConnectionIdleTimeFromPseudoConfigFile(t *testing.T) {
```
fosterseth: are these tests ensuring that the mesh fails to start because of a bad `maxidleconnectiontimeout` setting?

I'd be content with just having the 3 single node tests that you have above, since they basically test the same thing but are less involved.

sarabrajsingh: yes, thats exactly what they're doing (bad config file). we didn't have any functional tests around passing bad config files, so i thought i'd create some. we can remove these, i understand that there are already failsafes for bad configuration keys in the config file within receptor (config file checks done at bootstrap-time).

fosterseth:
pulled down and tested

`WARNING 2022/05/05 13:54:08 Timing out connection, idle for the past 40s`

Something to keep in mind, this setting is per-node, so if nodeA has a timeout set to 40s, and nodeB has the default 21s, then a backend connection between them will timeout in 21s (whichever is less).

@sarabrajsingh this would be a good candidate setting to have a blurb in the docs. I think a dedicated page for it would be fine, like we did for firewalls.rst

fosterseth (approved):

sarabrajsingh:
@fosterseth i added documentation around the `maxidleconnectiontimeout` feature. please check out the new docs and let me know what you think :)

docs/source/edge_networks.rst
```diff
@@ -0,0 +1,66 @@
...
+The configuration files for the diagram above are listed below.
+
+foo.yml
+
+.. code-block:: yaml
+
+    ---
+    - node:
+        id: foo
+        maxidleconnectiontimeout: 60s
```
fosterseth, line 37: all 3 of these nodes would need the same `maxidleconnectiontimeout`. Because the monitorConnectionAging running on bar will still timeout the connection to foo after 21 seconds (default). So the true timeout is the minimum of this value for a set of 2 peers

docs/source/edge_networks.rst
```diff
@@ -0,0 +1,66 @@
...
+Consider the following environment:
+
+.. image:: edge.png
+
+Configurable-Items
+^^^^^^^^^^^^^^^^^^
+
+Receptor encapsulates the concepts of `below-the-mesh` and `above-the-mesh` connections. Please refer to :doc:`tls` for a better understanding of these networking layers.
+
+If a particular node in a network has higher than normal latency, we allow the users to define a finely-grained idle connection timeout value for any given Receptor node. This will help Receptor keep `below-the-mesh` tcp connections alive. Receptor will attempt to reconnect to a timed-out node, to a maximum retry limit and if this retry limit is surpassed, Receptor will drop the connection from the routing table.
```
fosterseth: "to a maximum retry limit and if this retry limit is surpassed"

it's not so much retry logic and limits, rather just monitoring traffic that flows over a backend connection.

If the last statement in this paragraph should be something like,
```
Receptor will monitor backend connections for traffic and will timeout any connection that hasn't seen traffic for a period of time. Once the connection is dropped, a new connection is formed automatically. 
```

docs/source/edge_networks.rst
```diff
@@ -0,0 +1,66 @@
+Support for Edge Networks
+=========================
+
+Recptor out-of-the-box has the ability to support complicated networking environments including edge networks.
+
+Consider the following environment:
+
+.. image:: edge.png
```
fosterseth, line 8: awesome diagram and docs around this!

docs/source/edge_networks.rst
sarabrajsingh: verbatim copy/paste :) thanks for the suggesstion

fosterseth (approved):
looks great, ready to merge when rebase + squash to get the lint changes

PR #611, opened 2022-05-05
add podman-remote to image

TheRealHaoLiu:
This PR modifies the Dockerfile to add `podman-remote` to the container and symlink `podman-remote` as `podman`

The goal is to have containerized receptor be able to create container on the host machine

Symlink `podman-remote` to `podman` avoids the AWX having to know the differences between receptor running in container and receptor running directly on VM.

### DIFF
--- a/packaging/container/Dockerfile
+++ b/packaging/container/Dockerfile
@@ -22,7 +22,9 @@ COPY receptor.conf /etc/receptor/receptor.conf
 
 RUN dnf -y update && \
     dnf -y install python3-pip && \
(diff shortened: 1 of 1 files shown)


eqrx:
Hi, thank you for the PR! May I ask what your use case is for this? :)

shanemcd:
Hao is working on a prototype for a hosted service offering. This actually isn’t the right image to add this. @TheRealHaoLiu let’s chat later today. I think you may want to build a new EE for this, I don’t think it belongs in any of our upstream images.

PR #612, opened 2022-05-09
fixing linting issues introduced by golangci-lint 1.46.0

sarabrajsingh:
thought about pinning golangci-lint version to something like `1.45.0` but this new update exposes changes that should be made.

### DIFF
--- a/pkg/workceptor/lock_test.go
+++ b/pkg/workceptor/lock_test.go
@@ -63,11 +63,11 @@ func TestStatusFileLock(t *testing.T) {
 				}
 				fileHasExisted = true
 				if err != nil {
-					t.Fatal(fmt.Sprintf("Error loading status file: %s", err))
+					t.Fatalf(fmt.Sprintf("Error loading status file: %s", err))
 				}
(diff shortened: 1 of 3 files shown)


pkg/workceptor/lock_test.go
```diff
@@ -63,11 +63,11 @@ func TestStatusFileLock(t *testing.T) {
 				}
 				fileHasExisted = true
 				if err != nil {
-					t.Fatal(fmt.Sprintf("Error loading status file: %s", err))
+					t.Fatalf(fmt.Sprintf("Error loading status file: %s", err))
```
eqrx: Shouldn't all of those not just be `t.Fatalf("Error loading status file: %s", err)`?

eqrx (changes requested):

pkg/workceptor/lock_test.go
sarabrajsingh: @eqrx you are right, will make the change now.

eqrx (approved):

fosterseth (approved):

PR #636, opened 2022-06-30
Fix goroutine leak in utils.BridgeConns

shyiko:
one of the two `done <- true` (in bridgeHalf) blocks forever as "receive" from unbuffered done channel is executed only once


fosterseth:
good catch

fosterseth (approved):

PR #641, opened 2022-07-11
CloseConnection considered harmful

ghjm:
What's wrong with this code?  (Other than the lack of error checking.)

```
n := netceptor.New(...)
conn, _ := n.Dial("node2", "mysvc", nil)
fmt.Fprintf(conn, "Hello, world!\n")
conn.Close()
```

Answer: It doesn't close the connection.  A PacketConn and QUIC connection, with the associated dozen-or-so goroutines, are still running, and will remain running for the life of your program.

Somehow (and I'm well aware that this is code that I probably wrote), in Receptor-land it turns out that `Close()` has the meaning we would normally associate with `CloseWrite()` - it shuts down the write side of the connection and indicates an EOF to the other side, but leaves the connection open and ready to continue receiving data from the other side.  If you want to _actually close_ the Receptor connection, you have to call `CloseConnection()`, which has the effect we would normally associate with `Close()`.

If you don't know about this and just call `Close()`, expecting Receptor to behave like any other networking library, you'll wind up leaking goroutines and memory like crazy.  And given that `CloseConnection()` only appears in the Receptor code in a few places in Workceptor, I'm imagining that long-running Receptor instances probably _do_ leak goroutines and memory like crazy.

This PR renames the functions to the sensible names, and updates Workceptor to call the right functions.  If other parts of the code are relying on `Close()` not really closing the connection, then this will break them, but arguably they were already broken since they were leaking connections.  Also, this is a breaking change for API users, so should be in a new minor release.

### DIFF
--- a/pkg/netceptor/conn.go
+++ b/pkg/netceptor/conn.go
@@ -406,6 +406,7 @@ func (c *Conn) Read(b []byte) (int, error) {
 	if ok && aerr.ErrorMessage == "normal close" {
 		err = io.EOF
 	}
+
 	return n, err
 }
 
@@ -421,6 +422,7 @@ func (c *Conn) Write(b []byte) (int, error) {
 	if ok && aerr.ErrorMessage == "normal close" {
 		err = net.ErrClosed
 	}
+
 	return n, err
 }
 
--- a/pkg/netceptor/conn.go
+++ b/pkg/netceptor/conn.go
@@ -9,6 +9,7 @@ import (
 	"crypto/x509/pkix"
 	"encoding/pem"
 	"fmt"
+	"io"
 	"math/big"
 	"net"
 	"os"
@@ -399,13 +400,13 @@ func monitorUnreachable(pc *PacketConn, doneChan chan struct{}, remoteAddr Addr,
 }
 
 // Read reads data from the connection.
-func (c *Conn) Read(b []byte) (n int, err error) {
-	n, err = c.qs.Read(b)
+func (c *Conn) Read(b []byte) (int, error) {
+	n, err := c.qs.Read(b)
 	aerr, ok := err.(*quic.ApplicationError)
 	if ok && aerr.ErrorMessage == "normal close" {
-		err = nil
+		err = io.EOF
 	}
-	return
+	return n, err
 }
 
 // CancelRead cancels a pending read operation.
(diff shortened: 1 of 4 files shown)


AaronH88:
Hi @ghjm 
I am reviewing all PRs to tidy up and I really like this one.
Are you ok to rebase? 
No problem if not, I can take this as inspiration and create a new PR :smile:

pkg/netceptor/conn.go
```diff
@@ -412,27 +421,49 @@ func (c *Conn) CancelRead() {
...
 	})
 
 	return c.qs.Close()
 }
 
-func (c *Conn) CloseConnection() error {
-	c.pc.cancel()
+// Close closes the connection.
+func (c *Conn) Close() error {
+	if !c.fromListener {
```
AaronH88, line 440: Can you explain why this packet conn only can be cancelled this way please?

I am struggling to understand when `fromListener` will be set to `false`

pkg/backends/websockets.go
```diff
@@ -140,8 +140,9 @@ func (b *WebsocketListener) Start(ctx context.Context, wg *sync.WaitGroup) (chan
 	}
 	wg.Add(1)
 	b.server = &http.Server{
-		Addr:    b.address,
-		Handler: mux,
+		Addr:              b.address,
+		Handler:           mux,
+		ReadHeaderTimeout: 10 * time.Second,
```
AaronH88: any other reason for this timeout other than tests?

pkg/netceptor/conn.go
ghjm: If I remember right, fromListener is set based on whether a Conn object was created by the user opening it, or was created by a listener receiving a connection.  If the Conn is standalone, then when we close it, we need to close the underlying PacketConn.  But if the Conn came from a listener, the PacketConn is shared, and should only be closed when the listener is closed.

pkg/backends/websockets.go
ghjm: Yes, this is a mitigation for the Slowloris attack.

ghjm:
I can take a look at rebasing it later today, but I don't have time to fix non-trivial issues with it if it doesn't apply cleanly, or fails unit testing or what have you.

I vaguely remember there being some problem why this didn't get merged, but I don't remember what.

AaronH88:
> I can take a look at rebasing it later today, but I don't have time to fix non-trivial issues with it if it doesn't apply cleanly, or fails unit testing or what have you.
> 
> I vaguely remember there being some problem why this didn't get merged, but I don't remember what.

Sounds great @ghjm 

Let me know, happy to take over this PR if its more suitable as I think it adds a lot of value! 
Thanks :smile:

ghjm:
@AaronH88 if you've got time to work on this then maybe you'd better take it over.  I'd love to mess with it but I'm just not finding the time.

AaronH88:
> @AaronH88 if you've got time to work on this then maybe you'd better take it over. I'd love to mess with it but I'm just not finding the time.

No problem at all! 
Thanks for getting back to me.

PR #651, opened 2022-08-09
Remove go v1.17 from build matrix, use go v1.19 by default

eqrx:


shanemcd:
I cherry-picked this into https://github.com/ansible/receptor/pull/653, sans 1.17 removal.

PR #652, opened 2022-08-15
Publish ppc64le and ARM images

shanemcd:


Sunidhi-Gaonkar1:
Hello @shanemcd -Even after merging the changes there is no multiarch image being built. https://github.com/ansible/receptor/runs/7848442327?check_suite_focus=true#step:9:1 Thank you in advance for your time!

ssbarnea:
@Sunidhi-Gaonkar1 My impression is that it works fine, I pulled the container and got the arch64 one. The real problem is the building speed, or lack of if TBH.

PR #673, opened 2022-09-30
Retry logic for stdin and stdout from pod

fosterseth:
Currently receptor has one attempt at opening stdin and stdout streams from the pod.

An error during this prevents any chance of the pod running successfully.

This change provides some retry logic when opening these streams to offer better stability when receptor interacts with a k8s cluster.

see https://github.com/ansible/awx/issues/12881

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -295,7 +295,16 @@ func (kw *kubeUnit) runWorkUsingLogger() {
 		Container: "worker",
 		Follow:    true,
 	})
-	logStream, err := logreq.Stream(kw.ctx)
+	var logStream io.ReadCloser
+	for retries := 5; retries > 0; retries-- {
+		logStream, err = logreq.Stream(kw.ctx)
+		if err != nil {
+			logger.Warning("Problem opening stdout from pod %s, unit %s. Retrying.", kw.pod.Name, kw.unitID)
+			time.Sleep(time.Second * 5)
+		} else {
+			break
+		}
+	}
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/workceptor/kubernetes.go
```diff
@@ -295,7 +295,14 @@ func (kw *kubeUnit) runWorkUsingLogger() {
 		Container: "worker",
 		Follow:    true,
 	})
-	logStream, err := logreq.Stream(kw.ctx)
+	var logStream io.ReadCloser
+	for retries := 0; retries < 5; retries++ {
+		logStream, err = logreq.Stream(kw.ctx)
+		if err != nil {
```
shanemcd, line 301: Shouldn't we be breaking out of this loop early if err is nil?

pkg/workceptor/kubernetes.go
```diff
@@ -397,10 +404,16 @@ func (kw *kubeUnit) runWorkUsingLogger() {
...
-			errStdin = exec.Stream(remotecommand.StreamOptions{
-				Stdin: stdin,
-				Tty:   false,
-			})
+			for retries := 0; retries < 5; retries++ {
+				errStdin = exec.Stream(remotecommand.StreamOptions{
+					Stdin: stdin,
+					Tty:   false,
+				})
+				if errStdin != nil {
```
shanemcd, line 414: Same question here - unless I am missing something won't we always run this 5 times?

pkg/workceptor/kubernetes.go
fosterseth: you're right, missing a break

shanemcd (approved):

PR #683, opened 2022-10-25
Add retry mechanism to handle intermittent connection issues with Kubernetes logging stream

TheRealHaoLiu:
Due to a issue in Kubernetes, AWX can't currently run jobs longer than 4 hours when deployed on Kubernetes. More context on that in https://github.com/ansible/awx/issues/11805.

This PR adds logic that will pick back up from the last line we saw, using Kubernetes log timestamp 

require fix in Kubernetes for [Pod logs: long lines are corrupted when using timestamps=true](https://github.com/kubernetes/kubernetes/issues/77603)
fixed in https://github.com/kubernetes/kubernetes/pull/113481

Fixe backported into Kubernetes release branches in the following PRs:
[release-1.23](https://github.com/kubernetes/kubernetes/pull/113517) (1.23.14)
[release-1.24](https://github.com/kubernetes/kubernetes/pull/113516) (1.24.8)
[release-1.25](https://github.com/kubernetes/kubernetes/pull/113515) (1.25.4)

Fixes ported to OpenShift in the following PRs
[release-4.9](https://github.com/openshift/kubernetes/pull/1417) (4.9.x) (not yet merged)
[release-4.10](https://github.com/openshift/kubernetes/pull/1415) (4.10.42)
[release-4.11](https://github.com/openshift/kubernetes/pull/1408) (4.11.16)
[release-4.12](https://github.com/openshift/kubernetes/pull/1407) (4.12.0)

the fix in this PR should detect the Kubernetes version and use `--timestamp` according however due to the "wild wild west" nature of Kubernetes world we added `RECEPTOR_KUBE_SUPPORT_RECONNECT` environment variable to force enable/disable the fix.

`RECEPTOR_KUBE_SUPPORT_RECONNECT`  have following options:
- “enabled”: this option will use timestamp with the log and enable our new code path
- “disabled”: this option will not use timestamp and use the original code path
- “auto”: auto detect if it's appropriate to enable timestamp base on kube version

this flag can be set via awx custom resource 
```
apiVersion: awx.ansible.com/v1beta1
kind: AWX
metadata:
  name: awx
  namespace: awx
spec:
  …
  ee_extra_env: |
      value: enabled
```

NOTE: `RECEPTOR_KUBE_SUPPORT_RECONNECT` will bypass check for ALL container group when set to "enabled/disabled". If a specific container group is not contain have the right Kubernetes version and `RECEPTOR_KUBE_SUPPORT_RECONNECT` is set to enabled job execution with that container group will fail due to corrupted log stream 

NOTE: it is also possible for kublet version to be different from kube-apiserver detecting the presence of the fix using kube-apiserver version is not the safest option, but its the only option we have here.

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -4,6 +4,7 @@
 package workceptor
 
 import (
+	"bufio"
 	"context"
 	"errors"
 	"fmt"
@@ -23,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/version"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -194,15 +196,19 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
 		}
 		pod.Spec.Containers[0].Env = evs
 	}
+
+	// get pod and store to kw.pod
 	kw.pod, err = kw.clientset.CoreV1().Pods(ked.KubeNamespace).Create(kw.ctx, pod, metav1.CreateOptions{})
 	if err != nil {
 		return err
 	}
+
 	select {
 	case <-kw.ctx.Done():
 		return fmt.Errorf("cancelled")
 	default:
 	}
+
 	kw.UpdateFullStatus(func(status *StatusFileData) {
 		status.State = WorkStatePending
 		status.Detail = "Pod created"
@@ -229,26 +235,31 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
 	if kw.podPendingTimeout != time.Duration(0) {
 		ctxPodReady, _ = context.WithTimeout(kw.ctx, kw.podPendingTimeout)
 	}
+
 	ev, err := watch2.UntilWithSync(ctxPodReady, lw, &corev1.Pod{}, nil, podRunningAndReady())
 	if ev == nil || ev.Object == nil {
 		return fmt.Errorf("did not return an event while watching pod for work unit %s", kw.ID())
 	}
+
 	var ok bool
 	kw.pod, ok = ev.Object.(*corev1.Pod)
 	if !ok {
 		return fmt.Errorf("watch did not return a pod")
 	}
+
 	if err == ErrPodCompleted {
+		// Hao: shouldn't we also call kw.Cancel() in these cases?
 		if len(kw.pod.Status.ContainerStatuses) != 1 {
 			return fmt.Errorf("expected 1 container in pod but there were %d", len(kw.pod.Status.ContainerStatuses))
 		}
+
 		cstat := kw.pod.Status.ContainerStatuses[0]
 		if cstat.State.Terminated != nil && cstat.State.Terminated.ExitCode != 0 {
 			return fmt.Errorf("container failed with exit code %d: %s", cstat.State.Terminated.ExitCode, cstat.State.Terminated.Message)
 		}
 
 		return err
-	} else if err != nil {
+	} else if err != nil { // any other error besides ErrPodCompleted
 		kw.Cancel()
 		if len(kw.pod.Status.ContainerStatuses) == 1 {
 			if kw.pod.Status.ContainerStatuses[0].State.Waiting != nil {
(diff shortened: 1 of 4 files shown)


egmar:
Hi,
is this PR also covering the issue described here https://github.com/ansible/awx/issues/11805?

shanemcd:
@egmar Yes, we're trying to figure that out. Hopefully along with https://github.com/ansible/awx/issues/11338

Vitexus:
I can confirm, that testing job keep running whole night.

```yaml
---
- hosts: "{{ target|default('localhost') }}"
  become: false
  gather_facts: false
  tasks:
   - name: Run Job
     ansible.builtin.shell: 'while true ; do date ; sleep 3600; done'
     register: long_output_job
     ignore_errors: true

```

Good job! 
Many Thanks  to @TheRealHaoLiu and looking forward The release 1.2.4.

thees (approved):

shanemcd:
@Vitexus You are awesome, thanks for testing. We iterated a bit more. Can you try with the latest code here?

Vitexus:
Sure.

obrazek

shanemcd:
I found another bug where it breaks when very long lines are emitted. Will try to figure out a solution later today.

shanemcd:
Well... it's not a bug in _this_ code 😰 

I have a pod that starts and runs this:

```
for i in {1..30}; do echo $i; printf "%0.s-" {1..100000}; echo; done
```

When running `$ kubectl logs echo-tcfjg --timestamps`, I can see random timestamps in the middle of the output:

image

For whatever reason, there are only 16385 bytes printed, followed by a timestamp, then more bytes.

Since AWX expects each line to be either in-tact json or a full base64 encoded blob of zip data, this will definitely break certain workloads ☹️

shanemcd:
Maybe related: https://github.com/kubernetes/kubernetes/pull/77822

TheRealHaoLiu:
@Vitexus btw love your github bio
are u on our matrix server by any chance? https://matrix.to/#/#awx:ansible.com (there may be a rare IRL angry spud plushy in it for you 😉 )

shanemcd:
The problem seems to be even worse in OpenShift. There's only 4097 bytes between the timestamps...

Vitexus:
obrazek

Sad 👎

Vitexus:
Ok, next try with latest changes tomorrow  ...

TheRealHaoLiu:
latest idea is to just use local time to resume the log the implementation
this implementation have the problem that we may get repeated lines in the log

if the last few line of job output gets repeated this cause a problem with `ansible-runner process` and could cause ansible-runner to hang forever

gonna pivot to https://github.com/ansible/receptor/pull/685 to leverage socat and portfowarding to stream the log back via TCP

gundalow:
Hao & Seth, can you please update this and get it tested against https://github.com/kubernetes/kubernetes/pull/113481

Thanks

TheRealHaoLiu:
good news, we were able to convince Kubernetes maintainer that the log issue @shanemcd discover is indeed a bug 🥳 

detail of the fix is in https://github.com/kubernetes/kubernetes/pull/113481

with that we can proceed with flushing this PR out 

now we should be able to use exact log timestamp to prevent duplication of log messages and resume from the specific message at the time of disconnection

TheRealHaoLiu:
@Vitexus thanks for caring about this topic and helping us out with this issue, we really appreciate it. Have you heard about our community Matrix chatroom https://matrix.to/#/#awx:ansible.com. Its a great place for collaborating with the AWX community, join us!

pkg/workceptor/kubernetes.go
```diff
@@ -263,182 +273,326 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
...
+				break
+			}
+
+			// get pod, with retry
+			for retries := 5; retries > 0; retries-- {
+				kw.pod, err = kw.clientset.CoreV1().Pods(podNamespace).Get(kw.ctx, podName, metav1.GetOptions{})
+				if err == nil {
+					break
+				} else {
+					time.Sleep(100 * time.Millisecond)
```
shanemcd: On my machine, this wasn't quite long enough to recover gracefully when the kubelet is restarted. Changing this to 1 second made it work. I see where we changed 4 other places from 5 seconds to 100 milliseconds. Let's go with somewhere in the middle... maybe 1 or 2 seconds.

TheRealHaoLiu:
got a test image up `quay.io/haoliu/awx-ee:kubernetes-logstream-retry` (will continue to update to track the latest change)

TheRealHaoLiu:
https://github.com/kubernetes/kubernetes/pull/113481 merged, in our logic we do need to deal with and preserve the old behavior (where job fail after 4hr) if k8s does not contain the logging fix so if anyone have suggestion for how we can deal with that gracefully please comment

we might need a external param to detect if the deployment is on a kube with the patch and enable/disable timestamp accordingly

egmar:
> [kubernetes/kubernetes#113481](https://github.com/kubernetes/kubernetes/pull/113481) merged, in our logic we do need to deal with and preserve the old behavior (where job fail after 4hr) if k8s does not contain the logging fix so if anyone have suggestion for how we can deal with that gracefully please comment
> 
> we might need a external param to detect if the deployment is on a kube with the patch and enable/disable timestamp accordingly

I'd use the https://pkg.go.dev/k8s.io/client-go/discovery#DiscoveryClient.ServerVersion and fallback to old behaviour on K8S without the fix.

pkg/workceptor/kubernetes.go
```diff
@@ -263,182 +273,326 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
...
+
+						break
+					}
+
+					return
+				}
+				if err != nil {
+					stdoutErr = err
+
+					break
```
fosterseth: we don't check for stdoutErr anywhere in the outer loop -- this could result in an infinite loop. maybe we need to return here

pkg/workceptor/kubernetes.go
```diff
@@ -263,182 +273,326 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
...
+					SinceTime:  &metav1.Time{Time: sinceTime},
+				},
+			)
+			// get logstream, with retry
+			for retries := 5; retries > 0; retries-- {
+				logStream, err = logReq.Stream(kw.ctx)
+				if err == nil {
+					break
+				} else {
+					time.Sleep(100 * time.Millisecond)
```
shanemcd: Let's add a logger.warning here.

pkg/workceptor/kubernetes.go
shanemcd: logger.warn here too.

pkg/workceptor/kubernetes.go
shanemcd: damn good catch

pkg/workceptor/kubernetes.go
```diff
@@ -263,182 +273,326 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
...
+
+				break
+			}
+
+			// read from logstream
+			streamReader := bufio.NewReader(logStream)
+			for stdinErr == nil { // check between every line read to see if we need to stop reading
+				line, err := streamReader.ReadString('\n')
+				if err == io.EOF {
+					kw.Debug("detected EOF for pod %s/%s, attempt %d", podNamespace, podName, numEOF)
```
shanemcd:
```suggestion
					kw.Debug("detected EOF for pod %s/%s, attempt %d", podNamespace, podName, numEOF+1)
```

pkg/workceptor/kubernetes.go
```diff
@@ -263,182 +273,326 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
...
+					numEOF++
+					if numEOF <= 5 {
```
shanemcd:
```suggestion
					if numEOF < 5 {
```

TheRealHaoLiu:
@egmar yah that's definitely one of the strategy we can use literally test server version there's one slight concern is that we need to know what version of this gets backported to for different version of kube...

awhile ago i was doing this for something i was working on and find that what goes into which version of kube in the different kube varient is kinda wild wild west

little bit hesitant but i guess i can ask in the kubernetes slack to see if there's more order now

pkg/workceptor/kubernetes.go
```diff
@@ -263,182 +273,331 @@ func (kw *kubeUnit) createPod(env map[string]string) error {
...
+				errMsg := fmt.Sprintf("Error opening log stream for pod %s/%s: %s", podNamespace, podName, err)
+				kw.UpdateBasicStatus(WorkStateFailed, errMsg, 0)
+				kw.Error(errMsg)
+
+				break
+			}
+
+			// read from logstream
+			streamReader := bufio.NewReader(logStream)
+			for stdinErr == nil { // check between every line read to see if we need to stop reading
```
shanemcd, line 587: Why are we checking `stdinErr` here?

pkg/workceptor/kubernetes.go
```diff
@@ -651,6 +654,63 @@ func (kw *kubeUnit) runWorkUsingLogger() {
...
+	if err != nil {
+		logger.Warning("could parse Kubernetes server version %s, will not use reconnect support", serverVerInfo.String())
+
+		return false
+	}
+
+	// The patch was backported to minor version 23 and 24. We must check z stream
+	// based on the minor version
+	var compatibleVer string
+	switch serverVerInfo.Minor {
```
gundalow, line 706: I might be reading ths wrong, though what happens when `serverVerInfo.Minor` is 26? is `compatibleVer` null, does `semver.AtLeast` handle that correctly?

What about `2.23.1`?

fosterseth: good point, just fixed this to use a default in case we are given a minor version that is NOT 23, 24, or 25. By default we compare the server version with `v1.23.14`.

We aren't explicitly handling the Major version, but we should still get the correct return value each time.

0.23.14 - return false
2.23.14 - return true

the edge cases are really if the minor version is 24 or 25

0.24.8 - return false
2.24.8 - return true

same with 25

M4XMU3:
This fix would be awesome as some teams in our company are turning away from ansible as they have very long lasting jobs.
Is there a chance to get an approximate guess when this will be merged to master? :)

M4XMU3:
This fix would be awesome as some teams in our company are turning away from ansible as they have very long lasting jobs.

fosterseth:
@M4XMU3 we are currently doing internal testing on this branch, but so far things are looking promising. I'd say there is a good chance we'll be able to merge this in over the next couple of weeks, and we'll be sure to kick off an AWX release with this change.

Our solution does require a k8s patch, so you'll want to run on a cluster that has the following versions (or newer)
https://github.com/ansible/receptor/pull/683/files#diff-792611baeb730234abface9c9bc33204ea86453469e7fcdfbe8de8d4e04f2598R659

obaranov (approved):
Looks good from QE. Verified against 4.10 patched cluster as well as against old cluster.

oliverf1:
Hello,
I can also confirm that this fixed the issue for me :-) Thank you all for the fix!
However, I had to build a custom image for the execution environment because in the Containerfile of quay.io/ansible/awx-ee:latest it references quay.io/ansible/receptor:devel but this patch is only available in the latest image
So is there any chance that the devel image being updated with this 1.3.0 version, It's currently using the 1.3.0.dev2 which is older? Or should I create a MR in the ansible/awx-ee repo to update the Containerfile to target quay.io/ansible/receptor:latest?

kingtutt1906:
> Hello, I can also confirm that this fixed the issue for me :-) Thank you all for the fix! However, I had to build a custom image for the execution environment because in the Containerfile of quay.io/ansible/awx-ee:latest it references quay.io/ansible/receptor:devel but this patch is only available in the latest image So is there any chance that the devel image being updated with this 1.3.0 version, It's currently using the 1.3.0.dev2 which is older? Or should I create a MR in the ansible/awx-ee repo to update the Containerfile to target quay.io/ansible/receptor:latest?

Could you show your custom execution environment build? I tried using the  quay.io/ansible/awx-ee:latest execution environment and it still timed out after 4hrs.

oliverf1:
> > Hello, I can also confirm that this fixed the issue for me :-) Thank you all for the fix! However, I had to build a custom image for the execution environment because in the Containerfile of quay.io/ansible/awx-ee:latest it references quay.io/ansible/receptor:devel but this patch is only available in the latest image So is there any chance that the devel image being updated with this 1.3.0 version, It's currently using the 1.3.0.dev2 which is older? Or should I create a MR in the ansible/awx-ee repo to update the Containerfile to target quay.io/ansible/receptor:latest?
> 
> Could you show your custom execution environment build? I tried using the quay.io/ansible/awx-ee:latest execution environment and it still timed out after 4hrs.

I never experienced the 4h timeout (because I don´t have such long jobs), but I was experiencing the issue with the log rotation a lot.
The patch is now available in the quay.io/ansible/awx-ee:latest default image, No need to rebuild an image
`$ docker run -it quay.io/ansible/awx-ee:latest receptor --version`
`1.3.1+g3c0278f`

The patch was added in version 1.3.0

PR #693, opened 2022-11-29
Detect OCP version to determine reconnect support

fosterseth:
followup to https://github.com/ansible/receptor/pull/683

It doesn't suffice to check K8S server version if running on OpenShift, as the K8S versions reported don't align with upstream.

Therefore, we should try to first detect the OCP server version, and use that to determine if reconnect support should be used.

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/version"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -36,7 +35,6 @@ linters:
     - gosec
     - gosimple
     - govet
-    - ifshort
```
fosterseth: these are deprecated, removing.

pkg/workceptor/kubernetes.go
```diff
@@ -653,10 +655,142 @@ func (kw *kubeUnit) runWorkUsingLogger() {
...
+	// if minor version == 11, compare with v4.11.16
+	// all other minor versions compare with v4.10.42
+	var compatibleVer string
+	switch semver.Minor() {
+	case 12:
+		compatibleVer = "v4.12.0"
+	case 11:
+		compatibleVer = "v4.11.16"
+	default:
+		compatibleVer = "v4.10.44"
```
TheRealHaoLiu: 4.10.42

pkg/workceptor/kubernetes_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+func Test_isCompatibleOCP(t *testing.T) {
+	type args struct {
+		version      string
+		isCompatible bool
+	}
+
+	kw := &kubeUnit{}
+
+	tests := []args{
```
gundalow, line 15: I'm a lot happier now we have these version checks in. Thank you.

.golangci.yml
```diff
@@ -7,7 +7,6 @@ linters:
     - asciicheck
     - bodyclose
     #- cyclop # TODO: Reduce code complexity.
-    - deadcode
```
shanemcd: Please open a one-off PR for these changes that we will merge before this PR.

shanemcd (changes requested):
Nice work.

Before we merge this:

- Move linter changes out into a separate PR (preferably), or at least another commit.
- I don't think the changes to go.mod and go.sum are necessary. If anything they seem dangerous. Unless there was a specific reason for doing that, please revert the changes to those files.

TheRealHaoLiu:
this will require us to give the awx serviceaccount permission to get clusteroperator resource

TheRealHaoLiu:
we decided to not do this and ask user to add 
```
  ee_extra_env: |
      value: enabled
 ```
 
 in awx spec

fosterseth:
not the direction we wish to go

Issue #708, opened 2022-12-13
reconnect issue in gke version detection


thees:
The new feature of https://github.com/ansible/receptor/pull/683#event-7906156366
has an issue with the kubernetes version detection.

In our GKE cluster (running 1.24.8) we get the log line:
"Kubernetes version v1.24.8-gke.401 not at least v1.24.8, not using reconnect support"

fosterseth:
a hyphen `-` indicates pre-release in semantic versioning

so 1.24.8-gke.401 < 1.24.8 < 1.24.8+build1

https://semver.org/

`Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.`

If 1.24.8-gke.401 is indeed a build derived from 1.24.8 (and contains the essential fix) then you can force receptor to use the reconnect feature by setting the following in your AWX.yaml resource under `spec:`

```
ee_extra_env: |
  - name: RECEPTOR_KUBE_SUPPORT_RECONNECT
    value: enabled
```

thees:
@fosterseth Yes I know about semantic versioning. However GKE does not use it.
There is always something after the minor version in GKE.
See screenshot.

Yes, we already enforced it via the env variable.
I just wanted to inform you, that the version detection does not work with google cloud.
<img width="517" alt="Bildschirmfoto 2022-12-13 um 21 52 41" src="https://user-images.githubusercontent.com/1259967/207452176-759251f8-00fb-40cb-8fb9-7ef94babcbd2.png">

TheRealHaoLiu:
@thees do you mind helping us test out the https://github.com/ansible/receptor/pull/709

TheRealHaoLiu:
here's the control_plane_ee_image `quay.io/haoliu/awx-ee:ignore-prerelease` that i built

TheRealHaoLiu:
i only have access to EKS and EKS dont have the version we require yet

TheRealHaoLiu:
to test this u would need a GKE cluster with version greater than the following 
[release-1.23](https://github.com/kubernetes/kubernetes/pull/113517) (1.23.14)
[release-1.24](https://github.com/kubernetes/kubernetes/pull/113516) (1.24.8)
[release-1.25](https://github.com/kubernetes/kubernetes/pull/113515) (1.25.4)

[...]

PR #709, opened 2022-12-14
ignore pre-release in version comparison

TheRealHaoLiu:
related to https://github.com/ansible/receptor/issues/708

GCP does not use proper semvar, example:
```
1.24.8-gke.401
```

AKS does the same thing 
```
v1.21.14-eks-fb459a0
```

instead of using + for build metadata they use - which indicate pre-release

this PR will ignore `-<prerelease tag>` when doing version comparison

currently we have no clue if this mess with other "variants" of Kubernetes

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -697,6 +697,9 @@ func shouldUseReconnect(kw *kubeUnit) bool {
 		return false
 	}
 
+	// ignore pre-release in version comparison
+	semver = semver.WithPreRelease("")
+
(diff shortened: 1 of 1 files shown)


TheRealHaoLiu:
```
DEBUG 2023/01/12 21:43:50 Kubernetes version v1.25.4-gke.2100 is at least v1.25.4, using reconnect support
DEBUG 2023/01/12 21:43:50 [IEAfM7DS] streaming stdout with reconnect support
```

tested greater version and lesser version and all works as expected 

@fosterseth can i get a review

fosterseth (approved):

PR #718, opened 2023-01-03
Refactor test suite

shanemcd:
These are notes that I took when writing on this. Some of it may be hard to understand without further explanation. If/when someone else starts contributing to this I can expound further where needed, or they will continue to serve as my personal notes:

* Move away from global logger instance (could be a separate PR)
    * Parallel tests were all spewing logs to stdout via the same logger instance, making it impossible correlate a specific log message to a specific test.
    * This requires us to explicitly pass logger objects through the application. This seems to be best practice: https://gogoapps.io/blog/passing-loggers-in-go-golang-logging-best-practices/
* TestDuplicateNodeDetection race condition (could be a separate PR)
    * Node2 has not yet sent a routing update to Node0-duplicate1, so Node0-duplicate1 is still connected
    * Node0-duplicate2 tries to connect to Node2, gets rejected.
     * Solution: only launch 1 duplicate node, not 5.
* Tests using work-command rely on pre-built binary for the `--command-runner` invocations.
* Fixed data race: Moved closing of recvChan from packetconn.go to netceptor.go. (could be a separate PR)
* acceptLoop did not have a context passed in (could be a separate PR)
* Netceptor.Listen was passing a context.Background() (could be a separate PR)
* acceptLoop was passing a context.Background() (could be a separate PR)
* Introduced sequenceLock to fix data race (could be a separate PR)
* Fixed race condition where knownNodeLock never got unlocked (could be a separate PR)
* Fixed race condition where 2 nodes with the same id could both establish connections at the same time (moved connLock) (could be a separate PR)
* Fixed websockets TLS by adding stuff to CipherSuites (could be a separate PR)
* Introduced WorkStateCanceled (could be a separate PR if we actually want this)
* Added lastUpdateErrorLock (could be a separate PR)

Todos:

- [x] Need to go through and mark appropriate struct fields as private
- [x] Go through and tighten up the contexts we pass through / cancel


AaronH88:
For the last few days I have been trying to take out smaller parts, but at this stage the PR seems to be tightly coupled. 
I am proposing reviewing and submitting the PR as it stands now, it will be less work in the long run. I will just take my time and review it slowly from here.

pkg/workceptor/command.go
```diff
@@ -281,6 +292,8 @@ func (cw *commandUnit) Cancel() error {
 
 	proc.Wait()
 
+	cw.UpdateBasicStatus(WorkStateCanceled, "Work unit canceled", -1)
```
AaronH88: Perhaps just "Cancelled" would be more appropriate.

pkg/workceptor/command.go
```diff
@@ -379,40 +400,63 @@ func filenameExists(filename string) error {
...
 
-func (cfg signingKeyPrivateCfg) Prepare() error {
+func (cfg SigningKeyPrivateCfg) Prepare() error {
+	duration, err := cfg.PrepareSigningKeyPrivateCfg()
+
+	if err != nil {
+		return fmt.Errorf(err.Error())
+	}
+
+	MainInstance.SigningExpiration = *duration
```
AaronH88, line 410: I might be misunderstanding this, but here is what I think is happening.
Main Instance is the global netceptor instance.
Assigning the keys to the global instance will allow all instances to use it, which is fine, however, will we ever have a case where we have multiple receptor instances, some signed and others unsigned?

And if yes, should netceptor be passed into this function instead? `Prepare(n *netceptor)`

pkg/workceptor/command.go
```diff
@@ -379,40 +400,63 @@ func filenameExists(filename string) error {
...
+	return nil, nil
+}
+
+func (cfg VerifyingKeyPublicCfg) Prepare() error {
+	err := filenameExists(cfg.PublicKey)
+	if err != nil {
+		return err
 	}
-	MainInstance.signingKey = cfg.PrivateKey
+	MainInstance.VerifyingKey = cfg.PublicKey
```
AaronH88, line 439: same question as the signing part

receptorctl/tests/test_workunit.py
```diff
@@ -50,12 +50,12 @@ def _wait_for_workunit_state(
         if expected_state_name is not None:
             for _ in range(remaining_time):
                 status = node_controller.simple_command("work status {}".format(unitid))
+                print(status["StateName"])
```
AaronH88: was used for debugging, needs to be removed

tests/functional/mesh/lib.go
```diff
@@ -0,0 +1,796 @@
...
+// CheckRoutes returns true if every node has a route to every other node.
+func (m *LibMesh) CheckRoutes() bool {
+	meshStatus, err := m.Status()
+	if err != nil {
+		return false
+	}
+	for _, status := range meshStatus {
+		// loop over m.MeshDefinition.Nodes instead... check for NodeConfig.ID, fall back to key
+		for _, node := range m.GetNodes() {
+			// if node.GetID() != "" {
```
AaronH88: remove commented out code

tests/functional/mesh/mesh_test.go
```diff
@@ -3,167 +3,125 @@ package mesh
...
 			}
+
 			// Test that each Node can ping each Node
-			for nodeName, nodeSender := range m.Nodes() {
-				t.Logf("Pinging nodes from %s (%s)", nodeName, nodeSender.Dir())
-				controller := receptorcontrol.New()
-				t.Logf("connecting to %s", nodeSender.ControlSocket())
-				err = controller.Connect(nodeSender.ControlSocket())
+			for _, node := range m.GetNodes() {
+				// t.Logf("Pinging nodes from %s (%s)", node.GetID(), node.GetDataDir())
```
AaronH88: remove commented out code

tests/functional/mesh/mesh_test.go
```diff
@@ -223,51 +181,71 @@ func TestTraceroute(t *testing.T) {
...
+			case "ws":
+				lsofProto = "TCP"
+			case "udp":
+				lsofProto = "UDP"
+			}
+
 			// Check that the connections are closed
 			pid := os.Getpid()
-			pidString := "pid=" + strconv.Itoa(pid)
+			t.Log("Current test pid: %\n", fmt.Sprint(pid))
```
AaronH88: this log dosent seem needed, ill remove it

tests/functional/mesh/tls_test.go
```diff
@@ -35,84 +32,71 @@ func TestTCPSSLConnections(t *testing.T) {
...
+			}
+			m.GetNodes()[node1.GetID()] = node1
+
+			node2 := m.NewLibNode("node2")
+			node2.Connections = []Connection{
+				{RemoteNode: node1, Protocol: proto, TLS: "client"},
+			}
+			node2.TLSServerConfigs = []*netceptor.TLSServerConfig{
+				{
+					Name:              "server",
```
AaronH88, line 65: maybe change name to server2

tests/functional/mesh/work_test.go
```diff
@@ -1,1003 +1,684 @@
...
 			if err != nil {
-				t.Fatal(err, m.Dir())
+				t.Fatal(err, m.GetDataDir())
 			}
 			err = controllers["node1"].AssertWorkResults(unitID, expectedResults[:1])
 			if err != nil {
-				t.Fatal(err, m.Dir())
+				t.Fatal(err, m.GetDataDir())
 			}
+			t.Log("Shutting down node2.")
```
AaronH88: looks like debugging logs, I will remove

pkg/workceptor/command.go
AaronH88: Refactored

receptorctl/tests/test_workunit.py
AaronH88: Removed

tests/functional/mesh/mesh_test.go
AaronH88: Removed

tests/functional/mesh/mesh_test.go
AaronH88: Removed

tests/functional/mesh/work_test.go
AaronH88: Removed

tests/functional/mesh/lib.go
AaronH88: Removed

pkg/workceptor/command.go
shanemcd: Isn't `Prepare` called by the `cmdline` package? I don't think we can influence which arguments this is called with. But I reserve the right to be wrong. 🙂

AaronH88: I get you! 
Happy to resolve this.

AaronH88 (approved):
LGTM :airplane:

pkg/logger/logger.go
```diff
@@ -283,7 +283,7 @@ func (cfg traceCfg) Prepare() error {
 }
 
 func init() {
-	logLevel = InfoLevel
+	logLevel = DebugLevel
```
shanemcd: @AaronH88 I can't remember if I changed this or if it was you. Seems like maybe something we did while debugging and forgot to flip back?

AaronH88: Cant remember :laughing: happy to flip it back :smile:

PR #723, opened 2023-01-20
Added Receptor logger, added an explicit logger to netceptor and used…

AaronH88:
… this logger in netceptor.go

and the blog here: https://gogoapps.io/blog/passing-loggers-in-go-golang-logging-best-practices/#global-state

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -5,9 +5,13 @@ import (
 	"io"
 	"log"
 	"os"
+	"path"
+	"runtime"
 	"strings"
+	"time"
 
 	"github.com/ghjm/cmdline"
+	logrus "github.com/sirupsen/logrus"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


pkg/netceptor/tlsconfig.go
```diff
@@ -150,7 +151,7 @@ func (cfg tlsServerCfg) Prepare() error {
 
 	if tlscfg.ClientAuth != tls.NoClientCert {
 		tlscfg.VerifyPeerCertificate = ReceptorVerifyFunc(tlscfg, pinnedFingerprints,
-			"", ExpectedHostnameTypeDNS, VerifyClient)
+			"", ExpectedHostnameTypeDNS, VerifyClient, logger.NewReceptorLogger(cfg.Name))
```
AaronH88: Im not sure here about using `name` as a replacement for `NodeID` alternatives are welcome!

pkg/logger/logger.go
```diff
@@ -16,10 +17,10 @@ var (
 
 // Log level constants.
 const (
-	ErrorLevel = iota + 1
-	WarningLevel
-	InfoLevel
-	DebugLevel
+	ERROR_LEVEL = iota + 1
```
AaronH88: The chosen linter does not like all caps CONST's 

Does the team prefer the original CamelCase? If so I will change it back.

If the team likes what I have done here I can look at modifying the linter, 
Let me know what you think :smile:

fosterseth: we've mostly stuck to camelcase, so might be best to stick to that

fosterseth:
does this append some netceptor info to the beginning of each line? can you post some examplar logs that are emitted with this change?

pkg/logger/logger.go
AaronH88: Thanks for the feedback, I have reverted this change :smile:

AaronH88:
> does this append some netceptor info to the beginning of each line? can you post some examplar logs that are emitted with this change?

Of course.
Yes, this logger will append a given message to the beginning of each log.
This `prefix` is set at instantiation, e.g. for netceptor we are suggesting adding the `NodeID` to the prefix of all logs `logger.NewReceptorLogger(nodeID)` 

The result of this is the node id will be printed for all logs (very useful in functional tests)
e.g. `rl.Log.Debug( "Hello World") ` will be printed as ` DEBUG NodeA Hello World `

pkg/netceptor/netceptor.go
```diff
@@ -338,6 +340,7 @@ func NewWithConsts(ctx context.Context, nodeID string,
 		serverTLSConfigs:         make(map[string]*tls.Config),
 		firewallLock:             &sync.RWMutex{},
 		workCommandsLock:         &sync.RWMutex{},
+		Logger:                   logger.NewReceptorLogger(nodeID),
```
shanemcd: I think we probably don't want to append the nodeID to our logs by default. It will definitely come in handy when we're refactoring the tests though.

shanemcd (approved):

PR #729, opened 2023-01-27
Do not use reconnect by default

fosterseth:
Our auto-detect is causing a lot of problems for folks. There can be a drift between the kube api server, which is what our code detects, and the underlying node kubelet.

see this issue https://github.com/ansible/awx/issues/13469

and specifically this comment

https://github.com/ansible/awx/issues/13469#issuecomment-1406796761

we should default to `false` (i.e. use legacy io.Copy) unless overridden by the `RECEPTOR_KUBE_SUPPORT_RECONNECT` variable

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -653,46 +653,10 @@ func (kw *kubeUnit) runWorkUsingLogger() {
 	kw.UpdateBasicStatus(WorkStateSucceeded, "Finished", stdout.Size())
 }
 
-func shouldUseReconnect(kw *kubeUnit) bool {
-	// Attempt to detect support for streaming from pod with timestamps based on
-	// Kubernetes server version
-	// In order to use reconnect method, Kubernetes server must be at least
-	//   v1.23.14
-	//   v1.24.8
-	//   v1.25.4
-	// These versions contain a critical patch that permits connecting to the
-	// logstream with timestamps enabled.
-	// Without the patch, stdout lines would be split after 4K characters into a
-	// new line, which will cause issues in Receptor.
-	// https://github.com/kubernetes/kubernetes/issues/77603
-	// Can override the detection by setting the RECEPTOR_KUBE_SUPPORT_RECONNECT
-	// accepted values: "enabled", "disabled", "auto" with "auto" being the default
-	// all invalid value will assume to be "auto"
-
-	env, ok := os.LookupEnv("RECEPTOR_KUBE_SUPPORT_RECONNECT")
-	if ok {
-		switch env {
-		case "enabled":
-			return true
-		case "disabled":
-			return false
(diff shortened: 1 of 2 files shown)


pkg/workceptor/kubernetes.go
```diff
@@ -690,39 +724,8 @@ func shouldUseReconnect(kw *kubeUnit) bool {
...
-	semver = semver.WithPreRelease("")
+	return isCompatibleK8S(kw, serverVerInfo.String())
 
-	// The patch was backported to minor version 23, 24 and 25. We must check z stream
-	// based on the minor version
-	// if minor version == 24, compare with v1.24.8
-	// if minor version == 25, compare with v1.25.4
-	// all other minor versions compare with v1.23.14
-	var compatibleVer string
-	switch serverVerInfo.Minor {
```
fosterseth, line 709: @TheRealHaoLiu this was the issue

we use `serverVerInfo` when grabbing the compatible version, but using `semver` when doing the version check below.

pkg/workceptor/kubernetes.go
```diff
@@ -690,39 +724,8 @@ func shouldUseReconnect(kw *kubeUnit) bool {
...
 
-	semver, err := version.ParseSemantic(serverVerInfo.String())
-	if err != nil {
-		logger.Warning("could parse Kubernetes server version %s, will not use reconnect support", serverVerInfo.String())
-
-		return false
-	}
-
-	// ignore pre-release in version comparison
-	semver = semver.WithPreRelease("")
```
fosterseth, line 701: I'm inclined to remove this -- seems too risky to ignore the pre-release tags. I know there was at least one k8s cluster out there that doesn't follow version semantics, so reconnect should be possible but auto detect says no.

that is fine, they can enable it through the ENV. However, incorrectly enabling reconnect support when in fact the k8s cluster doesn't support it is really bad. I'd rather just have them experience the 4 hour log problem in the meantime.

TheRealHaoLiu: Main problem is that the fix is not in k8s API server code and it's in kubelet code so we actually needed to read node version not k8s apiserver version 🤦‍♂️

pkg/workceptor/kubernetes_test.go
```diff
@@ -0,0 +1,55 @@
+package workceptor
```
TheRealHaoLiu, line 1: @fosterseth ++

pkg/workceptor/kubernetes.go
TheRealHaoLiu: and i think we should still put the pre-release stripping in

fosterseth: added back

TheRealHaoLiu (approved):

PR #726, opened 2023-01-23
Configurable k8s clientset rate limit and exponential backoff on retry

TheRealHaoLiu:
fix situation where rate limiting potentially cause job to fail 
```
2023-01-18T14:37:00.203383244Z DEBUG 2023/01/18 14:37:00 [1WlBA2rY] Detected EOF for pod aap-jobs/automation-job-203398-gmlvn. Will retry 5 more times.
2023-01-18T14:37:00.273976994Z DEBUG 2023/01/18 14:37:00 Client connected to control service @
2023-01-18T14:37:00.283886572Z DEBUG 2023/01/18 14:37:00 Control service closed
2023-01-18T14:37:00.283886572Z DEBUG 2023/01/18 14:37:00 Client disconnected from control service @
2023-01-18T14:37:00.304442237Z WARNING 2023/01/18 14:37:00 [1WlBA2rY] Error getting pod aap-jobs/automation-job-203398-gmlvn. Will retry 5 more times.
2023-01-18T14:37:01.304595010Z WARNING 2023/01/18 14:37:01 [1WlBA2rY] Error getting pod aap-jobs/automation-job-203398-gmlvn. Will retry 4 more times.
2023-01-18T14:37:02.305715415Z WARNING 2023/01/18 14:37:02 [1WlBA2rY] Error getting pod aap-jobs/automation-job-203398-gmlvn. Will retry 3 more times.
2023-01-18T14:37:03.306418208Z WARNING 2023/01/18 14:37:03 [1WlBA2rY] Error getting pod aap-jobs/automation-job-203398-gmlvn. Will retry 2 more times.
2023-01-18T14:37:03.601594470Z ERROR 2023/01/18 14:37:03 Exceeded retries for reading stdout /tmp/receptor/tower-f78967dbc-cbl8f/1WlBA2rY/stdout
2023-01-18T14:37:03.644009428Z WARNING 2023/01/18 14:37:03 Could not read in control service: read unix /var/run/receptor/receptor.sock->@: use of closed network connection
2023-01-18T14:37:03.644009428Z DEBUG 2023/01/18 14:37:03 Client disconnected from control service @
2023-01-18T14:37:03.644009428Z WARNING 2023/01/18 14:37:03 Could not close connection: close unix /var/run/receptor/receptor.sock->@: use of closed network connection
[...]

PR #732, opened 2023-01-31
Reduce k8s API request when running ansible-runner workloads

TheRealHaoLiu:
on top of https://github.com/ansible/receptor/pull/726 

add ansible-runner specific EOF detection logic

if receptor is running ansible-runner workload
when we receive log message end with `{"eof": true}` before we receive EOF error we know that we finish reading from the job pod and will not need to attempt to reconnect to the pod


TheRealHaoLiu:
test image quay.io/haoliu/awx-ee:retry-alt

PR #734, opened 2023-02-02
Configureable k8s rate limit

TheRealHaoLiu:
- no error shall escape

```
RECEPTOR_KUBE_CLIENTSET_QPS
- default: 100 
```
```
RECEPTOR_KUBE_CLIENTSET_BURST
- default: 10*QPS 
```
```
- default: tokenbucket
```

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"net"
 	"os"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


shanemcd (commented):
Let's include some mention of these new options somewhere on https://receptor.readthedocs.io/en/latest/k8s.html

fosterseth:
@TheRealHaoLiu only other places I saw where the err message was not logged
https://github.com/ansible/receptor/blob/f4e472821333ed2afd1a32bb72357b9fd1cb544f/pkg/workceptor/kubernetes.go#L928
and a handful in the `runWorkUsingTCP()`, which is not a method we use for anything in AAP

fosterseth (approved):

PR #735, opened 2023-02-02
Configurable k8s rate limit with exponential backoff on retries

TheRealHaoLiu:
- increase retry interval and add exponential backoff for retries

```
RECEPTOR_KUBE_CLIENTSET_QPS
- default: 100
```
```
RECEPTOR_KUBE_CLIENTSET_BURST
- default: 10*QPS
```
```
- default: tokenbucket
```

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"net"
 	"os"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -25,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/version"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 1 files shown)


pkg/workceptor/kubernetes.go
```diff
@@ -990,6 +1075,49 @@ func (kw *kubeUnit) connectToKube() error {
...
+		if err != nil {
+			return fmt.Errorf("invalid value for RECEPTOR_KUBE_CLIENTSET_BURST: %s. Error %s", envBurst, err)
+		}
+	}
+
+	kw.config.QPS = float32(qps)
+	kw.config.Burst = burst
+
+	// RECEPTOR_KUBE_CLIENTSET_RATE_LIMITER
+	// default: tokenbucket
```
obaranov, line 1108: Here https://pkg.go.dev/k8s.io/client-go/rest I can see :


```
	// Rate limiter for limiting connections to the master from this client. If present overwrites QPS/Burst
	RateLimiter flowcontrol.RateLimiter
```

Does that mean that we use default rate limiter and just ignore QPS and BURST values?

pkg/workceptor/kubernetes.go
```diff
@@ -990,6 +1075,49 @@ func (kw *kubeUnit) connectToKube() error {
...
+	if ok {
+		var err error
+		burst, err = strconv.Atoi(envBurst)
+		if err != nil {
+			return fmt.Errorf("invalid value for RECEPTOR_KUBE_CLIENTSET_BURST: %s. Error %s", envBurst, err)
+		}
+	}
+
+	kw.config.QPS = float32(qps)
+	kw.config.Burst = burst
```
obaranov, line 1105: should we convert burst here to integer?

PR #754, opened 2023-03-24
Moved connLock to fix race condition.

AaronH88:

Fixed race condition where 2 nodes with the same id could both establish connections at the same time

Keeping in draft as this change causes tests to fail. I will change the tests in a different PR


shanemcd (approved):
🍰

PR #757, opened 2023-03-27
Added more ciphers to CipherSuite. Refactored how all TLS config is created

AaronH88:

### DIFF
--- a/pkg/backends/tcp.go
+++ b/pkg/backends/tcp.go
@@ -39,6 +39,14 @@ func NewTCPDialer(address string, redial bool, tls *tls.Config, logger *logger.R
 	return &td, nil
 }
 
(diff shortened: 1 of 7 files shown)


pkg/backends/tcp.go
```diff
@@ -227,8 +247,9 @@ func (ns *TCPSession) Close() error {
 // Command line
 // **************************************************************************
 
-// tcpListenerCfg is the cmdline configuration object for a TCP listener.
-type tcpListenerCfg struct {
+//TODO make these fields private
```
AaronH88, line 250: Leaving this as TODO for now as this PR is already 200+ lines of code.

Will make a separate PR refactoring the scope of our fields

shanemcd (approved):
🚀

PR #780, opened 2023-05-09
Respect ctx cancel

TheRealHaoLiu:
I think there's a race condition between controller and receptor

controller see the job finished and try to cancel workunit

receptor does not respect the cancel and continue to do GET to kube apiserver with a dead ctx which cause a very misleading error message
```
client rate limiter Wait returned an error: context canceled
```


AaronH88 (approved):
lgtm

PR #785, opened 2023-05-26
Lay groundwork for mocking and add initial workceptor unit tests

AaronH88:
Addresses issue #796


pkg/workceptor/workceptor.go
```diff
@@ -16,17 +17,29 @@ import (
...
 	"github.com/ansible/receptor/pkg/netceptor"
 	"github.com/ansible/receptor/pkg/randstr"
 	"github.com/ansible/receptor/pkg/utils"
 	"github.com/golang-jwt/jwt/v4"
 )
 
+// Netceptor is a interface to decouple workceptor from netceptor.
+// it includes only the functions that netceptor uses.
+// the interface name is work in progress.
+type Netceptor interface {
```
AaronH88: The golang naming convention for interfaces is to use an `er` postfix, and while this works in most cases `Netceptorer` is difficult to read. 
I originally had the interface name as a name pairing, outlining the interface relationship they represent, however `WorkceptorsNetceptor` while intuitive is overly verbose. 
Other languages have the `i` convention, and while `INetceptor` would be easy to understand, it breaks the golang naming convention and is overly verbose, with `i` and `interface` both being written on the same line.

@shanemcd had the suggestion of using just `Netceptor`. It is the cleanest to look at however, not the most obvious to understand the relationship. The reader needs to understand that in order to call the concrete netceptor then namespace syntax needs to be used, e.g. `netceptor.Netceptor`, and also needs to understand the current package, and that this interface is implemented in the `netceptor` package.

I am leaning toward following Shanes suggestion, and hoping that any readers of the code base will pick up that the `interface` keyword is on this line and understand its special meaning.

pkg/workceptor/workceptor.go
```diff
@@ -16,17 +17,28 @@ import (
...
+	"github.com/ansible/receptor/pkg/logger"
 	"github.com/ansible/receptor/pkg/netceptor"
 	"github.com/ansible/receptor/pkg/randstr"
 	"github.com/ansible/receptor/pkg/utils"
 	"github.com/golang-jwt/jwt/v4"
 )
 
+// NetceptorForWorkceptor is a interface to decouple workceptor from netceptor.
+// it includes only the functions that workceptor uses.
+type NetceptorForWorkceptor interface {
```
AaronH88, line 29: One final suggestion from Shane seems to be the winner, `NetceptorForWorkceptor` will be the new standard we set.

PR #786, opened 2023-06-02
Close go routines when context is closed

AaronH88:
This PR addresses potential thread leaking in receptor.

### DIFF
--- a/pkg/controlsvc/controlsvc.go
+++ b/pkg/controlsvc/controlsvc.go
@@ -236,7 +236,7 @@ func (s *Server) RunControlSession(conn net.Conn) {
 				cc, err = ct.InitFromJSON(jsonData)
 			}
 			if err == nil {
-				ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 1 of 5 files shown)


jjwatt (approved):

pkg/controlsvc/controlsvc.go
```diff
@@ -356,6 +356,12 @@ func (s *Server) RunControlSvc(ctx context.Context, service string, tlscfg *tls.
 					if ctx.Err() != nil {
 						return
 					}
+					select {
+					case <-ctx.Done():
```
fosterseth, line 357: ctx.Err() is non-nil when context is closed. does that mean we are covered by the lines just above this select block?

AaronH88: You're correct, the logic will remain the same. Good spot

After looking at it some more I can see online that typically `ctx.Err()` is used when we want to return information on why the context was cancelled. 
In this case, nothing is returned, so I'm wondering if I should replace the `Err()` with the `Done()`, 
Any thoughts?

PR #805, opened 2023-07-17
Add sonarqube workflow for PRs

AaronH88:
Target the devel branch, 
Each new PR to the devel branch will trigger a sonarqube scan.
Sonarqube server being used is a redhat corporate instance and not publicly available.


AaronH88:
Will pivot to sonarcloud

PR #810, opened 2023-07-27
Unit tests for Ping.go inside netceptor 

matoval:
Create interface for Ping & Packetconn and add unit tests to cover netceptor/ping.go

Increase ping.go coverage from 0% to 92.5


pkg/netceptor/packetconn.go
```diff
@@ -7,9 +7,25 @@ import (
...
+	SetHopsToLive(hopsToLive byte)
+	SubscribeUnreachable(doneChan chan struct{}) chan UnreachableNotification
+	ReadFrom(p []byte) (int, net.Addr, error)
+	WriteTo(p []byte, addr net.Addr) (n int, err error)
+	LocalAddr() net.Addr
+	Close() error
+	SetDeadline(t time.Time) error
+	SetReadDeadline(t time.Time) error
+	SetWriteDeadline(t time.Time) error
+	GetCancel() *context.CancelFunc
```
AaronH88: As this is a pointer to the cancel function, and when called the function is immediately evoked, I suggest changing this name to just `Cancel()`

pkg/netceptor/packetconn.go
```diff
@@ -7,9 +7,25 @@ import (
...
+	GetLocalService() string
+	GetNetceptorLogger() *logger.ReceptorLogger
```
AaronH88: The logger should be one that is re referenced throughout the application, i suggest changing this name to `GetLogger()` or just `Logger()` as it has the same meaning

pkg/netceptor/packetconn.go
```diff
@@ -25,9 +41,30 @@ type PacketConn struct {
 	cancel          context.CancelFunc
 }
 
+func NewPacketConnWithConst(s *Netceptor, service string, advertise bool, adtags map[string]string, connTypeDatagram byte) (*PacketConn, error) {
```
AaronH88: We have an opportunity here to "Accept interface". 
We can replace this `Netceptor` struct with an `interface` representing all the calls to `netceptor`. This will make future testing easier.

pkg/netceptor/packetconn.go
```diff
@@ -7,9 +7,25 @@ import (
...
+	SetReadDeadline(t time.Time) error
+	SetWriteDeadline(t time.Time) error
+	GetCancel() *context.CancelFunc
+	GetLocalService() string
+	GetNetceptorLogger() *logger.ReceptorLogger
+}
+
 // PacketConn implements the net.PacketConn interface via the Receptor network.
 type PacketConn struct {
 	s               *Netceptor
```
AaronH88, line 31: We can replace this `Netceptor` with an `interface` to make future testing easier.

pkg/netceptor/packetconn.go
AaronH88: Im also not a fan of the function name and am open to suggestions

pkg/netceptor/packetconn.go
```diff
@@ -25,9 +41,30 @@ type PacketConn struct {
...
+		hopsToLive:   s.maxForwardingHops,
+	}
+
+	npc.startUnreachable()
+	s.listenerRegistry[service] = npc
+
+	return npc, nil
+}
+
+func NewPacketConn(s *Netceptor, service string, connTypeDatagram byte) (*PacketConn, error) {
```
AaronH88: Same comment for `Netceptor` as above

pkg/netceptor/packetconn.go
AaronH88: The error in this function it pointless, it is never set to anything other than `nil`

pkg/netceptor/packetconn.go
```diff
@@ -42,35 +79,38 @@ func (s *Netceptor) ListenPacket(service string) (*PacketConn, error) {
...
-	s.listenerRegistry[service] = pc
 
 	return pc, nil
 }
 
 // ListenPacketAndAdvertise returns a datagram listener, and also broadcasts service
 // advertisements to the Receptor network as long as the listener remains open.
-func (s *Netceptor) ListenPacketAndAdvertise(service string, tags map[string]string) (*PacketConn, error) {
-	pc, err := s.ListenPacket(service)
+func (s *Netceptor) ListenPacketAndAdvertise(service string, tags map[string]string) (PacketConner, error) {
```
AaronH88, line 89: It seems `ListenPacket` does a lot more error checking than this function, why? 

As the checks not valid between the two functions or were they missed?

pkg/netceptor/ping.go
```diff
@@ -7,8 +7,21 @@ import (
...
+type NetceptorForPing interface {
+	ListenPacket(service string) (PacketConner, error)
+	NewAddr(target string, service string) Addr
+	NodeID() string
+	Context() context.Context
+}
+
 // Ping sends a single test packet and waits for a reply or error.
 func (s *Netceptor) Ping(ctx context.Context, target string, hopsToLive byte) (time.Duration, string, error) {
+	return Pinger(s, ctx, target, hopsToLive)
```
AaronH88: The `er` postfix is used for interfaces in golang, we will need to change this function name to avoid confusion.

pkg/netceptor/ping.go
```diff
@@ -7,8 +7,21 @@ import (
...
+}
+
+// Ping sends a single test packet and waits for a reply or error.
+func Pinger(s NetceptorForPing, ctx context.Context, target string, hopsToLive byte) (time.Duration, string, error) {
```
AaronH88: Typically `context` is the first argument in go, lets rearrange the arguments list.

pkg/netceptor/ping.go
```diff
@@ -7,8 +7,21 @@ import (
...
+	NodeID() string
+	Context() context.Context
+}
+
 // Ping sends a single test packet and waits for a reply or error.
 func (s *Netceptor) Ping(ctx context.Context, target string, hopsToLive byte) (time.Duration, string, error) {
+	return Pinger(s, ctx, target, hopsToLive)
+}
+
+// Ping sends a single test packet and waits for a reply or error.
```
AaronH88: Comment needs updating

AaronH88:
Overall looking great, a few small things to think about and make the future work easier.

pkg/netceptor/packetconn.go
matoval: We will create another PR to add the Netceptor interface as it will increase the complexity and focus of this PR.

pkg/netceptor/packetconn.go
matoval: We will create another PR to add the Netceptor interface as it will increase the complexity and focus of this PR.

pkg/netceptor/packetconn.go
matoval: We will create another PR to add the Netceptor interface as it will increase the complexity and focus of this PR.

pkg/netceptor/packetconn.go
```diff
@@ -42,35 +79,46 @@ func (s *Netceptor) ListenPacket(service string) (*PacketConn, error) {
...
-	pc.adTags = tags
+	if service == "" {
+		service = s.getEphemeralService()
+	}
+	s.listenerLock.Lock()
+	defer s.listenerLock.Unlock()
+	_, isReserved := s.reservedServices[service]
+	_, isListening := s.listenerRegistry[service]
+	if isReserved || isListening {
+		return nil, fmt.Errorf("service %s is already listening", service)
```
AaronH88: Lets make this error a little more verbose so we can tell the difference between here and line 79.

Im open to ideas, here is my suggestion `""service %s is already listening and advertising", service"`

pkg/netceptor/ping.go
```diff
@@ -7,8 +7,21 @@ import (
...
+	Context() context.Context
+}
+
+// Ping calls CreatePing to sends a single test packet and waits for a reply or error.
 func (s *Netceptor) Ping(ctx context.Context, target string, hopsToLive byte) (time.Duration, string, error) {
+	return CreatePing(ctx, s, target, hopsToLive)
+}
+
+// CreatePing creates Ping by sending a single test packet and waits for a replay or error.
+func CreatePing(ctx context.Context, s NetcForPing, target string, hopsToLive byte) (time.Duration, string, error) {
```
AaronH88: `CreatePing` suggests to me that a `ping` struct would be created. However this function seems to send a ping request, I suggest changing the name to something like `SendPing` or `DoPing` or trying to turn it into a verb of some kind

pkg/netceptor/ping.go
```diff
@@ -75,20 +88,30 @@ func (s *Netceptor) Ping(ctx context.Context, target string, hopsToLive byte) (t
...
 	Time time.Duration
 	Err  error
 }
 
-// Traceroute returns a channel which will receive a series of hops between this node and the target.
 func (s *Netceptor) Traceroute(ctx context.Context, target string) <-chan *TracerouteResult {
+	return CreateTraceroute(ctx, s, target)
+}
+
+// Traceroute returns a channel which will receive a series of hops between this node and the target.
```
AaronH88: Please update the comments :) 

I like the names you picked here as a `Traceroute` `chan` is returned, so it makes sense

pkg/netceptor/ping_test.go
```diff
@@ -0,0 +1,251 @@
...
+
+// setupTest sets up TestPing tests.
+func setupTest(t *testing.T) (*gomock.Controller, *mock_netceptor.MockNetcForPing, *mock_netceptor.MockPacketConner) {
+	ctrl := gomock.NewController(t)
+
+	// Prepare mocks
+	mockNetceptor := mock_netceptor.NewMockNetcForPing(ctrl)
+	mockPacketConn := mock_netceptor.NewMockPacketConner(ctrl)
+
+	mockPacketConn.EXPECT().SetHopsToLive(gomock.Any()).AnyTimes()
```
AaronH88: Interesting that you have to set the times to `Any` and `0` in the set ups and tear downs, any reason for this?

pkg/netceptor/ping_test.go
```diff
@@ -0,0 +1,251 @@
...
+		{"WriteTo Error", "target", byte(1), listenPacketReturn{mockPacketConn, nil, 1, "return"}, writeToReturn{0, errors.New("WriteTo error"), 1, "return"}, contextReturn{context.Background(), 2, "return"}, readFromReturn{0, nil, nil, 1, "return"}, 0, "", errors.New("WriteTo error")},
+		{"Timeout Error", "target", byte(1), listenPacketReturn{mockPacketConn, nil, 1, "return"}, writeToReturn{0, nil, 1, "return"}, contextReturn{context.Background(), 2, "return"}, readFromReturn{0, nil, nil, 1, "do"}, 0, "", errors.New("timeout")},
+		{"User Cancel Error", "target", byte(1), listenPacketReturn{mockPacketConn, nil, 1, "return"}, writeToReturn{0, nil, 1, "return"}, contextReturn{context.Background(), 2, "return"}, readFromReturn{0, nil, nil, 1, "doAndReturn"}, 0, "", errors.New("user cancelled")},
+	}
+
+	for _, testCase := range pingTestCases {
+		ctx := context.Background()
+		t.Run(testCase.name, func(t *testing.T) {
+			// Set up the mock behaviours
+			if testCase.name != "ListenSubscribeUnreachable Error" {
```
AaronH88: I see why you have branched so many times here, good job for working through this logic. 

Perhaps we could look at alternative ways of doing this? 
First thing that comes to mind is function pointers, perhaps `pingTestCases` could have two function pointer fields, `setUp` and `checkResults`. 
This way each test case could have a unique set up and results check without this if conditions being needed. WDYT?

pkg/netceptor/ping_test.go
```diff
@@ -0,0 +1,253 @@
...
+		"NewAddrReturn": func() {
+			mockNetceptor.EXPECT().NewAddr(gomock.Any(), gomock.Any()).Return(netceptor.Addr{}).Times(testCase.returnNewAddrTimes)
+		},
+		"NodeID":         func() { mockNetceptor.EXPECT().NodeID().Return("nodeID") },
+		"CreateChannel":  func() { createChannel(mockPacketConn) },
+		"SleepOneSecond": func() { time.Sleep(time.Second * 1) },
+	}
+
+	for _, expect := range testCase.expects {
+		testExpects[expect]()
```
AaronH88, line 115: Very nice

AaronH88 (approved):
Great work! lgtm

PR #818, opened 2023-08-15
Expand retry conditions for K8 logs

### LINKED ISSUE #14293: AWX jobs can't tolerate the EKS scale out
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/awx/issues) for duplicates.
- [X] I understand that AWX is open source software provided for free and that I might not receive a timely response.
- [X] I am **NOT** reporting a (potential) security vulnerability. (These should be emailed to `security@ansible.com` instead.)

### Bug Summary

We have an environment for performance tests where we try to run more than 400 heavy workflows simultaneously. It causes the EKS control plane (master nodes) to scale out, and about half of our pods fail with errors. We've already opened an issue about it when we were reproducing the K8s master node termination in the kOps cluster (since we cannot do it with EKS) and it was fixed for the kOps clusters [here](https://github.com/ansible/awx/issues/13350), thanks to the community and especially to @TheRealHaoLiu. We also validated it with AWS support, asked them to terminate K8s nodes manually, and validated that the issue was resolved. The problem is that it is not fully applied to the scale-in or scale-out process of the EKS control plane.

In our case, it fails for EKS when the control plane scales out. For now, we are running the **control-plane-ee** with the following configuration:
```
  ee_extra_env: |-
      value: enabled
    - name: RECEPTOR_LOG_LEVEL
      value: debug
```

After the [bugfix](https://github.com/ansible/receptor/pull/760) is merged, some pods can tolerate it and finish running successfully when the scale-out happens, and it can be considered an intermediate result. I see the following logs related to them (the second error happens occasionally and doesn't affect the AWX functionality)

**automation-job-141264-fxwvc (finished successfully)**

```
INFO 2023/07/27 20:17:38 [OqRdMEsY] Detected http2.GoAwayError for pod awx-workers/automation-job-141264-fxwvc. Will retry 5 more times. Error: http2: server sent GOAWAY and closed the connection; LastStreamID=6491, ErrCode=NO_ERROR, debug=""
ERROR 2023/07/27 20:26:47 [OqRdMEsY] Error reading from pod awx-workers/automation-job-141264-fxwvc: context canceled
```

But there are still many pods that fail with error. Here are variations of logs that we can see (we show all logs in **control-plane-ee** that are related to the particular job)

**automation-job-143331-7lggn (failed)**
`ERROR 2023/07/27 20:19:56 [JwdC5LdB] Error reading from pod awx-workers/automation-job-143331-7lggn: unexpected EOF`

**automation-job-140003-gq82x (failed)**
```
INFO 2023/07/27 20:17:04 [xJDCo0r6] Detected http2.GoAwayError for pod awx-workers/automation-job-140003-gq82x. Will retry 5 more times. Error: http2: server sent GOAWAY and closed the connection; LastStreamID=3639, ErrCode=NO_ERROR, debug=""
ERROR 2023/07/27 20:19:56 [xJDCo0r6] Error reading from pod awx-workers/automation-job-140003-gq82x: unexpected EOF
```

**automation-job-142724-mhdcm (failed)**
`ERROR 2023/07/27 20:17:04 [7Ewis9Pm] Error reading from pod awx-workers/automation-job-142724-mhdcm: read tcp 100.64.85.69:40134->172.27.2.235:443: read: connection reset by peer`

We still have our performance environment up and running and will gladly provide more logs if needed. Please tell us if changing the DEBUG log level to TRACE makes sense and can give us more info about the issue.

### AWX version

22.5.0

### Select the relevant components

- [ ] UI
- [ ] UI (tech preview)
- [X] API
- [ ] Docs
- [X] Collection
- [ ] CLI
- [ ] Other

### Installation method

kubernetes

### Modifications

no

### Ansible version

_No response_

### Operating system

_No response_

### Web browser

_No response_

### Steps to reproduce

Run more than 400 heavy workflows in EKS cluster that will cause the EKS control plane scale out

### Expected results

Jobs can tolerate the EKS control plane scale out

### Actual results

Half of the jobs fail with error

### Additional information

The EKS cluster version is 1.24

AaronH88:
Fixes: https://github.com/ansible/awx/issues/14293 
Expanded retry conditions for K8 logs to retry at least 5 times regardless of the error returned. This is due to K8 api timings where the GOAWAY error may not be sent or sent already before log streams get attached.

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -18,7 +18,6 @@ import (
 
 	"github.com/ghjm/cmdline"
 	"github.com/google/shlex"
-	"golang.org/x/net/http2"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -239,43 +238,22 @@ func (kw *kubeUnit) kubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout 
 		streamReader := bufio.NewReader(logStream)
 		for *stdinErr == nil { // check between every line read to see if we need to stop reading
 			line, err := streamReader.ReadString('\n')
-			if err == io.EOF {
-				kw.Debug(
-					"Detected EOF for pod %s/%s. Will retry %d more times. Error: %s",
-					podNamespace,
-					podName,
-					remainingRetries,
-					err,
-				)
-				successfulWrite = false
-				remainingRetries--
-				if remainingRetries > 0 {
-					time.Sleep(200 * time.Millisecond)
-
-					break
-				}
-
-				return
-			} else if _, ok := err.(http2.GoAwayError); ok {
-				// GOAWAY is sent by the server to indicate that the server is gracefully shutting down
-				// this happens if the kube API server we are connected to is being restarted or is shutting down
-				// for example during a cluster upgrade and rolling restart of the master node
-				kw.Info(
-					"Detected http2.GoAwayError for pod %s/%s. Will retry %d more times. Error: %s",
-					podNamespace,
-					podName,
-					remainingRetries,
-					err,
-				)
-				successfulWrite = false
-				remainingRetries--
-				if remainingRetries > 0 {
-					time.Sleep(200 * time.Millisecond)
-
-					break
-				}
-			}
 			if err != nil {
+				kw.Info(
+					"Detected Error: %s for pod %s/%s. Will retry %d more times.",
+					err,
+					podNamespace,
+					podName,
+					remainingRetries,
+				)
+
+				successfulWrite = false
+				remainingRetries--
+				if remainingRetries > 0 {
+					time.Sleep(200 * time.Millisecond)
+
+					break
+				}
 				*stdoutErr = err
 				kw.Error("Error reading from pod %s/%s: %s", podNamespace, podName, err)
 


erz4 (approved):

elibramson (approved):

pkg/workceptor/kubernetes.go
```diff
@@ -256,17 +257,18 @@ func (kw *kubeUnit) kubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
 				}
 
 				return
-			} else if _, ok := err.(http2.GoAwayError); ok {
+			} else if err != nil || goaway {
```
TheRealHaoLiu: wouldn't `err != nil` cover `goaway` also? Any reason why `|| goaway` is needed?

AaronH88: Good point, I was playing with printing a specific `goaway` error, but never fully cleaned up once I was done. Ill change this now

pkg/workceptor/kubernetes.go
```diff
@@ -256,17 +255,15 @@ func (kw *kubeUnit) kubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
 				}
 
 				return
-			} else if _, ok := err.(http2.GoAwayError); ok {
-				// GOAWAY is sent by the server to indicate that the server is gracefully shutting down
-				// this happens if the kube API server we are connected to is being restarted or is shutting down
-				// for example during a cluster upgrade and rolling restart of the master node
+			} else if err != nil {
```
TheRealHaoLiu: now that we expend the err handling i think the `if err == io.EOF` can also be folded in to `err != nil` and this logic can be restructured a bit

pkg/workceptor/kubernetes.go
AaronH88: removed

pkg/workceptor/kubernetes.go
AaronH88: Thanks for this! All folded into one `err != nil`

pkg/workceptor/kubernetes.go
```diff
@@ -275,7 +255,7 @@ func (kw *kubeUnit) kubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
 					break
 				}
 			}
-			if err != nil {
+			if err != nil && remainingRetries > 0 {
```
TheRealHaoLiu: I don't think the additional conditional `&& remainingRetries > 0` is needed here

I think the only way that the code reach this point is if we exhausted the retries?

TheRealHaoLiu: @fosterseth am i right?

fosterseth: agreed. I don't think this conditional will ever hit

AaronH88: Removed, i think Seth is right, now with the changes this condition will never be hit, the first `err != nil` will be triggered and this will never be called. 

I have removed the whole check.

pkg/workceptor/kubernetes.go
```diff
@@ -275,12 +255,6 @@ func (kw *kubeUnit) kubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
 					break
 				}
 			}
-			if err != nil {
-				*stdoutErr = err
-				kw.Error("Error reading from pod %s/%s: %s", podNamespace, podName, err)
-
-				return
```
TheRealHaoLiu: these code will still need to be executed when the retries is exhausted

TheRealHaoLiu (approved):

PR #819, opened 2023-08-15
Add unit tests for framer.go

resoluteCoder:


pkg/framer/framer_test.go
```diff
@@ -41,3 +41,69 @@ func TestSendData(t *testing.T) {
...
+		if testCase.expectedResult != receivedResult {
+			t.Errorf("%s - expected: %+v, received: %+v", testCase.name, testCase.expectedResult, receivedResult)
+		}
+	}
+}
+
+func TestGetMessage(t *testing.T) {
+	buffer := []byte{1, 2, 3}
+	framedBuffer := append([]byte{3, 0}, buffer...)
+	t.Run("Message read", func(t *testing.T) {
```
AaronH88: Good change here, Can you use `t.Run()` throughout the tests please. 

Can you also change these individual t.Runs into a table like above.

pkg/framer/framer_test.go
```diff
@@ -41,3 +41,69 @@ func TestSendData(t *testing.T) {
...
+		{
+			name:           "message not ready",
+			inputBuffer:    buffer,
+			expectedResult: false,
+		},
+	}
+
+	for _, testCase := range messageReadyTestCases {
+		f := framer.New()
+		f.RecvData(testCase.inputBuffer)
```
AaronH88: Lets wrap this code block in a `t.Run()`

AaronH88: Same on line 37, lets wrap this in a `t.Run()` block

AaronH88:
In the PR message `the rest` is very unclear, we need to understand via the commit history whats changed. Can I suggest something like `Add unit tests for framer.go` . 

This way your commit history will be the one from yesterday that called out the `sendData()` function and now this.

AaronH88 (approved):

PR #820, opened 2023-08-15
Test the randstr.go RandomString function

resoluteCoder:


pkg/randstr/randstr_test.go
```diff
@@ -0,0 +1,15 @@
+package randstr_test
+
+import (
+	"testing"
+
+	"github.com/ansible/receptor/pkg/randstr"
+)
+
+func TestRandStr(t *testing.T) {
+	randomStr := randstr.RandomString(10)
```
AaronH88, line 54: Lets wrap this in a `t.Run()` and lets add edge cases to this please.

What happens when we pass in 0, -1, and 100. Also are the return values truely random, lets check to make sure the return string is never the seed `"0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"` or a sub section of the seed. e.g. if 10 is passed then the return is never `0123456789`

pkg/randstr/randstr_test.go
```diff
@@ -0,0 +1,71 @@
...
+			expectedLength: 0,
+		},
+		{
+			name:           "length of -1",
+			inputLength:    -1,
+			expectedLength: 0,
+		},
+	}
+
+	t.Run("Test the random string length", func(t *testing.T) {
```
AaronH88: Please move this `t.Run` inside the for loop and use the test case name to create 3 distinct tests. `t.Run(testCase.name, func(t *testing.T) {`

resoluteCoder: ooops :smile:

pkg/randstr/randstr_test.go
```diff
@@ -0,0 +1,71 @@
...
+		},
+		{
+			name:           "length of -1",
+			inputLength:    -1,
+			expectedLength: 0,
+		},
+	}
+
+	for _, testCase := range randStringTestCases {
+		t.Run("Test the random string length", func(t *testing.T) {
```
AaronH88: Lets change `"Test the random string length"` to `testCase.name` please.

This will give us a more verbose message when we run the tests

pkg/randstr/randstr_test.go
```diff
@@ -0,0 +1,71 @@
...
+	for index, char := range randomStr {
+		if index == 0 {
+			continue
+		}
+		currentCharsetIndex := strings.Index(charset, string(char))
+		if charsetIndex+1 != currentCharsetIndex {
+			break
+		}
+		if index+1 == len(randomStr) {
+			t.Error("rand str is continuous")
```
AaronH88, line 67: Very nice, good work

AaronH88 (approved):

PR #824, opened 2023-08-21
Removed typecheck from the linting suite

AaronH88:
"`typecheck` is not a real linter, it's just a way to parse compiling errors (produced by the types.Checker) and some linter errors." 
As it is becoming flakey we will relay on the compiler.


resoluteCoder (approved):
lgtm

PR #827, opened 2023-08-22
Add unit tests for Controlsvc

resoluteCoder:


pkg/controlsvc/controlsvc.go
```diff
@@ -22,51 +24,117 @@ import (
...
+
+func (t *Tls) NewListener(inner net.Listener, config *tls.Config) net.Listener {
+	return tls.NewListener(inner, config)
+}
+
+// SockControl implements the ControlFuncOperations interface that is passed back to control functions.
+type SockControl struct {
+	conn  net.Conn
+	utils Utiler
+	io    Copier
```
AaronH88: So utils and copier are a field in Sock control. 
An alternative would be that `Utiler` and `Copier` are arguments for the functions that call them, have you considered this alternative?

pkg/controlsvc/controlsvc.go
```diff
@@ -22,51 +24,117 @@ import (
...
 		if err != nil {
 			return err
 		}
 	}
-	utils.BridgeConns(s.conn, "control service", bc, bcName, logger)
+	return nil
+}
+
+// BridgeConn bridges the socket to another socket.
+func (s *SockControl) BridgeConn(message string, bc io.ReadWriteCloser, bcName string, logger *logger.ReceptorLogger) error {
```
AaronH88: As part of the same comment for the fields in SockControl, and alternative would be that `BridgeConn` takes in a `utilier` argument

pkg/controlsvc/controlsvc.go
```diff
@@ -78,23 +146,30 @@ func (s *sockControl) WriteToConn(message string, in chan []byte) error {
...
 }
 
 // Server is an instance of a control service.
 type Server struct {
-	nc              *netceptor.Netceptor
+	nc              NetceptorForControlsvc
 	controlFuncLock sync.RWMutex
 	controlTypes    map[string]ControlCommandType
+	// new stuff
+	serverUtils Utiler
```
AaronH88: Same comment as the `SocketControl` struct

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+				})
+				mockNetceptor.EXPECT().GetLogger().Return(logger)
+			},
+		},
+	}
+
+	for _, testCase := range connectionListenerTestCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			ctx, ctxCancel := context.WithCancel(context.Background())
+			defer ctxCancel()
```
AaronH88, line 54: Your calling `defer ctxCancel()` and then explicitly calling `ctxCancel()` in both cases, this dosnet seem needed. Consider removing the explicit calls and just keeping the `defer` one.

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+
+			if testCase.expectedCalls != nil {
```
AaronH88: I personally prefer a redundant `expectedCalls()` rather than an `if` condition. Consider using `expectedCalls: func(ctxCancel context.CancelFunc) {}` in the first test case

AaronH88: if you fill in this blank `expectedCalls` then you can get rid of the `expectedError` field

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+	t.Errorf("expected error %s", err)
+}
+
+func TestConnectionListener(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	mockNetceptor := mock_controlsvc.NewMockNetceptorForControlsvc(ctrl)
+	mockListener := mock_controlsvc.NewMockListener(ctrl)
+	logger := logger.NewReceptorLogger("")
+
+	connectionListenerTestCases := []struct {
```
AaronH88, line 32: I dont see a positive case here, was it too tricky to get to due to the following call?

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+
+func TestSetupConnection(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	mockNetceptor := mock_controlsvc.NewMockNetceptorForControlsvc(ctrl)
+	mockConn := mock_controlsvc.NewMockConn(ctrl)
+	logger := logger.NewReceptorLogger("")
+
+	setupConnectionTestCases := []struct {
+		name          string
+		expectedError bool
```
AaronH88: this field is not needed

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+				"service":    "",
+				"unixSocket": "",
+				"tcpListen":  "",
+			},
+		},
+	}
+
+	for _, testCase := range runControlSvcTestCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			if testCase.expectedCalls != nil {
```
AaronH88: Same as the other test, a blank `ExpectedCalls` here?

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+			message: "message",
+			expectedCalls: func() {
+				mockCon.EXPECT().Write(gomock.Any()).Return(0, errors.New("cannot write message"))
+			},
+		},
+	}
+
+	for _, testCase := range writeMessageTestCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			if testCase.expectedCalls != nil {
```
AaronH88: same here

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+				testCase.expectedCalls()
+			}
+			err := sockControl.WriteMessage(testCase.message)
+
+			if testCase.message == "" && err != nil {
+				t.Errorf("should be nil")
```
AaronH88: More explicit error here please

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+			}
+			if testCase.message != "" && err.Error() != "cannot write message" {
+				t.Errorf("%s %s", testCase.name, err)
```
AaronH88: The testCase.name is already printed as part of the `t.Run` command above, please rewrite this error :smile:

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+	for _, testCase := range bridgeConnTestCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			testCase.expectedCalls()
+			err := sockControl.BridgeConn(testCase.message, mockCon, "test", logger)
+
+			if testCase.message == "" && err != nil {
+				t.Errorf("should be nil")
+			}
+			if testCase.message != "" && err.Error() != "terminated" {
+				t.Errorf("stuff %v", err)
```
AaronH88: Better error messages in both these tests please :smile:

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+
+func TestSockControlWriteMessage(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	mockCon := mock_controlsvc.NewMockConn(ctrl)
+	mockUtil := mock_controlsvc.NewMockUtiler(ctrl)
+	mockCopier := mock_controlsvc.NewMockCopier(ctrl)
+
+	sockControl := controlsvc.NewSockControl(mockCon, mockUtil, mockCopier)
+
+	writeMessageTestCases := []struct {
```
AaronH88, line 202: And a positive case?

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+			errorMessage:  "",
+		},
+	}
+
+	for _, testCase := range bridgeConnTestCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			testCase.expectedCalls()
+			err := sockControl.ReadFromConn(testCase.message, mockCon)
+
+			if testCase.expectedError {
```
AaronH88: Im having trouble understanding everything in this `if` block. 

It looks like you only need 1 `if` statement: 
```
if err != nil && err.Error() != testCase.errorMessage {
					printExpectedError(t, err)
				}
```

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+			testCase.expectedCalls()
+			c := make(chan []byte)
+			go func(c chan []byte) {
+				c <- []byte{7}
+				defer close(c)
+			}(c)
+
+			err := sockControl.WriteToConn(testCase.message, c)
+
+			if testCase.expectedError {
```
AaronH88: Same as the previous test, i dont understand this `if` block.

How can `err == nill` AND `err.Error()` exist?

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+}
+
+func TestAddControlFunc(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	mockCtrlCmd := mock_controlsvc.NewMockControlCommandType(ctrl)
+	mockNetceptor := mock_controlsvc.NewMockNetceptorForControlsvc(ctrl)
+	controlFuncTestsCases := []struct {
+		name          string
+		input         string
+		expectedError bool
```
AaronH88: This field isnt needed

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+		errorMessage  string
+		testCase      func(msg string, err error)
+	}{
+		{
+			name:          "ping command",
+			input:         "ping",
+			expectedError: true,
+			errorMessage:  "control function named ping already exists",
```
AaronH88: interesting, why does `ping` already exist? is this something left over from a previous test?

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+
+	mandatoryExpects := func() {
+		mockNetceptor.EXPECT().GetLogger().Return(logger).Times(3)
+		mockCon.EXPECT().RemoteAddr().Return(mockAddr).Times(2)
+		mockAddr.EXPECT().String().Times(2)
+		mockNetceptor.EXPECT().NodeID()
+	}
+
+	runControlSessionTestCases := []struct {
+		name          string
```
AaronH88, line 487: Looks like you only really use 2 fields here

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+
+func TestRunControlSessionTwo(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	mockCon := mock_controlsvc.NewMockConn(ctrl)
+	mockNetceptor := mock_controlsvc.NewMockNetceptorForControlsvc(ctrl)
+	logger := logger.NewReceptorLogger("")
+
+	runControlSessionTestCases := []struct {
+		name          string
+		message       string
```
AaronH88: looks like all these fields arent needed

pkg/controlsvc/controlsvc_test.go
```diff
@@ -0,0 +1,647 @@
...
+		expectedCalls func()
+		expectedError bool
+		errorMessage  string
+		commandByte   []byte
+	}{
+		{
+			name: "command must be a string",
+			expectedCalls: func() {
+				mockNetceptor.EXPECT().NodeID()
+				mockCon.EXPECT().Write(gomock.Any()).Return(0, nil).AnyTimes() // don't know why
```
AaronH88: Love this comment, lets investigate why :smile:

AaronH88:
Overall very good! Few small comments and one structural question, but this is great work!

pkg/controlsvc/controlsvc_test.go
resoluteCoder: I had it that way the first time but code smells wanted a comment on each empty function call. 

I can change it back with adding a meaning less comment to pass code smells.

Makefile
```diff
@@ -68,7 +68,7 @@ build-all:
 	GOOS=windows go build -o receptor.exe ./cmd/receptor-cl && \
 	GOOS=darwin go build -o receptor.app ./cmd/receptor-cl && \
 	go build example/*.go && \
-	go build -o receptor --tags no_controlsvc,no_backends,no_services,no_tls_config,no_workceptor,no_cert_auth ./cmd/receptor-cl && \
+	go build -o receptor --tags no_backends,no_services,no_tls_config,no_workceptor,no_cert_auth ./cmd/receptor-cl && \
```
AaronH88, line 71: @shanemcd can you tell us more about these "no_*" build tags please? 

As part of this refactor we are creating an interface in `controlsvc.go` that is being shared with `interface.go` and with the `no_controlsvc` tag this code would not compile. Just want to make sure these tags are ok to drop.

shanemcd: Look up on line 11 🙂 This was added in the early days of Receptor and I don't think we ever ended up using them.

AaronH88 (approved):
lgtm

PR #832, opened 2023-08-30
Add sysinfo unit test

thom-at-redhat:


AaronH88:
Here is my concern with testify/assert : https://github.com/stretchr/testify/issues/811 

This issue specifies a possible work around that upon viewing seems to add a lot to any benchmark using it.  

I also encouraged the use of table driven unit tests in golang, https://dev.to/boncheff/table-driven-unit-tests-in-go-407b 
You can see that in the code that Chris, Matthew and myself have recently merged. 

I have no problem revisiting these approaches, and so long as any tests we write are easy to read, debug and benchmark I am happy with the team decision (now that their is a team)

pkg/utils/sysinfo_test.go
```diff
@@ -0,0 +1,41 @@
...
+	got := utils.GetSysCPUCount()
+	assert.Positive(t, got)
+
+	commandOutput, err1 := exec.Command("nproc").CombinedOutput()
+
+	assert.Nilf(t, err1, "Unable to get nproc output, output = %s, err = %s", string(commandOutput), err1)
+
+	commandOutputWithout := strings.TrimSpace(string(commandOutput))
+	want, err2 := strconv.Atoi(commandOutputWithout)
+	assert.Nilf(t, err2, "Unable to convert nproc output %s, err = %s", string(commandOutput), err2)
```
AaronH88: Great work on getting info from `nproc` and `/proc/meminfo`, really cleaver. 
I would argue that the two `asserts` that check to ensure data has been returned are really asserting vars outside of the receptor code base and arent needed. 

The positive test and the final equal test are both testing the receptor code and I think these comparisons alone should be enough

pkg/utils/sysinfo_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+	assert.Equal(t, got, want)
+}
+
+func TestGetSysMemoryMiB(t *testing.T) {
+	got := utils.GetSysMemoryMiB()
+	assert.Positive(t, got)
+
+	commandOutput, err1 := exec.Command("sed", "-n", "s/^MemTotal:[[:space:]]*\\([[:digit:]]*\\).*/\\1/p", "/proc/meminfo").CombinedOutput()
```
shanemcd: I'm assuming this only works on Linux right? I'm not sure if we've broken anything else, but Receptor originally worked on macOS / Linux / Windows. I personally care less about Windows, but I _do_ work with it on my Mac so I really hope we can avoid platform specific stuff in the tests.

pkg/utils/sysinfo_test.go
thom-at-redhat: Asserts are now removed

pkg/utils/sysinfo_test.go
thom-at-redhat: added check for OS for each

thom-at-redhat:
> Here is my concern with testify/assert : [stretchr/testify#811](https://github.com/stretchr/testify/issues/811)
> 
> This issue specifies a possible work around that upon viewing seems to add a lot to any benchmark using it.
> 
> I also encouraged the use of table driven unit tests in golang, https://dev.to/boncheff/table-driven-unit-tests-in-go-407b You can see that in the code that Chris, Matthew and myself have recently merged.
> 
> I have no problem revisiting these approaches, and so long as any tests we write are easy to read, debug and benchmark I am happy with the team decision (now that their is a team)

As we discussed, removing the use of testify here

pkg/utils/sysinfo_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"strings"
+	"testing"
+
+	"github.com/ansible/receptor/pkg/utils"
+)
+
+func TestGetSysCPUCount(t *testing.T) {
+	got := utils.GetSysCPUCount()
+	if got <= 0 {
+		t.Fail()
```
AaronH88: Lets use `t.ErrorF()` in all locations and an appropriate error message please

AaronH88 (approved):
lgtm

PR #833, opened 2023-08-31
Add unit tests for packetconn

matoval:
Add unit tests for packetconn
Add interface NetcForPacketConn

### DIFF
--- a/pkg/netceptor/packetconn.go
+++ b/pkg/netceptor/packetconn.go
@@ -60,7 +60,7 @@ type PacketConn struct {
 	cancel          context.CancelFunc
 }
 
-func NewPacketConnWithConst(s NetcForPacketConn, service string, advertise bool, adtags map[string]string, connTypeDatagram byte) PacketConner {
+func NewPacketConnWithConst(s NetcForPacketConn, service string, advertise bool, adtags map[string]string, connTypeDatagram byte) *PacketConn {
 	npc := &PacketConn{
 		s:            s,
 		localService: service,
@@ -77,7 +77,7 @@ func NewPacketConnWithConst(s NetcForPacketConn, service string, advertise bool,
 	return npc
 }
 
-func NewPacketConn(s NetcForPacketConn, service string, connTypeDatagram byte) PacketConner {
+func NewPacketConn(s NetcForPacketConn, service string, connTypeDatagram byte) *PacketConn {
 	return NewPacketConnWithConst(s, service, false, nil, connTypeDatagram)
 }
 
--- a/pkg/netceptor/packetconn.go
+++ b/pkg/netceptor/packetconn.go
@@ -14,6 +14,7 @@ import (
 
 type PacketConner interface {
 	SetHopsToLive(hopsToLive byte)
+	GetHopsToLive() byte
 	SubscribeUnreachable(doneChan chan struct{}) chan UnreachableNotification
 	ReadFrom(p []byte) (int, net.Addr, error)
 	WriteTo(p []byte, addr net.Addr) (n int, err error)
@@ -23,7 +24,7 @@ type PacketConner interface {
 	SetReadDeadline(t time.Time) error
 	SetWriteDeadline(t time.Time) error
 	Cancel() *context.CancelFunc
-	GetLocalService() string
+	LocalService() string
 	GetLogger() *logger.ReceptorLogger
 	StartUnreachable()
 }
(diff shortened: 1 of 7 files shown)


pkg/netceptor/packetconn.go
```diff
@@ -41,24 +60,24 @@ type PacketConn struct {
 	cancel          context.CancelFunc
 }
 
-func NewPacketConnWithConst(s *Netceptor, service string, advertise bool, adtags map[string]string, connTypeDatagram byte) *PacketConn {
+func NewPacketConnWithConst(s NetcForPacketConn, service string, advertise bool, adtags map[string]string, connTypeDatagram byte) PacketConner {
```
AaronH88: Accept interfaces, return structs. Returning a `PacketConner` interface here forces anyone using this to use your interface and this is against the pattern we would like to use :smile:

pkg/netceptor/packetconn.go
```diff
@@ -41,24 +60,24 @@ type PacketConn struct {
...
-	npc.startUnreachable()
-	s.listenerRegistry[service] = npc
+	npc.StartUnreachable()
+	s.GetListenerRegistry()[service] = npc
 
 	return npc
 }
 
-func NewPacketConn(s *Netceptor, service string, connTypeDatagram byte) *PacketConn {
+func NewPacketConn(s NetcForPacketConn, service string, connTypeDatagram byte) PacketConner {
```
AaronH88: Same here as the withconsts function

AaronH88:
Where is sonarcloud? :open_mouth:

pkg/netceptor/packetconn_test.go
```diff
@@ -0,0 +1,298 @@
...
+		{"Service Already Listening Error", "ping", "service ping is already listening", "Expected service ping is already listening, but got: %v"},
+	}
+
+	for _, testCase := range listenPacketTestCases {
+		ctx := context.Background()
+		netc := netceptor.New(ctx, "node")
+
+		t.Run(testCase.name, func(t *testing.T) {
+			_, err := netc.ListenPacket(testCase.service)
+			if testCase.expectedErr == "" && err != nil {
```
AaronH88: This bank of `if` statements is repeated in the next set of tests, can they be combined into a function?

pkg/netceptor/packetconn_test.go
```diff
@@ -0,0 +1,298 @@
...
+		service                string
+		expectedCall           func(ctx context.Context)
+		funcCall               func(pc netceptor.PacketConner) interface{}
+		expectedReturnVal      interface{}
+		unexpectedReturnValMsg string
+	}{
+		{
+			"GetLocalService Success",
+			"test",
+			func(ctx context.Context) {
```
AaronH88, line 131: Ill leave this decision up to you, but, i wouldnt be against defining all these functions outside this test in order to keep this test case table smaller and easier to read. 

If you dont feel that will make a readability difference then im ok keeping it as is.

pkg/netceptor/packetconn_test.go
```diff
@@ -0,0 +1,298 @@
...
+		t.Run(testCase.name, func(t *testing.T) {
+			ctx := context.Background()
+			mockNetceptorForPacketConn.EXPECT().MaxForwardingHops().Return(byte(1))
+			mockNetceptorForPacketConn.EXPECT().GetListenerRegistry().Return(map[string]*netceptor.PacketConn{})
+
+			testCase.expectedCall(ctx)
+
+			var returnVal interface{}
+
+			if testCase.name == "Close Error" {
```
AaronH88: This `if else` block can be refactored to be more efficient or easier to read.

First off, only 1 test case seems to have `"Close Error"`, and yet its the first check on each loop, meaning that the `else` is hit for all bar 1 loops.

This means if we reverse the conditions we will save time as the else will only be hit once  `if testCase.name != "Close Error"` 

Taking it a step further, their is only 1 check really, so it would be even less instructions if we just assign `pc` and `returnval` outside of the `if else` block and then on the condition of `"Close Error"` we override the two vars. This would be easier to read and even less instructions.

AaronH88:
```
pc := netceptor.NewPacketConn(mockNetceptorForPacketConn, testCase.service, 0)
returnVal = testCase.funcCall(pc)
if testCase.name == "Close Error" {
    pc = netceptor.NewPacketConnWithConst(mockNetceptorForPacketConn, testCase.service, true, map[string]string{}, byte(0))
    returnVal = testCase.funcCall(pc)
}
```

pkg/netceptor/packetconn_test.go
AaronH88: happy to keep it the way it is

AaronH88:
Sonar cloud has found 3 code smells for you to address, once that is done we should be good to rebase and merge :smile:

AaronH88 (approved):
lgtm

PR #841, opened 2023-09-10
Fixed URL to python-receptor

jpmens:
The documented URL to python-receptor is 404; I assume this is the correct one.


AaronH88 (approved):
lgtm

Issue #840, opened 2023-09-08
Log the connection name when timing out


fosterseth:
We should name _which_ connection timed out here. It isn't easy from the logs to determine which connection was dropped.

https://github.com/ansible/receptor/blob/8d5dcca7d309ddce954ee63afbc2304b181a7ce7/pkg/netceptor/netceptor.go#L780

Easiest way to hit this error message is to drop traffic, e.g. if I have 2 local nodes, connected on tcp port 2222

`sudo ip6tables -D INPUT -p tcp -s localhost -d localhost --dport 2222 -j DROP`

same with this line https://github.com/ansible/receptor/blob/8d5dcca7d309ddce954ee63afbc2304b181a7ce7/pkg/netceptor/netceptor.go#L909

we should list the connection name

PR #843, opened 2023-09-11
Pass name of connection to logs

matoval:
Fixes issue #840 by passing the connection name for both timeout connection log in monitorConnectionAging method and connection cancelled log in flood method.

### DIFF
--- a/pkg/netceptor/netceptor.go
+++ b/pkg/netceptor/netceptor.go
@@ -764,19 +764,19 @@ func (s *Netceptor) monitorConnectionAging() {
 	for {
 		select {
 		case <-time.After(5 * time.Second):
-			timedOut := make([]context.CancelFunc, 0)
+			timedOut := make(map[string]context.CancelFunc, 0)
 			s.connLock.RLock()
 			for i := range s.connections {
 				conn := s.connections[i]
 				conn.lastReceivedLock.RLock()
 				if time.Since(conn.lastReceivedData) > s.maxConnectionIdleTime {
-					timedOut = append(timedOut, s.connections[i].CancelFunc)
+					timedOut[i] = s.connections[i].CancelFunc
 				}
 				conn.lastReceivedLock.RUnlock()
 			}
 			s.connLock.RUnlock()
 			for i := range timedOut {
-				s.Logger.Warning("Timing out connection, idle for the past %s\n", s.maxConnectionIdleTime)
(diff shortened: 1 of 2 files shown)


pkg/netceptor/netceptor.go
```diff
@@ -902,13 +902,13 @@ func (s *Netceptor) flood(message []byte, excludeConn string) {
 	defer s.connLock.RUnlock()
 	for conn, ci := range s.connections {
 		if conn != excludeConn {
-			go func(ci *connInfo) {
+			go func(conn string, ci *connInfo) {
 				select {
 				case ci.WriteChan <- message:
 				case <-ci.Context.Done():
-					s.Logger.Debug("connInfo cancelled during flood write")
+					s.Logger.Debug("connInfo for connection %s cancelled during flood write", conn)
```
AaronH88, line 929: Would `s.NodeID()` be helpful in this log also @fosterseth ?

AaronH88:
Overall the PR is looking good,
The commit message needs a bit of revision though please.

Please have a look at the guideline we are following here; https://cbea.ms/git-commit/
Specifically:

[Capitalize the subject line](https://cbea.ms/git-commit/#capitalize)
[Use the imperative mood in the subject line](https://cbea.ms/git-commit/#imperative)

pkg/netceptor/netceptor.go
fosterseth: probably not necessary in this case

pkg/netceptor/netceptor.go
```diff
@@ -784,20 +784,20 @@ func (s *Netceptor) monitorConnectionAging() {
...
 				}
-				conn.lastReceivedLock.RUnlock()
+				connInfo.lastReceivedLock.RUnlock()
 			}
 			s.connLock.RUnlock()
-			for i := range timedOut {
-				s.Logger.Warning("Timing out connection, idle for the past %s\n", s.maxConnectionIdleTime)
-				timedOut[i]()
+			for conn := range timedOut {
+				s.Logger.Warning("Timing out connection from %s to %s, idle for the past %s\n", s.nodeID, conn, s.maxConnectionIdleTime)
```
fosterseth: I think the `from nodeA to nodeB` implies connection direction. Meaning I interpret that as nodeA is dialing a tcp listener on nodeB. However, it could be that nodeB dialed nodeA.

I think we should drop `s.nodeID` and just say "timing out connection %s, idle for the past %s`

AaronH88 (approved):
lgtm

PR #866, opened 2023-10-03
Add unit tests for workunitbase and wrap fsnotify library calls

AaronH88:


pkg/workceptor/workunitbase.go
```diff
@@ -32,6 +32,28 @@ const (
 	WorkStateCanceled  = 4
 )
 
+type WatcherWrapper interface {
```
AaronH88, line 36: Comment this please

pkg/workceptor/workunitbase.go
```diff
@@ -74,10 +96,12 @@ type BaseWorkUnit struct {
 	lastUpdateErrorLock *sync.RWMutex
 	ctx                 context.Context
 	cancel              context.CancelFunc
+	fs                  FileSystemer
+	watcher             WatcherWrapper
 }
 
 // Init initializes the basic work unit data, in memory only.
-func (bwu *BaseWorkUnit) Init(w *Workceptor, unitID string, workType string) {
+func (bwu *BaseWorkUnit) Init(w *Workceptor, unitID string, workType string, fs FileSystemer, watcher WatcherWrapper) {
```
AaronH88, line 105: Consider moving these fields to become arguments for the `MonitorLocalStatus()` function

AaronH88: upon review this a bit more I have changed my mind about how good of an idea this becoming arguments is. It seems that it would bleed the internal workings of the status management onto the caller, and that dosent seem like a good idea. going to keep it the original way for now.

matoval (approved):

PR #873, opened 2023-10-12
Correct code coverage

thom-at-redhat:


Makefile
```diff
@@ -49,6 +49,66 @@ endif
...
+	@rm -fv receptorctl/dist/*
+	@rm -fv receptor-python-worker/dist/*
+	@rm -rfv receptorctl-test-venv/
+
+ARCH=amd64
+KIND_BINARY=./kind
+OS=linux
+STABLE_KIND_VERSION=v0.20.0
+
+kind:
```
AaronH88: Why add kind as part of our makefile?

Makefile
```diff
@@ -86,6 +146,19 @@ else
 TESTCMD = -run $(RUNTEST)
 endif
 
+BLOCKLIST='/tests/'
+COVERAGE_FILE='coverage.txt'
+
+coverage: build-all kind
+	PATH="${PWD}:${PATH}" go test $$(go list ./... | grep -v $(BLOCKLIST)) \
```
AaronH88, line 127: the indentation you picked in this PR dosent match the rest of the makefile, is their a reason for this?

Makefile
```diff
@@ -49,6 +49,66 @@ endif
 receptor: $(shell find pkg -type f -name '*.go') ./cmd/receptor-cl/receptor.go
 	CGO_ENABLED=0 GOFLAGS="-buildvcs=false" go build -o receptor $(DEBUGFLAGS) -ldflags "-X 'github.com/ansible/receptor/internal/version.Version=$(VERSION)'" $(TAGPARAM) ./cmd/receptor-cl
 
+clean:
```
AaronH88, line 52: why move this?

AaronH88:
This PR also added kind instillation, but this isnt reflected in either the commit message or PR message. 
If adding kind instillation via a makefile is something we want to do (im not convinced) then please adjust these messages appropriately

Makefile
thom-at-redhat: Per discussions, removed

Makefile
thom-at-redhat: Made it match

Makefile
thom-at-redhat: alphabetizing so we can find targets faster.  We need to reorder all of them in a future PR

Makefile
```diff
@@ -49,6 +49,40 @@ endif
...
+	@rm -fv $(KUBECTL_BINARY)
+	@rm -fv packaging/container/receptor
+	@rm -rfv packaging/container/RPMS/
+	@rm -fv packaging/container/*.whl
+	@rm -fv receptor receptor.exe receptor.app net
+	@rm -fv receptorctl/dist/*
+	@rm -fv receptor-python-worker/dist/*
+	@rm -rfv receptorctl-test-venv/
+
+ARCH=amd64
```
AaronH88, line 65: I think we can use the `uname` command here to discover the os and arch: https://stackoverflow.com/a/714135

thom-at-redhat:
> This PR also added kind instillation, but this isnt reflected in either the commit message or PR message. If adding kind instillation via a makefile is something we want to do (im not convinced) then please adjust these messages appropriately

Removed after discussions

Makefile
thom-at-redhat: we could but it's more involved since the strings don't match.  As we discussed, I would strongly prefer doing such a change in a separate PR

AaronH88 (approved):
lgtm

PR #878, opened 2023-10-18
Add test for pkg/utils/flock

thom-at-redhat:


pkg/utils/flock_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	t.Parallel()
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			if tt.name == "Good Test" {
+				f, err := os.CreateTemp("", "")
```
AaronH88: So this style of testing creates files and directories on the system running these tests. If we look at other unit tests in receptor, we wrap these types of calls so that nothing is created on the system and we mock the responses. 

Some examples of this can be found in how calls to the `os` library have been mocked in `stdio_utils.go` and then used in `stdio_utils_test.go`. 

Please consider wrapping the `syscall` 's in flock using the `Accept interfaces, return structs` principle

PR #879, opened 2023-10-18
Add pkg/utils/unixsock test

thom-at-redhat:


pkg/utils/unixsock_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	t.Parallel()
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			if tt.name == "Good Test" {
+				f, err := os.CreateTemp("", "")
```
AaronH88, line 51: This is similar to the comment I left in the flock tests. We have got examples of mocking calls to `os` in `stdio_utils.go` and we have got examples of mocking `net` in `pkg/netceptor/ping.go` so please reuse them. The only new part is the `TryFlock()` function that is called, and here you have two options. 

1. You can create an interface for `FlockforUnixsocket` that will help decouple these two files from eachother, 
2. You can use all the new mocks and stubs I have asked you to create for `flock_test.go` to control what is returned from `TryFlock()`

PR #896, opened 2023-11-17
Always clean up timed out connections

shanemcd:
We have several customers that have reported seeing errors like this:

```
DEBUG 2023/11/17 12:53:25 Sending routing update IIDrv3lY. Connections: foo(1.00)
DEBUG 2023/11/17 12:53:25 connInfo for connection foo cancelled during flood write
WARNING 2023/11/17 12:53:30 Timing out connection foo, idle for the past 2m0s
DEBUG 2023/11/17 12:53:30 Sending initial connection message
ERROR 2023/11/17 12:53:30 Backend receiving error read tcp [::1]:2222->[::1]:45096: use of closed network connection
ERROR 2023/11/17 12:53:30 Backend error: bar: rejected connection with node foo because it connected using a node ID we are already connected to
ERROR 2023/11/17 12:53:30 Backend sending error write tcp [::1]:2222->[::1]:45096: use of closed network connection
WARNING 2023/11/17 12:53:35 Timing out connection foo, idle for the past 2m0s
DEBUG 2023/11/17 12:53:35 Sending initial connection message
ERROR 2023/11/17 12:53:35 Backend sending error write tcp [::1]:2222->[::1]:45106: use of closed network connection
ERROR 2023/11/17 12:53:35 Backend receiving error read tcp [::1]:2222->[::1]:45106: use of closed network connection
ERROR 2023/11/17 12:53:35 Backend error: bar: rejected connection with node foo because it connected using a node ID we are already connected to
```

This happened if the connection died anywhere in here:

https://github.com/ansible/receptor/blob/e2c1090d60169d0944e6caeabcb3db0ec545c42c/pkg/netceptor/netceptor.go#L2002-L2039

Since it's possible for the connection to be tracked but not established, we need to forcibly remove it when a timeout situation occurs.

### DIFF
--- a/pkg/netceptor/netceptor.go
+++ b/pkg/netceptor/netceptor.go
@@ -798,6 +798,7 @@ func (s *Netceptor) monitorConnectionAging() {
 			for conn := range timedOut {
 				s.Logger.Warning("Timing out connection %s, idle for the past %s\n", conn, s.maxConnectionIdleTime)
 				timedOut[conn]()
+				s.removeConnection(conn)
 			}
 		case <-s.context.Done():
 			return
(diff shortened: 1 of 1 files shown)


matoval (approved):

fosterseth (approved):

fosterseth:
another note around this
at first I thought we could just remove `if established` conditional in the defer method, but that is not right.
In the case where this runProtocol _is_ a duplicate connection, then we don't want to clean up the already established s.connections[nodeid] entry

AaronH88 (approved):
LGTM

AaronH88:
This seems to stem from the `cancelFunc` not being correctly called from the `timedOut` slice. Will spend some time to understand why this `cancelFunc` isnt correctly called

PR #903, opened 2023-12-12
receptorctl: Remove setuptools runtime dependency

dsavineau:
Due to the `pkg_resources` import then setuptools is a runtime dependency.
The `pkg_resources` module is deprecated and we should use `importlib.metadata` instead which is a python builtin (since 3.8)

https://setuptools.pypa.io/en/latest/pkg_resources.html
https://docs.python.org/3.9/library/importlib.metadata.html

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -107,7 +107,7 @@ jobs:
     strategy:
       matrix:
         os-version: ["ubuntu-22.04"]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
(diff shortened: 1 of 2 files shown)


AaronH88:
Thanks for the PR @dsavineau ,

If I understand correctly this module is not available in python 3.7?

If so, please feel free to drop 3.7 support from our CI pipe [here](https://github.com/ansible/receptor/blob/devel/.github/workflows/pull_request.yml#L110) as part of this PR and Ill be happy to merge

AaronH88 (approved):
lgtm

PR #913, opened 2024-01-02
Document functional tests use lsof

kapouer:


shanemcd (approved):
Nice catch, thanks for the PR!

PR #914, opened 2024-01-04
Add support for Go 1.20 and 1.21

AaronH88:
This PR adds support for Go 1.20 and 1.21, 
Updates to the latest version of quic-go along with fixing the breaking changes as a result,
Also introduces a new mutex lock for the packet conns `readDeadline` field, which was not thread safe before.

This PR also addresses the following CVE's

- [CVE-2023-45287](https://access.redhat.com/security/cve/CVE-2023-45287)
- [CVE-2023-39326](https://access.redhat.com/security/cve/CVE-2023-39326)


kapouer:
Building receptor on debian/trixie (with or without your patch) gives the same two failing test_workunit.py tests.
Running those two tests independently, or skipping test_mesh.py, let the tests pass.
So it's probably something about a socket not being properly closed between two tests.

AaronH88:
> Building receptor on debian/trixie (with or without your patch) gives the same two failing test_workunit.py tests. Running those two tests independently, or skipping test_mesh.py, let the tests pass. So it's probably something about a socket not being properly closed between two tests.

Thanks so much for the hint, 
You were correct, the test right before the failing two was meant to clean up a work unit, but the test suite continued too fast, adding a sleep after the cleanup step fixed the problem.

matoval (approved):
LGTM

PR #917, opened 2024-01-09
Add openssl dependency to the test doc

matoval:
Add openssl dependency to the test doc since openssl is required of cli tests and for Receptor in general. Fedora doesn't ship with openssl installed so a new Fedora machine will fail the cli tests.


AaronH88 (approved):
lgtm

PR #924, opened 2024-01-17
Removed isCompatibleK8s, added tests

thom-at-redhat:


pkg/workceptor/kubernetes_test.go
```diff
@@ -1,75 +1,95 @@
...
+		},
+		{
+			name:     "Default test",
+			envValue: "default",
+			want:     false,
+		},
 	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.envValue != "" {
```
AaronH88, line 46: If your interested, this if/else block can be removed by turning `envValue` into a function pointer, and in each test case you `os.Setenv` if needed

AaronH88: Here is an example of it being done this way: https://github.com/ansible/receptor/blob/devel/pkg/workceptor/workceptor_test.go#L140

AaronH88:
Some non blocking feedback given, ill chat with Thom later and we will decide to merge as is or make a small modification

AaronH88 (approved):
lgtm

PR #939, opened 2024-02-05
Fix K8S log reconnect timestamp tracking

TheRealHaoLiu:
Fix bug where `kubeLoggingConnectionHandler` does not respect the tracking that happen's in the main for loop inside `kubeLoggingWithReconnect`

This bug causes us to try to fetch ALL log from beginning of time every time.


chrismeyersfsu (approved):

TheRealHaoLiu:
https://forum.ansible.com/t/awx-job-terminated-unexpectedly/2256/15?u=therealhaoliu 
for record purpose how we came to find this bug

PR #940, opened 2024-02-06
Update auto/default to true in RECEPTOR_KUBE_SUPPORT_RECONNECT

thom-at-redhat:


TheRealHaoLiu:
lets set to enabled by default instead of auto 

auto doesn't function correctly

AaronH88:
> lets set to enabled by default instead of auto
> 
> auto doesn't function correctly

The switch case in this PR only sets the reconnect behaviour to `false` in two cases:
When the env var is set to `disabled`
When the env var is set to something other than the three cases (`enabled`, `disabled`,`auto`) 

All other cases are set to true. So "default" is true unless garbage is given. Is this the behaviour you are expecting?

thom-at-redhat:
@TheRealHaoLiu The comments needed clarification, thanks for the catch!

AaronH88 (approved):
lgtm

TheRealHaoLiu:
need to validate support version range of OCP with supported version of Controller to make sure we are currently not supporting any OCP that do not have the log timestamp bug

PR #941, opened 2024-02-06
Do not set stdoutErr at the end of EOF retry

TheRealHaoLiu:
`stdoutErr` is use to determine `errDetail` that's used in `kw.UpdateBasicStatus(WorkStateFailed, errDetail, stdout.Size())`

in case where we retried 5 time and did not read any new log messages it is not an error it's the expected happy path so we should not set stdoutErr


AaronH88 (approved):
lgtm

PR #942, opened 2024-02-06
Fix erroneous override of work unit status

TheRealHaoLiu:
Certain failure condition that doesn't write stdinErr or stdoutErr result in the code erroneously overriden failed work unit status 

This hides the real failure and cause us to end up with
```
"result_traceback": "Receptor detail:\nFinished",
```

instead of the real reason why the job failed (in this case k8s API server problem...)


AaronH88 (approved):
lgtm

PR #951, opened 2024-02-16
Change default data directory

AaronH88:


AaronH88:
This feature has been merged as part of the installer for aap, we wont add this here

PR #960, opened 2024-03-04
Add cobra cli library for receptor

AaronH88:
Adding draft to show direction so far,
Adding cobra cli library and cli folder structure.

Split receptor inputs into root, backends, cert, resources, services and workers to cover all types of receptor inputs.

The `Run` field in the appropriate command will be used to instantiate the receptor codebase when called. e.g.

```
var tcpListener = &cobra.Command{
	Use:   "tcp-listener",
	Short: "Run a backend listener on a TCP port",
	Long:  ``,
	Args:  cobra.NoArgs,
	Run: func(cmd *cobra.Command, args []string) {
		fmt.Printf("tcpListener called with port %v\n", port)
	},
}
```

TODO:
Finish this format,
Check for ENV vars (cobra library has support for this)  

Adding viper to this will be in a separate PR


cmd/cmd/backends.go
```diff
@@ -0,0 +1,149 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+var (
```
AaronH88, line 9: Perhaps some of these vars can be reused. Check it out

cmd/cmd/backends.go
```diff
@@ -0,0 +1,149 @@
...
+	Long:  ``,
+	Args:  cobra.NoArgs,
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Println("resources called")
+	},
+}
+
+func init() {
+
+	tcpListener.Flags().StringVar(&bindAddr, "bindaddr", "0.0.0.0", "Local address to bind to (default: 0.0.0.0)")
```
AaronH88, line 102: The original message also included the var type, e.g. `bindaddr=<string>: Local address to bind to (default: 0.0.0.0)` do we want to also include the `=<string>` here :thinking:

AaronH88: and by extension all the other flags

cmd/cmd/backends.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/spf13/cobra"
+)
+
+var (
+	bindAddr string
+	// should port be a string?
+	port                   int
+	tlsConfig              string
+	connectionCost         float64
+	nodeCost               map[string]int
```
AaronH88, line 15: so this flag is moving from a `nodecost=<JSON dict of string to float64>:` to something like `string=int,string=int,string=int` is this acceptable?

AaronH88: same everywhere else we see a `map` type

AaronH88:
All the commands have changed, the `--` has been dropped, e.g. `--tcp-listener` has become `tcp-listener` is that ok? :thinking: 

The dashes can be re-added easily

PR #961, opened 2024-03-04
Add Porting Guide

thom-at-redhat:


docs/source/porting_guide/PORTING_2.rst
```diff
@@ -0,0 +1,62 @@
+------------------------
+Receptor 2 Porting Guide
+------------------------
+
+.. contents::
+   :local:
+
+We suggest you read this page along withthe Receptor 2 Changelog to
+understand what updates you may need to make
```
oraNod:
```suggestion
We suggest you read this page along with the Receptor 2 Changelog to understand what updates you may need to make.
```

oraNod: I'd suggest using one sentence per line (semantic line breaks) rather than some arbitrary character limit. RST automatically creates paragraphs from concurrent lines of text. It also makes it easier to do reviews and prevents reflows.

docs/source/porting_guide/PORTING_2.rst
```diff
@@ -0,0 +1,62 @@
+------------------------
+Receptor 2 Porting Guide
+------------------------
```
oraNod: I think you only need the heading format underneath the text. This also applies to the section headings below.

docs/source/porting_guide/receptor_porting_index.rst
```diff
@@ -0,0 +1,13 @@
+=======================
+Receptor Porting Guides
+=======================
+
+This section lists porting guides that can help you in updating your
+environment from one version of receptor to the next.
```
oraNod:
```suggestion
This section lists porting guides that can help you update your environment from one version of Receptor to the next.
```

docs/source/porting_guide/PORTING_2.rst
```diff
@@ -0,0 +1,62 @@
...
+Receptor 2 Porting Guide
```
oraNod: Is "2" the version number here or does it indicate something else? "2.x" might make that a little more obvious.

BhattacharjeeSutapa:
I take a look and agree with @oraNod suggestions.

docs/source/porting_guide/PORTING_2.rst
thom-at-redhat: Makes sense!  We'll try to use semantic line breaks

docs/source/porting_guide/receptor_porting_index.rst
thom-at-redhat: same issue, resolved

docs/source/porting_guide/PORTING_2.rst
thom-at-redhat: Reasonable, updated

docs/source/porting_guide/PORTING_2.rst
thom-at-redhat: It's a preference (see https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections which lists it as overlining)

thom-at-redhat:
> I take a look and agree with @oraNod suggestions.

Thank you for taking the time to review it!

PR #963, opened 2024-03-06
Add upgrade guide

thom-at-redhat:


samccann:
Since there is no difference between upgrading and installing, you can probably just not have this guide. You already have a porting guide where y'all can put important changes from one release to another that users need to know about or do something about, so this upgrade guide is optional.

AaronH88 (approved):
lgtm

thom-at-redhat:
@samccann thank you for your comments, we will consider removing the upgrade guide in the future.

PR #974, opened 2024-03-20
Explicitly remove remoteIDs from the connections table where appropriate

AaronH88:
what if a lot of time elapses between timedOut and removeconnection?
timedOut[conn]()
s.removeConnection(conn) <---- takes a long time to delete conn
during that "takes a long time" it's possible the runProtocol exited, tcp dialed a new connection, and a new runProtocol is started (as it should!).
but then once "takes a long time" occurs, we are going to call removeconnection on that conn again! not good
this is possible because of the structure locks that removeconnection has to acquire before modifying the connections array
this would cause us to be in runProtocol, but the connections array is nowmissing our tcp connection and nothing can get routed. No redialing will occur because runProtocol is running just fine. monitoringconnectionaging won't bail us out because it also no longer sees the conn

Copied from @fosterseth

### DIFF
--- a/pkg/netceptor/netceptor.go
+++ b/pkg/netceptor/netceptor.go
@@ -2022,8 +2022,12 @@ func (s *Netceptor) runProtocol(ctx context.Context, sess BackendSession, bi *Ba
 					select {
 					case initDoneChan <- true:
 					case <-ctx.Done():
+						s.removeConnection(remoteNodeID)
+
 						return nil
 					case <-ci.Context.Done():
+						s.removeConnection(remoteNodeID)
+
 						return nil
 					}
 					s.Logger.SanitizedInfo("Connection established with %s\n", remoteNodeID)
@@ -2043,8 +2047,12 @@ func (s *Netceptor) runProtocol(ctx context.Context, sess BackendSession, bi *Ba
 					select {
 					case s.sendRouteFloodChan <- 0:
 					case <-ctx.Done():
+						s.removeConnection(remoteNodeID)
+
 						return nil
 					case <-ci.Context.Done():
+						s.removeConnection(remoteNodeID)
+
 						return nil
 					}
 					select {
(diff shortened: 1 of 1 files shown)


shanemcd:
@AaronH88 Can you update the PR description to outline what this change does and why it's needed?

fosterseth:
seems there are more `return` statements we'd need to call removeconnection, e.g. https://github.com/ansible/receptor/pull/974/files#diff-23f6be7c946bcb25e244cc53843ea9dfff1ba921dfb1b2645a619c060aa6bd62R2046

I'd still prefer to do this in defer(), maybe before the `if established` block

fosterseth (approved):

pkg/netceptor/netceptor.go
```diff
@@ -1937,13 +1937,17 @@ func (s *Netceptor) runProtocol(ctx context.Context, sess BackendSession, bi *Ba
 						remoteCost, ok := ri.Connections[s.nodeID]
 						if !ok {
 							if remoteEstablished {
+								s.removeConnection(remoteNodeID)
```
shanemcd, line 1940: Do we actually want to do this here? Does this mean that a duplicate node will kick off the originally connected one?

pkg/netceptor/netceptor.go
```diff
@@ -1937,13 +1937,17 @@ func (s *Netceptor) runProtocol(ctx context.Context, sess BackendSession, bi *Ba
...
+
 								return s.sendAndLogConnectionRejection(remoteNodeID, ci, "remote node no longer lists us as a connection")
```
shanemcd, line 1942: This function name seems like it indicates something being rejected. So maybe we don't want to delete this here?

pkg/netceptor/netceptor.go
AaronH88: So I read this code as the remote node thinks it is no longer connected to the current node, however the current node still has the remote node under its connections and thinks it is actively connected.

As a result, the connection needs to be re-established and the connection table cleaned up.

PR #975, opened 2024-03-20
Fix context leaks

thom-at-redhat:


.golangci.yml
AaronH88, line 1: Why add all this?

AaronH88:
I think the name of the PR can do without the word "possible" :smile:

.golangci.yml
thom-at-redhat: There was a partial commented out list of linters, including deprecated ones.  I brought it up to date with all current linters (as I had mentioned this morning).  If you would prefer the commented lines to be removed, I'm fine with that - just let me know..

thom-at-redhat:
> I think the name of the PR can do without the word "possible" 😄

Done

.golangci.yml
thom-at-redhat: @aaronh88 ^

AaronH88: Yes please, remove them.

Im personally not a fan of adding code then commenting it out :smile:

thom-at-redhat: Done,  we need to review the excludes in this file at a later point also

AaronH88 (approved):
lgtm

PR #976, opened 2024-03-21
Remove weird leading g in front of the git sha in version

TheRealHaoLiu:
cause us to scratch our head finding specific receptor version for like...20 min


AaronH88 (approved):
lgtm

PR #980, opened 2024-03-26
[DO NOT MERGE] Log control service command

TheRealHaoLiu:
do not merge, might leak sensitive information to log need to verify

### DIFF
--- a/pkg/controlsvc/controlsvc.go
+++ b/pkg/controlsvc/controlsvc.go
@@ -274,6 +274,7 @@ func (s *Server) RunControlSession(conn net.Conn) {
 		if len(cmdBytes) == 0 {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


pkg/controlsvc/controlsvc.go
```diff
@@ -274,6 +274,7 @@ func (s *Server) RunControlSession(conn net.Conn) {
 		if len(cmdBytes) == 0 {
 			continue
 		}
+		s.nc.GetLogger().Debug("Control service received command: %s\n", string(cmdBytes))
```
AaronH88, line 277: We will need to sanitise this to prevent Log Injection Attacks i think. 
I have already created `SanitizedDebug` to do exactly this, however it comes with a performance cost :smile:

PR #982, opened 2024-04-01
Add load certificate test and support non-standard Go versions in make

thom-at-redhat:


Makefile
```diff
@@ -46,8 +46,10 @@ else
 	DEBUGFLAGS=
 endif
 
+GO ?= go
```
AaronH88, line 49: Any reason for this? 

If so, can you please add this reason to the PR :smile:

pkg/certificates/ca.go
```diff
@@ -21,6 +21,16 @@ import (
 	"github.com/ansible/receptor/pkg/utils"
 )
 
+type Oser interface {
+	ReadFile(name string) ([]byte, error)
+}
+
+type OsWrapper struct{}
+
+func (ow *OsWrapper) ReadFile(name string) ([]byte, error) {
```
AaronH88, line 30: very nice

Makefile
thom-at-redhat: Added, the makefile previously had tests failing with  go 1.20.  this added non-disruptive support for multiple  Go versions

AaronH88 (approved):
lgtm

PR #986, opened 2024-04-05
Add load_from_pem_data test

thom-at-redhat:


pkg/certificates/ca_test.go
```diff
@@ -835,3 +998,263 @@ func TestLoadCertificatesNegative(t *testing.T) {
...
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+
+			o := mock_certificates.NewMockOser(ctrl)
+			switch tt.args.filename {
```
AaronH88: This whole switch case can be swaped for a function pointer in the test case list, something like this: https://github.com/ansible/receptor/blob/devel/pkg/workceptor/workceptor_test.go#L140

pkg/certificates/ca_test.go
```diff
@@ -835,3 +998,263 @@ func TestLoadCertificatesNegative(t *testing.T) {
...
+			case failedToDecodeTestFilename:
```
AaronH88: same as previous comment :smile: 

I think keeping the expected calls with the test case definition in the list is more readable

AaronH88 (approved):
lgtm

PR #991, opened 2024-04-10
Add a noxfile to receptorctl

oraNod:
This PR converts the receptorctl tox file to use a noxfile instead as discussed with @thom-at-redhat The purpose of converting to a noxfile is to eventually add sessions for receptorctl documentation. Many other projects are using noxfiles for their docs so it seems good to be consistent.

Some notes:

- I left the tox file in place. If this PR is merged I can submit a follow up to remove it.
- Also did not update the workflows that call the tox file. I would also like to do that in a follow up PR.
- Rather than using `envlist = py3` from the tox file I included a matrix of versions taken from https://github.com/ansible/receptor/blob/8ab33508ae38cfb9f537063c6451f712059426d0/.github/workflows/pull_request.yml#L113
  I think we'll be able to use the nox session in the workflow and condense several of the steps. This pretty much means moving that matrix from the workflow to the noxfile. Again, I'd like to do that as a follow on. Just wanted to point it out here.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -35,8 +35,8 @@ jobs:
         with:
           python-version: '3.11'
 
-      - name: Install tox
-        run: pip install tox
+      - name: Setup nox
+        uses: wntrblm/nox@2024.03.02
 
       - name: Run receptorctl linters
         run: make receptorctl-lint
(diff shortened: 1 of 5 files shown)


oraNod:
@thom-at-redhat Bah, I guess I'll have to update the ci checks after all since they're calling into the makefile. Should I do that in this PR or would it be better to drop the changes to the makefile and handle all that as a follow on?

thom-at-redhat:
> @thom-at-redhat Bah, I guess I'll have to update the ci checks after all since they're calling into the makefile. Should I do that in this PR or would it be better to drop the changes to the makefile and handle all that as a follow on?

Add here so the pull request passes the checks.  Thank you!

thom-at-redhat:
Re-ran checks after branch protection changes

thom-at-redhat (approved):

PR #1002, opened 2024-04-16
Set Python version in receptorctl directory

oraNod:
Follows up on #998 and #994


receptorctl/.gitignore
```diff
@@ -73,7 +73,7 @@ target/
 .ipynb_checkpoints
 
 # pyenv
-.python-version
+#.python-version
```
webknjaz, line 76: I don't recommend doing this as it's often used to set local virtualenvs. Not everybody is going to have Python of that specific version or could have their own content written by the pyenv commands (which may include a list of versions and virtualenvs). It's quite disruptive for people using pyenv as their workflow tool.

oraNod: Thanks for taking a look @webknjaz I had the feeling this might not be the best idea with potentially unintended side effects. With the goal of having the same Python version to pip-compile requirements with automated bumps in PRs as well as with the noxfile, is the best way likely a workflow that runs the nox session and then submits a PR? Or could there be another way to configure the version that dependabot uses?

webknjaz: I believe that Dependabot parses header comments of the pip-tools managed files and uses the Python version that is mentioned there. Here's an example of what would be used: https://github.com/ansible/receptor/blob/dd4efd9/receptorctl/requirements/tests.txt#L2C57-L2C61.

The Dependabot's logs should be available @ https://github.com/ansible/receptor/network/updates. Though, I can't see it, likely because I don't have enough privileges. Perhaps, you should be able to see it.

Note that what Dependabot does is limited in terms of the environment it runs under. They will always use some Ubuntu VM with x86_64 architecture. So the resulting tree will always be built for this runtime. It may differ for macOS/Windows and arm64/x86_64 etc.

So I'd say you don't really *have to* do anything in this regard, except for changing that header in the constraint files.

oraNod: Thanks, @webknjaz ! Unfortunately I can't access the logs either (limited privileges in this repo). But I do see that happening in the community-website repo. I'll change that header now.

PR #1003, opened 2024-04-16
Create separate receptorctl workflows with matrices

oraNod:
This PR makes a number of changes:

- Add separate receptorctl and lint-receptorctl workflows that the pull request workflow calls. This provides a bit more granularity so it's easier to update and reuse workflows.
- Adds a nox session to create the `receptorctl/.VERSION` file so there is no need to call the makefile to run the receptorctl tests and lint sessions. It's probably better to call nox directly than to do makefile > noxfile.
- Adds python version and session matrices to the receptorctl workflows, which ensures the correct version of Python is used for each step. This also makes it easier to add and update steps, for instance we could potentially add isort to the lint matrix.


.github/workflows/pull_request.yml
```diff
@@ -4,6 +4,9 @@ name: CI
 on:  # yamllint disable-line rule:truthy
   pull_request:  # yamllint disable-line rule:empty-values
 
+env:
+  PYTHON_MATRIX: "3.8, 3.9, 3.10, 3.11"
```
webknjaz: Nit: it seems like the strings are quoted explicitly only in places where there's ambiguous floats. All other places don't add quotes unnecessarily, which happens to be what I prefer as well. How about making this more consistent with the style around?
```suggestion
  PYTHON_MATRIX: 3.8, 3.9, 3.10, 3.11
```

.github/workflows/receptorctl.yml
webknjaz, line 1: I recommend prefixing all the reusable "modules" that only have a `workflow_call` trigger set with `reusable-`. This allows distinguishing the workflow types that can be included in others from the rest.

.github/workflows/reusable_receptorctl_tests.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    - name: Setup nox
+      uses: wntrblm/nox@2024.04.15
+      with:
+        python-versions: "${{ matrix.python-version }}"
+
+    - name: "Run receptorctl nox ${{ matrix.session }} session"
+      run: |
+        nox -s "${{ matrix.session }}"
+      working-directory: ./receptorctl
```
webknjaz: This seems like a lot of duplication. How about
```suggestion
        matrix:
          python-version:
          # NOTE: The highest and the lowest versions come
          # NOTE: first as their statuses are most likely to
          # NOTE: signal problems early
          - 3.11
          - 3.8
          - >-
            3.10
          - 3.9

    env:
      NOXSESSION: tests-${{ matrix.python-version }}

    name: >-  # can't use `env` in this context:
      Run `receptorctl` tests-${{ matrix.python-version }} session

    steps:
    - name: Download the `receptor` binary
      uses: actions/download-artifact@v4
      with:
        name: receptor
        path: /usr/local/bin/

    - name: Set executable bit on the `receptor` binary
      run: sudo chmod a+x /usr/local/bin/receptor

    - name: Set up nox
      uses: wntrblm/nox@2024.04.15
      with:
        python-versions: ${{ matrix.python-version }}

    - name: Check out the source code from Git
      uses: actions/checkout@v4
      with:
        fetch-depth: 0
        sparse-checkout: receptorctl

    - name: Run `receptorctl` nox ${{ env.NOXSESSION }} session
      run: |
        nox
      working-directory: ./receptorctl
```
?

I've reshuffled a few things there, keeping more heavy operations at the end. Prioritizing the supported Python range boundaries to run first and reducing the portion of the Git tree being downloaded. I also made the quoting more consistent.

Let's try it out.

receptorctl/noxfile.py
```diff
@@ -28,12 +29,44 @@ def install(session: nox.Session, *args, req: str, **kwargs):
...
+            "git", "describe", "--tags", "--always", silent=True, external=True
+        )
+        rev = session.run(
+            "git", "rev-parse", "--short", "HEAD", silent=True, external=True
+        )
+        version = tag.split("-")[0] + "+" + rev
+
+    f = open(".VERSION", "w")
+    f.write(version)
+    f.close()
```
webknjaz: It's safer to use a context manager with resources. So it'd be
```suggestion
    with open(".VERSION", "w") as f:
        f.write(version)
```

However, it's even better to use the higher-level APIs that hide all the complexity away:
```suggestion
    Path(".VERSION").write_text(version)
```

receptorctl/noxfile.py
```diff
@@ -28,12 +29,42 @@ def install(session: nox.Session, *args, req: str, **kwargs):
...
+    Path(".VERSION").write_text(version)
+
+
 @nox.session(python=python_versions)
 def tests(session: nox.Session):
     """
     Run receptorctl tests
     """
     install(session, req="tests")
+    version(session)
```
webknjaz, line 70: It doesn't look like this file is even used anywhere. And it's calculated from Git, calling subprocesses on every session in the matrix.
Why do you need it? Are you sure you want it as a part of the PR that simply changes the CI?

receptorctl/noxfile.py
oraNod: Ah, I actually did use the context manager with resources initially but changed it because I thought maybe it was overly complex.

Did not know about `write_text()`, awesome.

receptorctl/noxfile.py
```diff
@@ -28,12 +29,42 @@ def install(session: nox.Session, *args, req: str, **kwargs):
...
+    Create a .VERSION file.
+    """
+    try:
+        official_version = session.run(
+            "git",
+            "describe",
+            "--exact-match",
+            "--tags",
+            external=True,
+            stderr=open(os.devnull, "w"),
```
webknjaz: The `open()` call returns a file descriptor and leaves it open. So I'd expect it to leave an hanging unclosed resource in memory since you don't call `.close()` on that object, nor do you use a context manager that would do it for you.

Moreover, it's customary to pass the built-in `subprocess.DEVNULL` constant to `subprocess.run()` calls which nox's helper is supposed to use under the hood.

.github/workflows/reusable_receptorctl_tests.yml
```diff
@@ -0,0 +1,54 @@
...
+          # NOTE: first as their statuses are most likely to
+          # NOTE: signal problems early
+          - 3.11
+          - 3.8
+          - >-
+            3.10
+          - 3.9
+
+    env:
+      NOXSESSION: tests-${{ matrix.python-version }}
```
oraNod: @webknjaz I see this now. For whatever reason it didn't click from looking at the cherrypy workflow and I was beating myself up trying to get all the sessions to run. Thanks!

.github/workflows/reusable_receptorctl_tests.yml
```diff
@@ -0,0 +1,54 @@
...
+        matrix:
+          python-version:
+          # NOTE: The highest and the lowest versions come
+          # NOTE: first as their statuses are most likely to
+          # NOTE: signal problems early
+          - 3.11
+          - 3.8
+          - >-
+            3.10
+          - 3.9
```
webknjaz: I added a nested matrix by accident the the previous snippet. This should fix it.
```suggestion
        python-version:
        # NOTE: The highest and the lowest versions come
        # NOTE: first as their statuses are most likely to
        # NOTE: signal problems early:
        - 3.11
        - 3.8
        - >-
          3.10
        - 3.9
```

.github/workflows/reusable_receptorctl_tests.yml
webknjaz: Yeah, I had to check the nox's docs to see their env var naming scheme.

receptorctl/noxfile.py
oraNod: The tests sessions fail without it. I was attempting to avoid the need to call nox from the makefile so CI can call nox directly.

In the makefile the file is created here: https://github.com/ansible/receptor/blob/dd4efd9a0b703d8cf975bcf0a3d9af5f21acfdf0/Makefile#L3

And then it gets used for receptorctl targets like this one: https://github.com/ansible/receptor/blob/dd4efd9a0b703d8cf975bcf0a3d9af5f21acfdf0/Makefile#L91

.github/workflows/reusable_receptorctl_checks.yml
```diff
@@ -0,0 +1,34 @@
...
+
+    - name: Setup nox
+      uses: wntrblm/nox@2024.04.15
+      with:
+        python-versions: "${{ matrix.python-version }}"
+
+    - name: "Run receptorctl nox ${{ matrix.session }} session"
+      run: |
+        nox -s "${{ matrix.session }}"
+      working-directory: ./receptorctl
```
webknjaz: Let's reshuffle this for consistency with the other workflow:
```suggestion
    env:
      NOXSESSION: ${{ matrix.session }}

    name: >- # can't use `env` in this context:
      Run `receptorctl` ${{ matrix.session }} session

    steps:
    - name: Set up nox
      uses: wntrblm/nox@2024.04.15
      with:
        python-versions: ${{ matrix.python-version }}

    - name: Check out the source code from Git
      uses: actions/checkout@v4
      with:
        fetch-depth: 0
        sparse-checkout: receptorctl

    - name: >-
        Run `receptorctl` ${{ env.NOXSESSION }} session
      run: |
        nox
      working-directory: ./receptorctl
```

receptorctl/noxfile.py
oraNod: Well spotted. Yeah that's my bad that I didn't call `.close()` and should have caught that.

Thinking about it now, the purpose was to suppress the `fatal:` message when `--exact-match` was not found, as it is done in the makefile. Although I did print a message that the closest annotated tag was being used instead so maybe we just keep them both so it's more informative?

.github/workflows/reusable_receptorctl_tests.yml
```diff
@@ -0,0 +1,53 @@
...
+        # NOTE: first as their statuses are most likely to
+        # NOTE: signal problems early:
+        - 3.11
+        - 3.8
+        - >-
+          3.10
+        - 3.9
+
+    env:
+      NOXSESSION: tests-${{ matrix.python-version }}
```
webknjaz:
```suggestion
      FORCE_COLOR: 1
      NOXSESSION: tests-${{ matrix.python-version }}
```

.github/workflows/reusable_receptorctl_tests.yml
```diff
@@ -0,0 +1,53 @@
...
+    - name: Check out the source code from Git
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        sparse-checkout: receptorctl
+
+    - name: Run `receptorctl` nox ${{ env.NOXSESSION }} session
+      run: |
+        nox
+      working-directory: ./receptorctl
```
webknjaz: Let's separate provisioning from testing:
```suggestion
    - name: Provision nox environment for ${{ env.NOXSESSION }}
      run: nox --install-only
      working-directory: ./receptorctl
    - name: Run `receptorctl` nox ${{ env.NOXSESSION }} session
      run: nox
      working-directory: ./receptorctl
```

.github/workflows/reusable_receptorctl_tests.yml
```diff
@@ -0,0 +1,54 @@
...
+        # NOTE: first as their statuses are most likely to
+        # NOTE: signal problems early:
+        - 3.11
+        - 3.8
+        - >-
+          3.10
+        - 3.9
+
+    env:
+      FORCE_COLOR: 1
```
webknjaz: This one should colorize the output, which makes the CI log nicer. But I recommend playing with the rest of vars from https://github.com/ansible/pylibssh/blob/57e2802/.github/workflows/ci-cd.yml#L68-L73 / https://github.com/cherrypy/cheroot/blob/3591a1c/.github/workflows/ci-cd.yml#L61-L67 to see if there's more improvements from other vars. I know that there's a separate one for pre-commit, for example.

receptorctl/noxfile.py
webknjaz: Oh, I see. Then, I'd move this in-memory, into the pytest runs through its hooks. But maybe, later.

receptorctl/noxfile.py
webknjaz: You can't call `.close()` if you don't save the file descriptor to memory...

.github/workflows/reusable_receptorctl_tests.yml
oraNod: @webknjaz Not sure why we need to do this. And it looks like the sessions are failing now during provisioning.

receptorctl/noxfile.py
```diff
@@ -28,12 +29,41 @@ def install(session: nox.Session, *args, req: str, **kwargs):
     session.install("-r", requirements_directory / f"{req}.in", *args, **kwargs)
 
 
+def version(session: nox.Session):
+    """
+    Create a .VERSION file.
+    """
+    try:
+        official_version = session.run(
```
webknjaz:
```suggestion
        official_version = session.run_install(
```

receptorctl/noxfile.py
```diff
@@ -28,12 +29,41 @@ def install(session: nox.Session, *args, req: str, **kwargs):
...
+            "describe",
+            "--exact-match",
+            "--tags",
+            external=True,
+        )
+    except nox.command.CommandFailed:
+        official_version = None
+
+    if not official_version:
+        tag = session.run(
```
webknjaz:
```suggestion
        tag = session.run_install(
```

receptorctl/noxfile.py
```diff
@@ -28,12 +29,41 @@ def install(session: nox.Session, *args, req: str, **kwargs):
...
+            "git", "describe", "--tags", "--always", silent=True, external=True
+        )
+        rev = session.run(
```
webknjaz:
```suggestion
        rev = session.run_install(
```

.github/workflows/reusable_receptorctl_tests.yml
```diff
@@ -0,0 +1,56 @@
...
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        sparse-checkout: receptorctl
+
+    - name: Provision nox environment for ${{ env.NOXSESSION }}
+      run: nox --install-only
+      working-directory: ./receptorctl
+    - name: Run `receptorctl` nox ${{ env.NOXSESSION }} session
+      run: nox
```
webknjaz:
```suggestion
      run: nox --no-install
```

webknjaz (commented):
Let's try this: https://nox.thea.codes/en/stable/config.html#nox.sessions.Session.run_install.

*(because `session.run()` seems to return early and give us a `None` under `--no-install`; ref: https://nox.thea.codes/en/stable/usage.html#re-using-virtualenvs)*

webknjaz:
Additionally, having https://github.com/marketplace/actions/alls-green integrated would help with setting up a unified gate for branch protection. But that should probably be separate.

.github/workflows/reusable_receptorctl_tests.yml
webknjaz: It's needed to visualize failures of provisioning vs. failures of actual tests more clearly. I've suggested a fix already: https://github.com/ansible/receptor/pull/1003#pullrequestreview-2009328302.

oraNod:
> Additionally, having https://github.com/marketplace/actions/alls-green integrated would help with setting up a unified gate for branch protection. But that should probably be separate.

Reminds me I still need to integrate that in the ansible-documentation repo.

receptorctl/noxfile.py
oraNod: Hmm. I see what you mean. I wonder if it's better to do that now or later. @thom-at-redhat What do you think?

receptorctl/noxfile.py
oraNod: Hey @webknjaz could you take another look at this please? I've updated in an attempt to improve this and would appreciate some feedback. Thanks.

receptorctl/noxfile.py
thom-at-redhat: I would say later, in a follow-on PR

.github/workflows/reusable_receptorctl_tests.yml
thom-at-redhat: I like the idea of standardizing our approach

.github/workflows/reusable_receptorctl_tests.yml
```diff
@@ -0,0 +1,56 @@
...
+
+    - name: Set up nox
+      uses: wntrblm/nox@2024.04.15
+      with:
+        python-versions: ${{ matrix.python-version }}
+
+    - name: Check out the source code from Git
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
```
webknjaz: It just occurred to me that this might not be needed and I was about to suggest deleting this setting. Then I remembered there was something relying on Git in the noxfile... So this confusion with a setting needed in a place disconnected from the rest of the context calls for a code comment with a note:
```suggestion
        fetch-depth: 0  # Needed for the automation in Nox to find the last tag
```

*(copy it to the other workflow too?)*

receptorctl/noxfile.py
```diff
@@ -28,12 +29,47 @@ def install(session: nox.Session, *args, req: str, **kwargs):
...
+    """
+    with open(os.devnull, "w") as devnull:
+        try:
+            official_version = session.run_install(
+                "git",
+                "describe",
+                "--exact-match",
+                "--tags",
+                external=True,
+                stderr=devnull,
```
webknjaz: Have you tried `subprocess.DEVNULL` which is what's customary to use with `subprocess.run()`? Accessing `os.devnull` seems like a low-level *NIX-only hack that I'd avoid.

.github/workflows/reusable_receptorctl_checks.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+    - name: Set up nox
+      uses: wntrblm/nox@2024.04.15
+      with:
+        python-versions: ${{ matrix.python-version }}
+
+    - name: Check out the source code from Git
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
```
webknjaz: Is this even needed for linters?
```suggestion
```

.github/workflows/pull_request.yml
```diff
@@ -100,29 +86,10 @@ jobs:
...
-        run: sudo chmod a+x /usr/local/bin/receptor
-
-      - name: Setup nox
-        uses: wntrblm/nox@2024.04.15
-
-      - name: Run receptorctl tests
-        run: make receptorctl-test
+    uses: ./.github/workflows/reusable_receptorctl_tests.yml
+  lint-receptorctl:
+    uses: ./.github/workflows/reusable_receptorctl_checks.yml
```
webknjaz: So I've been staring at this for a while and couldn't understand why I had a feeling of a déjà vu.. I realized that I solved this much duplication in the past [elsewhere](https://github.com/aio-libs/frozenlist/blob/8c15ec9/.github/workflows/ci-cd.yml#L126-L157). I moved the matrixes out of the reusable workflow, which allowed me to have a single workflow called twice with different matrices. Here, it'd look as follows:
```suggestion
    strategy:
      fail-fast: false
      matrix:
        python-version:
        # NOTE: The highest and the lowest versions come
        # NOTE: first as their statuses are most likely to
        # NOTE: signal problems early:
        - 3.11
        - 3.8
        - >-
          3.10
        - 3.9
    uses: ./.github/workflows/reusable-nox.yml
    with:
      python-version: ${{ matrix.python-version }}
      session: tests-${{ matrix.python-version }}

  lint-receptorctl:
    strategy:
      fail-fast: false
      matrix:
        session:
        - check_style
        - check_format
    uses: ./.github/workflows/reusable-nox.yml
    with:
      python-version: 3.11
      session: ${{ matrix.session }}
```

And the, in `reusable-nox.yml`, you'd have to declare `inputs:` with entries for the Python version and session, remove the matrix definition from the workflow job, and reference the passed data via `${{ inputs }}`. Then, you could reuse this for the tests too, you'd just need to make normalization of how the inputs are used a bit more dynamic inside the reusable workflow.

receptorctl/noxfile.py
oraNod: argh, I think I'm making you repeat yourself as this was mentioned in the previous comment. thanks for catching it, again... maybe I just got fixated, which I tend to do.

webknjaz: No worries! It needs to be tested, though. Since nox's API is a wrapper around subprocess so I don't know how it'll behave..

webknjaz: Oh, and have you considered using `silent=True` instead? https://nox.thea.codes/en/stable/config.html#nox.sessions.Session.run suggests that it suppresses any output unless the command fails which seems useful.

.github/workflows/reusable_receptorctl_checks.yml
oraNod: good call. probably not.

receptorctl/noxfile.py
oraNod: thanks! yes, `subprocess.DEVNULL` seems to work as expected and is a lot simpler than that hack I cobbled together.

I had `silent=True` included but the output is a `fatal` message and the command fails so I guess it doesn't really fit a need.

webknjaz (approved):

thom-at-redhat (approved):

PR #1004, opened 2024-04-16
Add missing workstate

thom-at-redhat:


AaronH88 (approved):
lgtm

Please rebase and we should be good

PR #1009, opened 2024-04-18
Receptorctl: pip compile with python 3.11

oraNod:
Follows up on #1002 and changes the Python version used to pip-compile requirements to 3.11. This PR also updates the version in the header of the requirements.txt files so that Dependabot uses the matching version when bumping dependencies.


oraNod:
@thom-at-redhat Hi, could you please take a look? Python 3.11 seems like a reasonable choice given it is the latest version from the matrix. However maybe we should use a lower version?

thom-at-redhat:
Yes, this seems right.  Nice work!

thom-at-redhat (approved):

PR #1010, opened 2024-04-18
Add nox to receptor_python_worker

thom-at-redhat:


.github/workflows/pull_request.yml
```diff
@@ -221,3 +226,50 @@ jobs:
...
+        working-directory: ./receptor-python-worker
+
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@v4
+        with:
+          directory: receptor-python-worker
+          files: receptor-python-worker_coverage.xml
+          fail_ci_if_error: true
+          token: ${{ secrets.CODECOV_TOKEN }}
+          verbose: true
```
webknjaz: I suggest reusing `.github/workflows/reusable-nox.yml` just like #1003 did. Its steps aren't a 100% match, but you can copy the coverage one there and make it conditional. You'll likely need to add an extra input for toggling it. And declare a secret, and pass it there. You'll also need a new input for customizing the working directory. Nothing more should be necessary.

packaging/container/Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN dnf -y update && \
     dnf clean all && \
     pip3.11 install --no-cache-dir wheel && \
```
webknjaz, line 27: *(probably not exactly related to this PR, but still worth noting)*

Why do you need `wheel` here, by the way? It shouldn't be necessary to use it directly because PEP 517 ephemeral build envs would use their own copy. Besides, it's moved into `setuptools` just last week.

receptor-python-worker/.coveragerc
```diff
@@ -0,0 +1,3 @@
+[run]
+omit = 
+    tests/*
```
webknjaz: FYI, the best practice is to measure the coverage in test files. Moreover, it should be 100% at all times — otherwise, there'll be dead tests that never run.

https://nedbatchelder.com/blog/202008/you_should_include_your_tests_in_coverage.html

receptor-python-worker/setup.cfg
```diff
@@ -1,5 +1,5 @@
 [metadata]
-name = receptor-python-worker
+name = receptor_python_worker
```
webknjaz, line 2: Why do you want to change it? It's usually good to keep the project name separate from the importable name.

receptor-python-worker/tests/test_receptor_python_worker.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+import receptor_python_worker
+
+
+class TestRun:
+    def test_run(self):
+        with pytest.raises(SystemExit) as e:
```
webknjaz, line 7: nit: the best practice when using `pytest.raises()` and similar is to always provide a `match=` argument with a regex.
`flake8-pytest-style` can help you lint this.

receptor-python-worker/tests/test_receptor_python_worker.py
```diff
@@ -0,0 +1,9 @@
...
+            receptor_python_worker.work.run()
+        assert e.type == SystemExit, f"Unknown exception: '{e.type}'"
```
webknjaz, line 9: Do you want to exclude subclasses?

receptor-python-worker/tests/test_receptor_python_worker.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+import receptor_python_worker
+
+
+class TestRun:
```
webknjaz, line 5: In general, when using `pytest`, classes are overkill. You already have regular modules to namespace different tests. So unless there's a good reason, I tend to just use simple functions.

receptor-python-worker/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+import os
+
+test_dir = os.path.dirname(os.path.realpath(__file__))
```
webknjaz, line 3: Is this needed for Python 2 compat? If not — use `pathlib.Path` internally. Also, having `__init__.py` in the test dir changes how `pytest` treats things. It's usually discouraged. And common helpers and configurations would go into `conftest.py` instead.

receptor-python-worker/receptor_python_worker/work.py
```diff
@@ -5,7 +5,7 @@
 import queue
 from pathlib import Path
 import json
-import pkg_resources
+import importlib.metadata
```
webknjaz, line 8: Changing this is not enough. There's code in this file that still calls `pkg_resources` — it will fail with `NameError` until fixed to make new calls.

receptor-python-worker/pyproject.toml
```diff
@@ -1,3 +1,20 @@
...
+description = "The receptor-python-worker command is called by Receptor to supervise the operation of a Python worker plugin."
+dynamic = [
+    "dependencies",
+    "scripts",
+    "version"
+]
+name = "receptor-python-worker"
+
+[project.optional-dependencies]
+dev = [
```
webknjaz, line 20: FWIW, extras is not the best place to put contributor-facing deps. Extras are public API for the end-users. Do you want to expose `pip install receptor-python-worker[dev]` to every arbitrary person on the internet?
Perhaps, when the dependency groups PEP is accepted, this config file will be usable for this dep category, but not today. Today, it's best to describe your dev env as a set of requirements outside the package metadata.

receptor-python-worker/setup.cfg
webknjaz: I also noticed it diverges from the mirror setting in `pyproject.toml`.
```suggestion
name = receptor-python-worker
```

receptorctl/pyproject.toml
```diff
@@ -2,5 +2,23 @@
 requires = ["setuptools"]
 build-backend = "setuptools.build_meta"
 
+[project]
```
webknjaz, line 5: FYI, I saw that using PEP 621 `[project]` table together with `setup.cfg` may result in rather surprising behaviors, so it's best to use just one place for the metadata.

receptor-python-worker/pyproject.toml
```diff
@@ -1,3 +1,20 @@
...
+]
+description = "The receptor-python-worker command is called by Receptor to supervise the operation of a Python worker plugin."
+dynamic = [
+    "dependencies",
+    "scripts",
+    "version"
+]
+name = "receptor-python-worker"
+
+[project.optional-dependencies]
```
webknjaz, line 19: By the way, this is another thing that diverges from the metadata in `setup.cfg`. It'll probably be hard to keep both in sync and make sure that `setuptools` behaves correctly across its versions.

PR #1025, opened 2024-05-02
Add receptorctl coverage reporting

thom-at-redhat:


.github/workflows/pull_request.yml
```diff
@@ -85,39 +91,41 @@ jobs:
...
       matrix:
         session:
-        - check_style
-        - check_format
+          - check_style
+          - check_format
     uses: ./.github/workflows/reusable-nox.yml
     with:
-      python-version: 3.11
+      python-version: '3.11'
```
matoval, line 126: Should this use DESIRED_PYTHON_VERSION?

matoval (commented):
Looks good, just have one question

.github/workflows/pull_request.yml
thom-at-redhat: I would like to, there was an issue which will be addressed after this PR

matoval (approved):

PR #1039, opened 2024-05-15
Fix makefile targets so version files are properly created

thom-at-redhat:


shanemcd:
This seems like you might just be debugging something? Please open PRs like this against your own fork, to keep the notification spam down.

PR #1049, opened 2024-05-22
Add receptorctl reload doc

thom-at-redhat:


matoval (approved):

docs/source/receptorctl/receptorctl_reload.rst
```diff
@@ -0,0 +1,24 @@
+------
+reload
+------
+
+.. contents::
+   :local:
+
+``receptorctl reload`` Reload configuration on the Receptor node to which it is connected.
```
oraNod:
```suggestion
``receptorctl reload`` reloads the Receptor configuration for the connected node.
```

oraNod: Just noticed we're being inconsistent with the capitalization in the description. Most of the docs are lower-case but version is an exception "Displays": https://github.com/ansible/receptor/blob/devel/docs/source/receptorctl/receptorctl_version.rst

oraNod (approved):

PR #1053, opened 2024-05-30
Fix Go action failures when artifact already exists

thom-at-redhat:


matoval (approved):

.github/workflows/reusable-nox.yml
```diff
@@ -17,7 +17,10 @@ on:
         description: Whether to perform go binary download.
         required: false
         default: false
-
+      go-version:
+        type: string
+        description: The Go version to use.
+        required: false
```
webknjaz, line 23: Since it's not `required: true`, I'd at least set a `default: unversioned` or smth.

PR #1064, opened 2024-06-11
Add receptorctl work results doc

thom-at-redhat:


docs/source/receptorctl/receptorctl_work_results.rst
```diff
@@ -0,0 +1,26 @@
+------------
+work results
+------------
+
+.. contents::
+   :local:
+
+``receptorctl work results`` Get results for previously or currently running unit(s) of work.
```
oraNod:
```suggestion
``receptorctl work results`` get results for completed, or currently running, units of work.
```

docs/source/receptorctl/receptorctl_work_results.rst
```diff
@@ -0,0 +1,26 @@
...
+.. code-block:: text
+
+  ss --listening --processes --unix 'src = unix:<socket_path>'
+  Netid         State          Recv-Q         Send-Q                   Local Address:Port                     Peer Address:Port        Process
+  u_str         LISTEN         0              4096                   /tmp/local.sock 38130170                            * 0            users:(("receptor",pid=3226769,fd=7))
+
+``ps -fp $(pidof receptor)``
+``lsof -p <pid>``
+
+``Unit ID`` is a unique identifier for a work unit (job).  When running the ``work results`` command, you should specify the ``Unit ID`` for the Receptor node to which you are connected.
```
oraNod:
```suggestion
``Unit ID`` is a unique identifier for a work unit (job). When running the ``work results`` command, you should specify the ``Unit ID`` for the Receptor node to which you are connected.
```

docs/source/receptorctl/receptorctl_work_results.rst
oraNod: I feel like "previous" and "currently running" are two separate things so adding commas makes the two things distinct. Conjoining with or implies something was "previously running" which is more ambiguous to the final state. Does that include failed or aborted jobs?

Using "completed" here might make it a bit more clear that the job ran in the past. If it is the case that the command returns results for failed or aborted jobs, maybe we could flip things and say "currently or no longer running units of work". Not sure if that is great though. The best way might be really explicit but more verbose. "successfully completed, failed, stopped, or currently running units of work"

thom-at-redhat: The last is the winner, thank you for suggesting it

docs/source/receptorctl/receptorctl_work_results.rst
```diff
@@ -0,0 +1,26 @@
+------------
+work results
+------------
+
+.. contents::
+   :local:
+
+``receptorctl work results`` Get results for successfully completed, failed, stopped, or currently running, units of work.
```
oraNod:
```suggestion
``receptorctl work results`` gets results for successfully completed, failed, stopped, or currently running, units of work.
```

oraNod (approved):

resoluteCoder (approved):

PR #1065, opened 2024-06-11
Add goleak test to test receptor with basic config

matoval:
Jira: https://issues.redhat.com/browse/AAP-16223

Add goleak test to test receptor with basic config

Added context done case to sendResult to resolve goroutine leak


AaronH88 (approved):
lgtm

PR #1069, opened 2024-06-14
add receptorctl work submit doc

thom-at-redhat:


docs/source/receptorctl/receptorctl_work_submit.rst
```diff
@@ -0,0 +1,39 @@
...
+  u_str         LISTEN         0              4096                   /tmp/local.sock 38130170                            * 0            users:(("receptor",pid=3226769,fd=7))
+
+``ps -fp $(pidof receptor)``
+``lsof -p <pid>``
+
+``WorkType`` is the execution request type for the work unit.
+
+``Command Parameters`` are parameters passed to the work command.
+
+``-a``, ``--param TEXT`` adds additional Receptor parameter in key=value format.
```
oraNod:
```suggestion
``-a``, ``--param TEXT`` adds additional Receptor parameter in key=value format.
```
```suggestion
``-a``, ``--param TEXT`` adds a Receptor parameter in key=value format.
```

oraNod: "adds additional" sounds like redundant language to me.

docs/source/receptorctl/receptorctl_work_submit.rst
```diff
@@ -0,0 +1,39 @@
...
+``-f``, ``--follow`` causes Receptorctl to remain attached to the job and print its results.
```
oraNod:
```suggestion
``-f``, ``--follow`` causes Receptorctl to remain attached to the job and print its results.
```
```suggestion
``-f``, ``--follow`` keeps Receptorctl attached to the job and displays the job results.
```

docs/source/receptorctl/receptorctl_work_submit.rst
```diff
@@ -0,0 +1,39 @@
...
+``-l``, ``--payload-literal TEXT`` uses the ``TEXT`` as the literal unit of work data.
```
oraNod:
```suggestion
``-l``, ``--payload-literal TEXT`` uses the ``TEXT`` as the literal unit of work data.
```
```suggestion
``-l``, ``--payload-literal <<TEXT>>`` uses the value of ``<<TEXT>>`` as the literal unit of work data.
```

docs/source/receptorctl/receptorctl_work_submit.rst
```diff
@@ -0,0 +1,39 @@
...
+``-n``, ``--no-payload`` sends an empty payload.
+``--node <<Node ID>>`` is the Receptor node to run the work on. It defaults to the local node.
```
oraNod:
```suggestion
``--node <<Node ID>>`` is the Receptor node to run the work on. It defaults to the local node.
```
```suggestion
``--node <<Node ID>>`` is the Receptor node on which the work runs. The default is the local node.
```

docs/source/receptorctl/receptorctl_work_submit.rst
```diff
@@ -0,0 +1,39 @@
...
+``-p``, ``--payload TEXT`` specifies the file containing unit of work data. Use - for stdin.
```
oraNod:
```suggestion
``-p``, ``--payload TEXT`` specifies the file containing unit of work data. Use - for stdin.
```
```suggestion
``-p``, ``--payload <<TEXT>>`` specifies the file that contains data for the unit of work. Specify ``-`` for standard input (stdin).
```

docs/source/receptorctl/receptorctl_work_submit.rst
```diff
@@ -0,0 +1,39 @@
...
+``--rm`` releases the work unit after completion.
+``--signwork`` digitally signs remote work submissions to stdout.
```
oraNod:
```suggestion
``--signwork`` digitally signs remote work submissions to stdout.
```
```suggestion
``--signwork`` digitally signs remote work submissions to standard output (stdout).
```

docs/source/receptorctl/receptorctl_work_submit.rst
```diff
@@ -0,0 +1,39 @@
...
+``--tls-client TEXT`` specifies the TLS client used when submitting work to a remote node.
```
oraNod:
```suggestion
``--tls-client TEXT`` specifies the TLS client used when submitting work to a remote node.
```
```suggestion
``--tls-client <<TEXT>>`` specifies the TLS client that submits work to a remote node.
```

docs/source/receptorctl/receptorctl_work_submit.rst
```diff
@@ -0,0 +1,39 @@
...
+``--ttl TEXT`` specifies the time to live until remote work must start, e.g. 1h20m30s or 30m10s.
```
oraNod:
```suggestion
``--ttl TEXT`` specifies the time to live until remote work must start, e.g. 1h20m30s or 30m10s.
```
```suggestion
``--ttl <<TEXT>>`` specifies the time to live (TTL) for remote work requests in ``##h##m##s`` format; for example ``1h20m30s`` or ``30m10s``.
```

oraNod: Assuming TTL here is the lifespan, what happens when that value gets exceeded? Does the request fail? Is there an error message somewhere?

docs/source/receptorctl/receptorctl_work_submit.rst
```diff
@@ -0,0 +1,39 @@
+-----------
+work submit
+-----------
+
+.. contents::
+   :local:
+
+``receptorctl work submit`` submits a new unit of work.
```
oraNod: Maybe it's a bit more wordy but I never like using the same word as the command option to describe what happens. What about?

"requests a Receptor node to run a unit of work."?

docs/source/receptorctl/receptorctl_work_submit.rst
oraNod: I'm being kind of lazy here and not setting things up to find out for myself - just going to ask the question. But does this add a new unit of work to a processing queue on the Receptor node?

If so, is there a way for users to find out what other work items are already in the queue and use that information to determine the best TTL value?

Also, does Receptor run jobs concurrently? Say there is a thread per job, what performance considerations are there?

docs/source/receptorctl/receptorctl_work_submit.rst
```diff
@@ -0,0 +1,39 @@
...
+.. code-block:: text
+
+  ss --listening --processes --unix 'src = unix:<socket_path>'
+  Netid         State          Recv-Q         Send-Q                   Local Address:Port                     Peer Address:Port        Process
+  u_str         LISTEN         0              4096                   /tmp/local.sock 38130170                            * 0            users:(("receptor",pid=3226769,fd=7))
+
+``ps -fp $(pidof receptor)``
+``lsof -p <pid>``
+
+``WorkType`` is the execution request type for the work unit.
```
oraNod: Another question here is what are the available request types? Are they enumerated somewhere or are they defined by the user?

docs/source/receptorctl/receptorctl_work_submit.rst
thom-at-redhat: > I'm being kind of lazy here and not setting things up to find out for myself - just going to ask the question. But does this add a new unit of work to a processing queue on the Receptor node?
Yes, it will add a new unit of work on the local node and a separate unit of work on the remote node.
> 
> If so, is there a way for users to find out what other work items are already in the queue and use that information to determine the best TTL value?
Yes for the other work items (receptorctl work list).

The TTL question is interesting but some work would need to be done to suggest an approach.
> 
> Also, does Receptor run jobs concurrently? Say there is a thread per job, what performance considerations are there?
Yes, Receptor runs jobs concurrently on each execution plane node.  Some work would need to be done to suggest approach for performance considerations.

docs/source/receptorctl/receptorctl_work_submit.rst
thom-at-redhat: the current work types for each Receptor execution plane node are returned by the receptorctl status command.

oraNod:
```suggestion
``WorkType`` specifies an execution request type for the work unit. Use the ``receptorctl status`` command to find available work types for Receptor nodes.
```

docs/source/receptorctl/receptorctl_work_submit.rst
```diff
@@ -0,0 +1,39 @@
...
+
+``Command Parameters`` are parameters passed to the work command.
```
oraNod:
```suggestion
``Command Parameters`` modify the requests that you send with the ``work`` command. The following are the available parameters:
```

oraNod: Not sure what happened but my previous suggestion seems to have been botched. Hopefully this new suggestion is good. I would def recommend introducing the list of parameters with something like "the following:" or "as follows:"

docs/source/receptorctl/receptorctl_work_submit.rst
```diff
@@ -0,0 +1,39 @@
...
+``-a``, ``--param <<KEY>>=<<VALUE>>`` adds a Receptor parameter in key=value format.
+``-f``, ``--follow`` keeps Receptorctl to remain attached to the job and displays the job results.
+``-l``, ``--payload-literal <<TEXT>>`` uses the value of ``<<TEXT>>`` as the literal unit of work data.
+``-n``, ``--no-payload`` sends an empty payload.
+``--node <<Node ID>>`` is the Receptor node on which the work runs. The default is the local node.
+``-p``, ``--payload <<TEXT>>`` specifies the file that contains data for the unit of work. Specify ``-`` for standard input (stdin).
+``--rm`` releases the work unit after completion.
+``--signwork`` digitally signs remote work submissions to standard output (stdout).
+``--tls-client <<TEXT>>`` specifies the TLS client that submits work to a remote node.
+``--ttl <<TEXT>>`` specifies the time to live (TTL) for remote work requests in ``##h##m##s`` format; for example ``1h20m30s`` or ``30m10s``.
```
oraNod:
```suggestion
``--ttl <<TEXT>>`` specifies the time to live (TTL) for remote work requests in ``##h##m##s`` format; for example ``1h20m30s`` or ``30m10s``. Use the ``receptorctl work list`` command to display units of work on Receptor nodes and determine appropriate TTL values.
```

docs/source/receptorctl/receptorctl_work_submit.rst
oraNod: Cool. Performance considerations are always tricky with things clustered across networks. There are so many factors that can impact performance.

It might be useful to give some general guidance like "do some evaluation in a test environment to get a feel for how long work units take to run under your specific conditions before you request work units in production". Maybe also something along the lines of "try to balance requests for units of work across Receptor nodes to ensure the workload is distributed as evenly as possible". Although that might be a bit too vague and come across as unhelpful in the docs. Finding the optimal number of work units per node really depends and is something users need to find for themselves - docs for that are well out of the scope of this PR so maybe we should just leave it. If you do want to have a stab at any of that stuff, I'm game.

thom-at-redhat: i would say leave it for now.  If needed, I would suggest a Performance doc

docs/source/receptorctl/receptorctl_work_submit.rst
thom-at-redhat: Please see https://ansible.readthedocs.io/projects/receptor/en/latest/user_guide/workceptor.html#runtime-parameters for how Command Parameters work.
They are distinct from the "Options" listed below.  The terms are confusingly similar and we may want to distinguish between Receptor-related parameters/options and worktype parameters.

oraNod: Can we change ``Command Parameters`` to ``Runtime Parameters`` here?

docs/source/receptorctl/receptorctl_work_submit.rst
```diff
@@ -0,0 +1,39 @@
...
+  Netid         State          Recv-Q         Send-Q                   Local Address:Port                     Peer Address:Port        Process
+  u_str         LISTEN         0              4096                   /tmp/local.sock 38130170                            * 0            users:(("receptor",pid=3226769,fd=7))
+
+``ps -fp $(pidof receptor)``
+``lsof -p <pid>``
+
+``WorkType`` specifies an execution request type for the work unit.  Use the ``receptorctl status`` command to find available work types for Receptor nodes.
+
+``Command Parameters`` are parameters passed to the work command.
+
```
oraNod, line 35: Maybe we could create a section here for the options to try and make it clear that these are different to the cmd parameters?

```
^^^^^^^^^^^^^^^^^^
Work submit options
^^^^^^^^^^^^^^^^^^

You can use the following options with the ``work submit`` command:
```

oraNod (approved):
LGTM. Thanks @thom-at-redhat

matoval (approved):

PR #1105, opened 2024-08-02
Add unit test for addr.go

matoval:
Add unit test for addr.go

Addr struct properties need to be public in order to create a test Addr struct outside of netceptor for this test.

### DIFF
--- a/pkg/netceptor/addr.go
+++ b/pkg/netceptor/addr.go
@@ -5,8 +5,8 @@ import "fmt"
 // Addr represents an endpoint address on the Netceptor network.
 type Addr struct {
 	network string
-	node       string
-	service    string
+	node    string
+	service string
 }
 
 // Network returns the network name.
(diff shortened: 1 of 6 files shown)


pkg/netceptor/addr.go
```diff
@@ -4,17 +4,17 @@ import "fmt"
 
 // Addr represents an endpoint address on the Netceptor network.
 type Addr struct {
-	network string
-	node    string
-	service string
+	NetworkStr string
```
AaronH88: Great work on this, can you quickly run through why the vars need to be exported please?

matoval: The vars need to be exported in order for netceptor_test to create an Addr with set vars for this test. The vars need to be set so that Network() and String() return values

AaronH88 (approved):
lgtm

PR #1106, opened 2024-08-02
included -h filter for os.exit

resoluteCoder:
Added the `-h` to the check for exiting without spinning up receptor


AaronH88 (approved):
lgtm

PR #1111, opened 2024-08-05
change --legacy to --latest to switch v1 to default

resoluteCoder:
Changing the flag from `--legacy` to `--latest` so that v1 config is set as the default.  `--legacy` broke backwards compatibility and having v2 with `--latest` will allow for further testing of the new config in the meantime.

### DIFF
--- a/cmd/receptor-cl/receptor.go
+++ b/cmd/receptor-cl/receptor.go
@@ -13,11 +13,11 @@ import (
 )
 
 func main() {
-	var latest bool
+	var isV2 bool
 	newArgs := []string{}
 	for _, arg := range os.Args {
-		if arg == "--latest" {
-			latest = true
+		if arg == "--config-v2" {
+			isV2 = true
 
 			continue
 		}
@@ -26,8 +26,8 @@ func main() {
 
 	os.Args = newArgs
 
-	if latest {
-		fmt.Println("Running latest cli/config")
+	if isV2 {
+		fmt.Println("Running v2 cli/config")
 		cmd.Execute()
 	} else {
 		cmd.RunConfigV1()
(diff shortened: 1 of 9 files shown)


packaging/container/Dockerfile
```diff
@@ -36,4 +36,4 @@ ENV RECEPTORCTL_SOCKET=/tmp/receptor.sock
 EXPOSE 7323
 
 ENTRYPOINT ["/usr/local/bin/dumb-init", "--"]
-CMD ["/usr/bin/receptor", "--config", "/etc/receptor/receptor.yaml"]
```
AaronH88, line 39: change `confg` to `conf` please, I think this is causing the container to fail

packaging/container/Dockerfile
```diff
@@ -36,4 +36,4 @@ ENV RECEPTORCTL_SOCKET=/tmp/receptor.sock
...
+CMD ["/usr/bin/receptor", "-c", "/etc/receptor/receptor.confg"]
```
AaronH88: change `confg` to `conf` please, I think this is causing the container to fail

AaronH88 (approved):
lgtm

cmd/receptor-cl/receptor.go
```diff
@@ -26,7 +26,7 @@ func main() {
 
 	os.Args = newArgs
 
-	if !legacy {
+	if latest {
 		cmd.Execute()
 	} else {
 		fmt.Println("Running old cli/config")
```
AaronH88: lets more this print to the new config, else we will be printing it each time

AaronH88 (approved):
lgtm

PR #1123, opened 2024-08-13
Randomize receptor_tmp_dir

thom-at-redhat:


AaronH88:
alot of the stack above us require the full dir path for receptor as well as the installer needs to know which file to grant permissions to, e.g. read and write permissions to /tmp/receptor

is their a reason for this randomisation?

thom-at-redhat:
> alot of the stack above us require the full dir path for receptor as well as the installer needs to know which file to grant permissions to, e.g. read and write permissions to /tmp/receptor
> 
> is their a reason for this randomisation?

Yes, we discussed this privately

thom-at-redhat:
Deferred and will reassess in the future

PR #1133, opened 2024-08-30
Add qlog tracer

matoval:
Add qlog tracer to the quic config when QLOGDIR environment variable is provided


AaronH88:
The code is looking good but I am very surprised at the amount of changes in `go.sum` 
Can you please double check that this amount is correct and qVis dependencies have not been added accidentally

matoval:
> The code is looking good but I am very surprised at the amount of changes in `go.sum` Can you please double check that this amount is correct and qVis dependencies have not been added accidentally

I double checked the deps and there was one unneeded dep that was added but the rest is need for using qlog

pkg/netceptor/conn.go
```diff
@@ -90,6 +92,7 @@ func (s *Netceptor) listen(ctx context.Context, service string, tlscfg *tls.Conf
 	s.Logger.Debug("%s added service %s to listener registry", s.nodeID, service)
 	s.listenerRegistry[service] = pc
 	cfg := &quic.Config{
+		Tracer:                  s.tracer,
```
AaronH88, line 95: What happens if this is nil?

matoval: Quic checks if tracer != nil and handles if nil. DefaultConnectionTracer returns nil if qlogDir is not defined: https://github.com/quic-go/quic-go/blob/master/qlog/qlog_dir.go#L41

AaronH88 (approved):
lgtm

PR #1135, opened 2024-09-04
added hot reload to receptor config

resoluteCoder:


cmd/config.go
```diff
@@ -157,6 +174,33 @@ func RunPhases(phase string, v reflect.Value) {
 	}
 }
 
+// ReloadServices iterates through key/values calling reload on applicable services.
+func ReloadServices(v reflect.Value) {
+	for i := 0; i < v.NumField(); i++ {
+		if reflect.Value.IsZero(v.Field(i)) {
+			continue
```
AaronH88, line 181: Add a comment here please, 
Suggestion:
"If the service is not part of the backends struct, do nothing"

cmd/config.go
```diff
@@ -157,6 +174,33 @@ func RunPhases(phase string, v reflect.Value) {
 	}
 }
 
+// ReloadServices iterates through key/values calling reload on applicable services.
+func ReloadServices(v reflect.Value) {
+	for i := 0; i < v.NumField(); i++ {
+		if reflect.Value.IsZero(v.Field(i)) {
```
AaronH88, line 181: Add a comment here please,
Suggestion:
"If the service is not initialised, skip"

cmd/config.go
```diff
@@ -157,6 +174,33 @@ func RunPhases(phase string, v reflect.Value) {
...
+			continue
+		}
+
+		var err error
+		switch v.Field(i).Kind() {
+		case reflect.Slice:
+			for j := 0; j < v.Field(i).Len(); j++ {
```
AaronH88, line 189: Add a comment here please,
Suggestion:
"Iterate over all the type fields"

cmd/config.go
```diff
@@ -157,6 +174,33 @@ func RunPhases(phase string, v reflect.Value) {
...
+				serviceItem := v.Field(i).Index(j).Interface()
+				switch c := serviceItem.(type) {
+				case Reloader:
```
AaronH88, line 194: Add a comment here please,
Suggestion:
"Check to see if the selected type field is of type Reloader interface, meaning reload can be called"

cmd/config.go
```diff
@@ -157,6 +174,33 @@ func RunPhases(phase string, v reflect.Value) {
...
+					err = c.Reload()
+					if err != nil {
+						PrintPhaseErrorMessage(v.Type().Name(), "reload", err)
+					}
+				default:
```
AaronH88, line 200: Add a comment here please,
Suggestion:
"If the service does not implement the Reloader interface, do nothing"

AaronH88 (commented):
Overall looking good, 
I would like a few more comments around the reflection function to make it quicker to read once we revisit later

The functionality looks good to me,
@thom-at-redhat and @matoval can you look also please

cmd/root.go
```diff
@@ -6,15 +6,20 @@ import (
...
 
 var (
-	cfgFile string
-	version bool
+	cfgFile       string
+	version       bool
+	backendConfig *BackendConfig
 )
 
+var FindMe = true
```
matoval: What is this var used for?

cmd/root.go
```diff
@@ -78,8 +60,85 @@ func initConfig() {
...
+		// allows to skip empty first config
+		isEmpty := isConfigEmpty(reflect.ValueOf(*newConfig))
+		if isEmpty {
+			return
+		}
+
+		SetBackendConfigDefaults(newConfig)
+
+		isEqual := reflect.DeepEqual(*backendConfig, *newConfig)
+		if !isEqual {
```
matoval, line 79: Should we add an else that alerts the user that the config change wasn't reloadable with a log?

cmd/root.go
```diff
@@ -78,8 +60,85 @@ func initConfig() {
 
 	viper.AutomaticEnv()
 
+	viper.OnConfigChange(func(e fsnotify.Event) {
+		fmt.Printf("Config file changed: %s\n", e.Name)
```
matoval: Could this be a log instead of a print?

matoval (commented):
Looks great, I have a few questions

matoval (approved):

PR #1146, opened 2024-09-18
Add debug log for work unit payloads

matoval:
Add debug log for work unit payloads using RECEPTOR_PAYLOAD_DEBUG env var

Control output:
Screenshot from 2024-09-18 13-11-28

Control status:
Screenshot from 2024-09-18 13-04-05

Execution output:
image

Screenshot from 2024-09-18 13-04-33

### DIFF
--- a/docs/source/user_guide/configuration_options.rst
+++ b/docs/source/user_guide/configuration_options.rst
@@ -63,19 +63,21 @@ Log level
       - string
 
 Add payload debuging using `RECEPTOR_PAYLOAD_DEBUG=int` envorment variable and using log level debug.
-... list-table: `RECEPTOR_PAYLOAD_DEBUG` options:
+
+.. list-table:: RECEPTOR_PAYLOAD_DEBUG options
     :header-rows: 1
     :widths: auto
-  * - Debug level
-    - Description
-  * - 0
-    - No payload debug log
-  * - 1
-    - Log connection type
-  * - 2
-    - Log connection type and work unit id
-  * - 3
-    - Log connection type, work unit id and payload
+
+    * - Debug level
+      - Description
+    * - 0
+      - No payload debug log
+    * - 1
+      - Log connection type
+    * - 2
+      - Log connection type and work unit id
+    * - 3
+      - Log connection type, work unit id and payload
 
 **Warning: Payload Debugging May Expose Sensitive Data**
 
--- a/docs/source/user_guide/configuration_options.rst
+++ b/docs/source/user_guide/configuration_options.rst
@@ -63,11 +63,19 @@ Log level
       - string
 
 Add payload debuging using `RECEPTOR_PAYLOAD_DEBUG=int` envorment variable and using log level debug.
-`RECEPTOR_PAYLOAD_DEBUG` options:
-* 0: No payload debug log
-* 1: Log connection type
-* 2: Log connection type and work unit id
-* 3: Log connection type, work unit id and payload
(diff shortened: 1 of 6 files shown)


AaronH88:
Great work on this, 
I am imagining using this now and I am throwing out ideas, 

First idea, lets change `RECEPTOR_PAYLOAD_DEBUG` to `RECEPTOR_PAYLOAD_TRACE_LEVEL` which will expect an `int` 
Now that we have different "levels" for tracing, lets build the debug message in a `switch case` where the higher the `int` number results in more of the payload data being printed. 

Imagine level 1 we just print something like "Unix message" , "Network message". 
Level 2 is like the full payload, which you have, 
Then Level 3 could be the payload plus the work unit id. (which you got!) 

Let me know what you think. 

The next thing to think about is code reusability (might be outside the scope of this PR) 
Could this code be put in the logger and reused?

AaronH88:
> Great work on this, I am imagining using this now and I am throwing out ideas,
> 
> First idea, lets change `RECEPTOR_PAYLOAD_DEBUG` to `RECEPTOR_PAYLOAD_TRACE_LEVEL` which will expect an `int` Now that we have different "levels" for tracing, lets build the debug message in a `switch case` where the higher the `int` number results in more of the payload data being printed.
> 
> Imagine level 1 we just print something like "Unix message" , "Network message". Level 2 is like the full payload, which you have, Then Level 3 could be the payload plus the work unit id. (which you got!)
> 
> Let me know what you think.
> 
> The next thing to think about is code reusability (might be outside the scope of this PR) Could this code be put in the logger and reused?


Thinking more on this, in my example maybe we should switch the level 2 and 3 😂

pkg/logger/logger.go
```diff
@@ -148,6 +148,31 @@ func (rl *ReceptorLogger) Debug(format string, v ...interface{}) {
 	rl.Log(DebugLevel, format, v...)
 }
 
+// Debug payload data.
+func (rl *ReceptorLogger) DebugPayload(payloadDebug int, payload string, workUnitID string, connectionType string) {
+	switch payloadDebug {
+	case 3:
```
AaronH88, line 157: Potentially 3 traces messages per payload seems nosey to me, Can we use these cases to "build" a log message instead and just log it once please?

docs/source/user_guide/configuration_options.rst
```diff
@@ -62,6 +62,27 @@ Log level
       - Error
       - string
 
+Add payload debuging using `RECEPTOR_PAYLOAD_DEBUG=int` envorment variable and using log level debug.
```
AaronH88: The Env var is called `RECEPTOR_PAYLOAD_TRACE_LEVEL` not this

docs/source/user_guide/configuration_options.rst
```diff
@@ -62,6 +62,27 @@ Log level
...
+
+.. list-table:: RECEPTOR_PAYLOAD_DEBUG options
+    :header-rows: 1
+    :widths: auto
+
+    * - Debug level
```
AaronH88: I suggest "Tracing Level"

docs/source/user_guide/configuration_options.rst
AaronH88: In general I think it would be a good idea to replace everywhere where you have the word "debug" with the word "trace"

AaronH88 (approved):
lgtm!

PR #1148, opened 2024-09-25
Convert test coverage report to junit format

jessicamack:
To track the test results from receptor we need them to be in a JUnit compatible format. This PR adds https://github.com/jstemmer/go-junit-report to convert the test output into an xml report.

### DIFF
--- a/.github/workflows/test-reporting.yml
+++ b/.github/workflows/test-reporting.yml
@@ -1,0 +1,65 @@
+---
+name: Generate junit test report
+
+on:  # yamllint disable-line rule:truthy
+  pull_request:  # yamllint disable-line rule:empty-values
+  push:
+    branches: [devel]
+
+env:
+  DESIRED_GO_VERSION: '1.20'
+
+jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-reporting.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Download kind binary
+        run: curl -Lo ./kind https://kind.sigs.k8s.io/dl/latest/kind-linux-amd64 && chmod +x ./kind
+
+      - name: Create k8s cluster
+        run: ./kind create cluster
+
+      - name: Interact with the cluster
+        run: kubectl get nodes
+
+      - name: Run receptor tests with coverage
```
AaronH88: Maybe something like this?:

      - name: Install go junit reporting
        run: make go-junit-report

      - name: Run receptor tests
        run: PATH="${PWD}:${PATH}" go test -v 2>&1 ./... | go-junit-report -set-exit-code > report.xml

thom-at-redhat (approved):

PR #1169, opened 2024-10-15
Add patch for arm64 mirrorlist

thom-at-redhat:


AaronH88 (approved):

AaronH88 (changes requested):
maybe we should look at moving away from centos?

PR #1170, opened 2024-10-16
Replace centos stream 9 with ubi9 for container

matoval:
Replace centos stream 9 with ubi9 for container to fix mirror list issue in centos stream 9


AaronH88 (approved):
Lets look at making this more generic in the future and just pin to ubi 9. 
Ok for now to merge

PR #1174, opened 2024-10-16
Update Docker base image to 9.4

matoval:
Update Docker base image to ubi9:9.4


thom-at-redhat (changes requested):
Per the code smell, change from "latest" to "9.4" in both places

thom-at-redhat (approved):

PR #1194, opened 2024-10-24
Create consistent workunit IDs across all nodes

matoval:
Create consistent workunit IDs across all nodes by passing workunit ID to remote node


AaronH88 (approved):
lgtm

Matthew ran integration tests with his changes and the new work unit id's do not break anything with the stack above. 

Good to merge.

PR #1216, opened 2024-11-20
Fix missing last line when kube err

matoval:
This PR shows an issue reading last line when kube api returns an error because the error string doesn't end in a new line. The last line includes eof and never gets read.

I discovered this issue when getting `failed to create fsnotify watcher: too many open files` error in kube logs but receptor never got that error string.

We should look at refactoring way this file gets logs from kube api.


AaronH88 (approved):
lgtm

PR #1221, opened 2024-11-21
Update kube deps to use go 1.22

matoval:
Update the kube api packages to the newest possible with go 1.22 in receptor


AaronH88 (approved):
lgtm

PR #1232, opened 2024-12-02
Make new workunit backward compatible

matoval:
make new workunit backward compatible


AaronH88 (approved):
lgtm

tested locally with 
control (devel) to ee (1.4.9) 
control (1.4.9) to ee (devel)
control (devel) to ee (devel) 
All pass

PR #1257, opened 2025-01-27
Update example config syntax

dleehr:
With documented config at https://ansible.readthedocs.io/projects/receptor/en/devel/getting_started_guide/creating_a_basic_network.html, `receptor` fails to start because it cannot parse the config:

```
$ ./receptor --config foo.yml
Error: error loading config file: yaml: unmarshal errors:
  line 2: cannot unmarshal !!map into []interface {}
```

I updated the structure/syntax, these proceed:

```
$ ./receptor --config foo.yml
...
DEBUG 2025/01/27 16:42:36 foo added service control to listener registry
INFO 2025/01/27 16:42:36 Running control service control
DEBUG 2025/01/27 16:42:36 Running TCP peer connection localhost:2222
INFO 2025/01/27 16:42:36 Connection to baz failed with error: context canceled
INFO 2025/01/27 16:42:36 Initialization complete
...
```

### DIFF
--- a/docs/source/getting_started_guide/creating_a_basic_network.rst
+++ b/docs/source/getting_started_guide/creating_a_basic_network.rst
@@ -24,20 +24,18 @@ Receptor configurations
 
 .. code-block:: yaml
 
-  ---
-  version: 2
-  node:
-    id: foo
+  - node:
+      id: foo
 
-  control-services:
-    - service: control
+  - control-service:
+      service: control
       filename: /tmp/foo.sock
 
-  tcp-peers:
-    - address: localhost:2222
+  - tcp-peer:
+      address: localhost:2222
 
-  log-level:
-    level: debug
+  - log-level:
+      level: debug
 
   ...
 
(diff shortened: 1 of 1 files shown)


matoval:
@dleehr The examples in the documentation are using config v2 although v1 is the default. The current docs are confusing and should be reverted back to v1.
Thank you for your PR 🙏🏽

dleehr:
> @dleehr The examples in the documentation are using config v2 although v1 is the default. The current docs are confusing and should be reverted back to v1. Thank you for your PR 🙏🏽

Thanks for the explanation. I suspected that might be the case, but I couldn't find in the code where it would handle a different version and there seemed to be a mix of old and new syntax in `baz`.

Overall my experience was that the examples in the doc didn't work with the command provided. But if there's a way to tell `receptor` to honor a v2 config, that would certainly be an option too :)

matoval:
> > @dleehr The examples in the documentation are using config v2 although v1 is the default. The current docs are confusing and should be reverted back to v1. Thank you for your PR 🙏🏽
> 
> Thanks for the explanation. I suspected that might be the case, but I couldn't find in the code where it would handle a different version and there seemed to be a mix of old and new syntax in `baz`.
> 
> Overall my experience was that the examples in the doc didn't work with the command provided. But if there's a way to tell `receptor` to honor a v2 config, that would certainly be an option too :)

This command: `./receptor --config foo.yml` should be `./receptor --config --config-v2 foo.yml` if using config v2.

matoval (approved):
LGTM

PR #1261, opened 2025-02-03
Add lock to fix flaky test

matoval:
Add lock to fix flaky test


AaronH88:
Strange, it looks like UntilWithSync is still causing a data race. 🤔

pkg/workceptor/workunitbase.go
```diff
@@ -402,7 +404,9 @@ func (bwu *BaseWorkUnit) MonitorLocalStatus() {
 	watcherErrors = make(chan error)
 
 	if bwu.watcher != nil {
+		watcherLock.Lock()
```
AaronH88: Strange that we need this. 
Have you tried using the existing status file lock?

AaronH88 (approved):
lgtm

PR #1268, opened 2025-02-13
Close all open files, clean up fsnotify watcher

matoval:
Close all open files, clean up fsnotify watcher to fix fsnotify watcher: too many open files bug

### DIFF
--- a/pkg/workceptor/command.go
+++ b/pkg/workceptor/command.go
@@ -124,7 +124,7 @@ func commandRunner(command string, params string, unitdir string) error {
 	defer func() {
 		err := stdin.Close()
 		if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/workceptor/command.go
```diff
@@ -209,6 +209,14 @@ loop:
 			MainInstance.nc.GetLogger().Error("Error updating status file %s: %s", statusFilename, err)
 		}
 	}
+	err = stdin.Close()
```
AaronH88, line 212: Im a bit worried about these two for longer running jobs, 
Have you tested this with either
A job lasting longer than 5 min
Using defer
😄

AaronH88:
Mostly this looks great! And i think it is probably good as it currently is, but as the code around command is both complicated and critical I want to make sure longer jobs and kube have been tested.

AaronH88 (approved):
lgtm

Ty for doing the longer tests

PR #1275, opened 2025-02-28
Add Released status to fix issue when status and release get called at the same time

matoval:
Add Released status to fix issue when status and release get called at the same time


chrismeyersfsu:
I don't think this fixes the issue.

The root cause of the `"Error locating unit:"` message is that the Controller is attempting to check the status of a `work_unit_id` that is no longer recognized by Receptor. This happens because the Controller previously instructed Receptor to release the `work_unit_id`, but it does not account for that when making subsequent status checks.

The proposed solution attempts to mitigate this by extending the window between when a `work_unit_id` is released and when it is no longer recognized by Receptor. However, this does not address the core issue; it merely delays the inevitable failure.

## Alternative Fixes

### 1. Checking Work List Before Status

A potential alternative fix—without modifying Receptor—would be for the Controller to call `work list` before invoking `work status <work_unit_id>` to verify that the `work_unit_id` is still known to Receptor. However, this is not a complete solution either:

- There is still a small race condition where the `work_unit_id` could be released between the `work list` call and the `work status` call.
- If I recall correctly, we previously had a `work list` call but removed it because it was considered expensive. My memory on this is a bit unclear.

### 2. Coordinating `work status` and `work release`

Another possible fix would be to modify the Controller so that the code responsible for calling `work status <work_unit_id>` coordinates with the code that calls `work release <work_unit_id>`. This would help ensure that a `work_unit_id` is not queried for status after it has been released.

This is also not a complete solution:

- At best, it would reduce but not completely eliminate instances where `work status <work_unit_id>` is executed on a `work_unit_id` that no longer exists.
- The Controller cannot always run `work release <work_unit_id>` just once. Imagine the first time the Controller issues this command, the TCP connection is severed right after writing the bits over the Unix domain socket. The Controller will, and should, issue this command again later. What should the code that runs `work status <work_unit_id>` consider the state to be? Should it assume the `<work_unit_id>` still exists, or should it treat it as released?

## 3. Two-Phase

A complete solution would involve a two-phase approach similar to the `work list` followed by `work status`, but with a guarantee that the `work_unit_id` remains valid between the two steps. One possible design:

1. **`work lock <work_unit_id>`**  
   - This would ensure the `work_unit_id` remains valid while checking its status.  
   - If the `work_unit_id` does not exist, it should fail silently (without logging an error in Receptor), otherwise, we end up back at the original problem.  

2. **`work status <work_unit_id>`**  
   - Since the unit is locked, this check should always succeed for valid `work_unit_id`s.

3. **`work unlock <work_unit_id>`** (Optional)  
   - This step releases the lock.  
   - It might not be strictly necessary if `work status` implicitly unlocks after execution.

### Issues with This Approach

While this would eliminate race conditions, it introduces new challenges:

- **What if a work unit is never unlocked?**  
  - If a failure or crash occurs before unlocking, the unit could be stuck indefinitely.  
  - There would need to be a mechanism to clean up orphaned locks.

- **Can a locked work unit be released?**  
  - If a unit is locked, should `work release <work_unit_id>` be blocked until it is unlocked?  
  - If not, what happens if a release occurs while the unit is locked?  

These complications make this solution potentially more problematic than the original issue. It would require additional logic to prevent deadlocks and ensure that locks do not persist indefinitely.

I vote for alternative 1.

matoval:
> I don't think this fixes the issue.
> 
> The root cause of the `"Error locating unit:"` message is that the Controller is attempting to check the status of a `work_unit_id` that is no longer recognized by Receptor. This happens because the Controller previously instructed Receptor to release the `work_unit_id`, but it does not account for that when making subsequent status checks.
> 
> The proposed solution attempts to mitigate this by extending the window between when a `work_unit_id` is released and when it is no longer recognized by Receptor. However, this does not address the core issue; it merely delays the inevitable failure.
> 
> ## Alternative Fixes
> ### 1. Checking Work List Before Status
> A potential alternative fix—without modifying Receptor—would be for the Controller to call `work list` before invoking `work status <work_unit_id>` to verify that the `work_unit_id` is still known to Receptor. However, this is not a complete solution either:
> 
> * There is still a small race condition where the `work_unit_id` could be released between the `work list` call and the `work status` call.
> * If I recall correctly, we previously had a `work list` call but removed it because it was considered expensive. My memory on this is a bit unclear.
> 
> ### 2. Coordinating `work status` and `work release`
> Another possible fix would be to modify the Controller so that the code responsible for calling `work status <work_unit_id>` coordinates with the code that calls `work release <work_unit_id>`. This would help ensure that a `work_unit_id` is not queried for status after it has been released.
> 
> This is also not a complete solution:
> 
> * At best, it would reduce but not completely eliminate instances where `work status <work_unit_id>` is executed on a `work_unit_id` that no longer exists.
> * The Controller cannot always run `work release <work_unit_id>` just once. Imagine the first time the Controller issues this command, the TCP connection is severed right after writing the bits over the Unix domain socket. The Controller will, and should, issue this command again later. What should the code that runs `work status <work_unit_id>` consider the state to be? Should it assume the `<work_unit_id>` still exists, or should it treat it as released?
> 
> ## 3. Two-Phase
> A complete solution would involve a two-phase approach similar to the `work list` followed by `work status`, but with a guarantee that the `work_unit_id` remains valid between the two steps. One possible design:
> 
> 1. **`work lock <work_unit_id>`**
>    
>    * This would ensure the `work_unit_id` remains valid while checking its status.
>    * If the `work_unit_id` does not exist, it should fail silently (without logging an error in Receptor), otherwise, we end up back at the original problem.
> 2. **`work status <work_unit_id>`**
>    
>    * Since the unit is locked, this check should always succeed for valid `work_unit_id`s.
> 3. **`work unlock <work_unit_id>`** (Optional)
>    
>    * This step releases the lock.
>    * It might not be strictly necessary if `work status` implicitly unlocks after execution.
> 
> ### Issues with This Approach
> While this would eliminate race conditions, it introduces new challenges:
> 
> * **What if a work unit is never unlocked?**
>   
>   * If a failure or crash occurs before unlocking, the unit could be stuck indefinitely.
>   * There would need to be a mechanism to clean up orphaned locks.
> * **Can a locked work unit be released?**
>   
>   * If a unit is locked, should `work release <work_unit_id>` be blocked until it is unlocked?
>   * If not, what happens if a release occurs while the unit is locked?
> 
> These complications make this solution potentially more problematic than the original issue. It would require additional logic to prevent deadlocks and ensure that locks do not persist indefinitely.
> 
> I vote for alternative 1.

If we make the changes listed in fix one in Controller and merge this PR in Receptor we shouldn't have a race condition. Controller checks that the work unit is still valid, the time between the work list and work status is the only window when a race condition could happen but Receptor sets the `Released` status during that window.

The only issue with this fix is work list being expensive.

AlanCoding:
I see 3 places where controller runs "work release". Documenting here.

1. in `finally:` block of [running administrative work](https://github.com/ansible/awx/blob/39cd09ce19a1ae1f4efce934f8728135006d22eb/awx/main/tasks/receptor.py#L289)
2. in a [cleanup task for interrupted administrative work](https://github.com/ansible/awx/blob/39cd09ce19a1ae1f4efce934f8728135006d22eb/awx/main/tasks/receptor.py#L224)
3. In the `finally:` block for [finishing a job](https://github.com/ansible/awx/blob/39cd09ce19a1ae1f4efce934f8728135006d22eb/awx/main/tasks/receptor.py#L441)
4. In a [cleanup task for interrupted jobs](https://github.com/ansible/awx/blob/39cd09ce19a1ae1f4efce934f8728135006d22eb/awx/main/tasks/system.py#L731)

You mentioned option (1), which involves getting a list of work before calling "release". This is already the case for my 2 and 4. The only reason we are calling "release" in the first place is because it was showing up in the list and was unexpected.

Adding a pre-check for (1) and/or (3) would be a non-starter for me. That's not the right way to approach that problem. It would insert an additional command in the happy path for running work which has questionable performance implications. It's also not the right way to address this problem.

I'd like to know which specific "work release" commands from AWX are hitting this. If I were to guess... I would venture a guess that we have problems because jobs are _deleted_ very soon after finishing. Because this can happen very soon after a job finishes, particularly for small jobs. By deleting the job, we lose any memory of when the job finished, so we can't have any sort of grace period to avoid these conflicts (because we lost the record). Running a job and quickly deleting it is an odd thing to do... unless we're in a test suite. Tests do this all the time, even though it's very low importance for any real users.

AlanCoding:
Staring at this code again:

```python
    jobs_with_unreleased_receptor_units = UnifiedJob.objects.filter(work_unit_id__in=unit_ids).exclude(status__in=ACTIVE_STATES)
    if settings.RECEPTOR_KEEP_WORK_ON_ERROR:
        jobs_with_unreleased_receptor_units = jobs_with_unreleased_receptor_units.exclude(status__in=ERROR_STATES)
    for job in jobs_with_unreleased_receptor_units:
        logger.debug(f"{job.log_format} is not active, reaping receptor work unit {job.work_unit_id}")
        receptor_ctl.simple_command(f"work cancel {job.work_unit_id}")
        receptor_ctl.simple_command(f"work release {job.work_unit_id}")
```

I want to note that there is _no_ grace period. We would welcome adding a filter to the `UnifiedJob.objects` queryset here to only get jobs that finished (error or successful) so-many seconds or longer ago. This is virtually standard practice in other cases and seems like an oversight here.

Such a grace period would almost completely avoid the contention possible, where both (3) and (4) mechanisms both try to release the same work unit. The `finally:` block for a job can be reasonably expected to run within a fixed time after the job execution finishing (or else, never run).

chrismeyersfsu:
> Running a job and quickly deleting it is an odd thing to do

Unfortunately, customers do this often.

chrismeyersfsu:
Thank you @matoval and @AlanCoding for your engagement and conversation. I should have prefaced my message with "I think this is how things work". This is all helping understand the _true_ behavior(s).


```
ERROR 2025/03/20 08:29:00 Error locating unit: bazoJrlxafG
ERROR 2025/03/20 08:29:00 : unknown work unit bazoJrlxafG
```

The addition of the `Released` state is a good thing. Receptor will not report the above "error" while the work unit is still _actually_ known to Receptor.

I'm still not convinced that AWX's usage of Receptor won't result in the above error messages happening.

For what it's worth, I played around with this PR and it seems to work. I can observe work being in the `Released` status after I've requested it to be released. Some time later, I can observe that absence of the work from `work list`.

I'm a :+1: for the addition of this new state.

AlanCoding:
Just to wrap up this conversation, I need to reverse something I said earlier.

Deleted jobs may still be a concern, but so minor of a concern that I don't honestly care. Because the point of the snippet I gave earlier was to say that we create **instantaneous** contention of releasing work unit even for the _happy path_ of normal jobs.

Here is a draft of the AWX patch that should work to reduce that window of contention.

```diff
diff --git a/awx/main/tasks/system.py b/awx/main/tasks/system.py
index be6f56fa8d..34eed9cf9f 100644
--- a/awx/main/tasks/system.py
+++ b/awx/main/tasks/system.py
@@ -730,7 +730,9 @@ def awx_receptor_workunit_reaper():
     receptor_work_list = receptor_ctl.simple_command("work list")
 
     unit_ids = [id for id in receptor_work_list]
-    jobs_with_unreleased_receptor_units = UnifiedJob.objects.filter(work_unit_id__in=unit_ids).exclude(status__in=ACTIVE_STATES)
+    # Only consider jobs with a final status that finished so-many seconds ago or before
+    time_cutoff = now() - timedelta(seconds=settings.WORK_UNIT_CLEANUP_GRACE_PERIOD)
+    jobs_with_unreleased_receptor_units = UnifiedJob.objects.filter(work_unit_id__in=unit_ids, finished__lte=time_cutoff).exclude(status__in=ACTIVE_STATES)
     if settings.RECEPTOR_KEEP_WORK_ON_ERROR:
         jobs_with_unreleased_receptor_units = jobs_with_unreleased_receptor_units.exclude(status__in=ERROR_STATES)
     for job in jobs_with_unreleased_receptor_units:
diff --git a/awx/settings/defaults.py b/awx/settings/defaults.py
index 48c4f775b9..4e4ed9cffe 100644
--- a/awx/settings/defaults.py
+++ b/awx/settings/defaults.py
@@ -889,6 +889,7 @@ AWX_RUNNER_KEEPALIVE_SECONDS = 0
 # Delete completed work units in receptor
 RECEPTOR_RELEASE_WORK = True
 RECEPTOR_KEEP_WORK_ON_ERROR = False
+WORK_UNIT_CLEANUP_GRACE_PERIOD = 60
 
 # K8S only. Use receptor_log_level on AWX spec to set this properly
 RECEPTOR_LOG_LEVEL = 'info'
```

If we do this, and if a job is not deleted right after finishing, and the worst-case happened, we would have to spend >1 min between changing the status of the job to "successful" and wrapping up the work. For all intents and purposes I would consider this solved.

However, it would still not solve the case of administrative work units, and I don't know if there's a similar method we could use for a grace period for those.

pkg/workceptor/controlsvc.go
```diff
@@ -391,6 +392,8 @@ func (c *workceptorCommand) ControlFunc(ctx context.Context, nc controlsvc.Netce
 		if c.subcommand == "cancel" {
 			err = unit.Cancel()
 		} else {
+			unit.UpdateBasicStatus(5, "released work", 0)
+			time.Sleep(time.Second * 3)
```
matoval, line 396: @AaronH88 How long after setting the status to release should we wait before removing the work units dirs? This will block the work release command from completing. I'm not sure if 3 seconds is long enough but I also don't want the release command to take too long.

matoval:
> Just to wrap up this conversation, I need to reverse something I said earlier.
> 
> Deleted jobs may still be a concern, but so minor of a concern that I don't honestly care. Because the point of the snippet I gave earlier was to say that we create **instantaneous** contention of releasing work unit even for the _happy path_ of normal jobs.
> 
> Here is a draft of the AWX patch that should work to reduce that window of contention.
> 
> ```diff
> diff --git a/awx/main/tasks/system.py b/awx/main/tasks/system.py
> index be6f56fa8d..34eed9cf9f 100644
> --- a/awx/main/tasks/system.py
> +++ b/awx/main/tasks/system.py
> @@ -730,7 +730,9 @@ def awx_receptor_workunit_reaper():
>      receptor_work_list = receptor_ctl.simple_command("work list")
>  
>      unit_ids = [id for id in receptor_work_list]
> -    jobs_with_unreleased_receptor_units = UnifiedJob.objects.filter(work_unit_id__in=unit_ids).exclude(status__in=ACTIVE_STATES)
> +    # Only consider jobs with a final status that finished so-many seconds ago or before
> +    time_cutoff = now() - timedelta(seconds=settings.WORK_UNIT_CLEANUP_GRACE_PERIOD)
> +    jobs_with_unreleased_receptor_units = UnifiedJob.objects.filter(work_unit_id__in=unit_ids, finished__lte=time_cutoff).exclude(status__in=ACTIVE_STATES)
>      if settings.RECEPTOR_KEEP_WORK_ON_ERROR:
>          jobs_with_unreleased_receptor_units = jobs_with_unreleased_receptor_units.exclude(status__in=ERROR_STATES)
>      for job in jobs_with_unreleased_receptor_units:
> diff --git a/awx/settings/defaults.py b/awx/settings/defaults.py
> index 48c4f775b9..4e4ed9cffe 100644
> --- a/awx/settings/defaults.py
> +++ b/awx/settings/defaults.py
> @@ -889,6 +889,7 @@ AWX_RUNNER_KEEPALIVE_SECONDS = 0
>  # Delete completed work units in receptor
>  RECEPTOR_RELEASE_WORK = True
>  RECEPTOR_KEEP_WORK_ON_ERROR = False
> +WORK_UNIT_CLEANUP_GRACE_PERIOD = 60
>  
>  # K8S only. Use receptor_log_level on AWX spec to set this properly
>  RECEPTOR_LOG_LEVEL = 'info'
> ```
> 
> If we do this, and if a job is not deleted right after finishing, and the worst-case happened, we would have to spend >1 min between changing the status of the job to "successful" and wrapping up the work. For all intents and purposes I would consider this solved.
> 
> However, it would still not solve the case of administrative work units, and I don't know if there's a similar method we could use for a grace period for those.

This seems like a reasonable solution.

I realize this PR won't fix the root of the issue but will avoid the `unknown work unit` when calling status and release at the same time.

Should I create an awx PR for handling the new status of `Released` this way we can merge this PR with that one?

pkg/workceptor/controlsvc.go
AaronH88: Lets make it configurable with a default of 3 seconds?

pkg/workceptor/workunitbase.go
```diff
@@ -497,15 +502,14 @@ func (bwu *BaseWorkUnit) Status() *StatusFileData {
...
-	defer bwu.statusLock.RUnlock()
+	status := bwu.getStatus()
+	bwu.statusLock.RUnlock()
 
-	return bwu.getStatus()
+	return status
 }
 
-// Release releases this unit of work, deleting its files.
+// / Release releases this unit of work, deleting its files.
```
AaronH88, line 412: extra / ?

AaronH88:
Overall this code is looking very clean. 

One comment on making the delay configurable, but this is looking great, good work

pkg/workceptor/workunitbase.go
```diff
@@ -417,7 +421,6 @@ func (bwu *BaseWorkUnit) Release(force bool) error {
 
 			if attemptsLeft > 0 {
 				bwu.w.nc.GetLogger().Warning("Error removing directory for %s. Retrying %d more times.", bwu.unitID, attemptsLeft)
-				time.Sleep(time.Second)
```
arrestle, line 424: just curious, why are we removing the sleep here without a watch? Or am I missing something?

pkg/workceptor/workunitbase.go
```diff
@@ -331,9 +336,9 @@ func (sfd *StatusFileData) UpdateBasicStatus(filename string, state int, detail
 // Passing -1 as stdoutSize leaves it unchanged.
 func (bwu *BaseWorkUnit) UpdateBasicStatus(state int, detail string, stdoutSize int64) {
 	bwu.statusLock.Lock()
-	defer bwu.statusLock.Unlock()
-
 	err := bwu.status.UpdateBasicStatus(bwu.statusFileName, state, detail, stdoutSize)
+	bwu.statusLock.Unlock()
```
arrestle, line 340: Should we change this to a defer so that both the `BasicStatus` and the `lastUpdateError` stay in sync?

arrestle (changes requested):
Looks good, overall, really I just have some questions.

pkg/workceptor/workunitbase.go
```diff
@@ -245,9 +249,10 @@ func (sfd *StatusFileData) Load(filename string) error {
 // Load loads status from a file.
 func (bwu *BaseWorkUnit) Load() error {
```
davemulford, line 250: It looks like this just removes the `defer` call but otherwise leaves the functionality intact. Is there something wrong with using `defer`? I notice the same thing in other parts of the code.

davemulford (approved):
A single question about the removal of a defer in a few places. Otherwise LGTM.

PR #1278, opened 2025-03-04
Add unit tests to internal version package

lranjbar:
Extremely small refactor to make it easier to capture the output for version here.

### DIFF
--- a/internal/version/version_test.go
+++ b/internal/version/version_test.go
@@ -1,0 +1,37 @@
+package version
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestValidateVersion(t *testing.T) {
+	type test struct {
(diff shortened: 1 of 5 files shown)


internal/version/version_test.go
```diff
@@ -0,0 +1,37 @@
+package version
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
matoval, line 6: To fix the depgaurd error you need to add new packages to this file: https://github.com/ansible/receptor/blob/devel/.golangci.yml#L84

lranjbar: Thanks! :)

matoval (approved):
lgtm

AaronH88 (approved):
lgtm

PR #1284, opened 2025-03-17
Add unit tests for the tickrunner.go package

fincamd:


fincamd:
I'm not sure how I should be testing whether the inner function gets called exactly in the times we set as parameters, but the 100% coverage is there. Should I work on mocking a struct/interface that I can expect calls from?

matoval:
> I'm not sure how I should be testing whether the inner function gets called exactly in the times we set as parameters, but the 100% coverage is there. Should I work on mocking a struct/interface that I can expect calls from?

A good way to test the inner function is to create a count func that counts how many times the inner function was executed. We should probably test the functionality of calling the inner function close to the same time but only executing once.

AaronH88 (approved):
lgtm 

+1 out of 2

matoval (approved):
lgtm

PR #1290, opened 2025-03-26
Add unit tests to command service

lranjbar:
Adding unit tests to the command service.

Summary of changes:
- Added mocks for command service
- Changes to netceptor/conn.go 
  - I made them the same as Daniel's PR https://github.com/ansible/receptor/pull/1294. We needed to expose the accept channel and done channel to pass values on them in the tests.
- Found a bug with runCommand(). If the empty string is passed to it, it causes a panic on [line 25](https://github.com/ansible/receptor/pull/1290/files#diff-69780efc7cd9cb4d7d4c29e686bfa3ba2787aa5f33cc1887e89b746523e7e44eL25). I added error checking on [line 45](https://github.com/ansible/receptor/pull/1290/files#diff-69780efc7cd9cb4d7d4c29e686bfa3ba2787aa5f33cc1887e89b746523e7e44eR45) to cover this case. Note that this is not reachable from Run() because it requires the command to be populated.
- Changes to mock_services/tcp_proxy.go are from the go generate command.
- Added malformed command test


AaronH88:
Just an tip as im reviewing this, 
Moving the tls certs in the PR is making it huge, 
Can we please split the tls work into another PR 😄

pkg/services/command.go
```diff
@@ -1,58 +1,74 @@
...
+	ListenAndAdvertise(service string, tlscfg *tls.Config, tags map[string]string) (*netceptor.Listener, error)
+}
+
+func runCommand(qc net.Conn, command string, logger *logger.ReceptorLogger, utilsLib UtilsLib) error {
+	// Note: shlex.Split does not return error for the empty string
 	args, err := shlex.Split(command)
 	if err != nil {
 		return err
 	}
+	if len(args) == 0 {
```
AaronH88, line 32: Nice!

pkg/services/command.go
```diff
@@ -1,58 +1,74 @@
...
+		return errors.New("command is malformed")
```
AaronH88: Perhaps, command is empty?

pkg/services/command.go
```diff
@@ -1,58 +1,74 @@
...
+	utilsLib.BridgeConns(tty, "external command", qc, "command service", logger)
 
 	return nil
 }
 
 // CommandService listens on the Receptor network and runs a local command.
-func CommandService(s *netceptor.Netceptor, service string, tlscfg *tls.Config, command string) {
+func CommandService(s NetCForCommandService, service string, tlscfg *tls.Config, command string, utilsLib UtilsLib) {
+	if command == "" {
+		s.GetLogger().Error("error initializing command service: %s\n", errors.New("command not provided"))
```
AaronH88: The Error() function in the logger already prefixes message, so we dont need to start our sentence with error. Also, as the logger can take any number of arguments, we dont need to create a new "errors.New()" error here, just the message will be enough 😄

AaronH88:
Mostly looking good, just a few smaller things but all the logic is great. 

Good work on this

pkg/services/command.go
lranjbar: There are a couple of ways to get here but mostly its that `shlex.split()` can return an empty array. 

You can get here by just sending in commented shell commands because it will strip out the all the comments in your shell command.  Let me add more verbiage here :)

pkg/services/command_test.go
```diff
@@ -0,0 +1,101 @@
...
+		},
+		// This is a case where shlex.Split() returns an empty array
+		// In order to stop execution of the CommandService() in this test we pass an error on the AcceptChannel
+		{
+			name:         "Malformed command",
+			acceptResult: []*netceptor.AcceptResult{{Conn: mockTCPConn, Err: nil}, {Conn: nil, Err: errors.New("failed to accept connection")}},
+			command:      "# nine # ten\n",
+			calls: func() {
+				mockNetceptor.EXPECT().ListenAndAdvertise(gomock.Any(), gomock.Any(), gomock.Any()).Return(&neceptorListener, nil)
+				mockTCPConn.EXPECT().Close().Times(1)
```
matoval: The `Times(1)` here is causing the service package to fail because tcp_proxy calls this same mock but it is limited to be only called once. Setting it to `Anytimes()` does fix it but there may be a better solution.

matoval (changes requested):
Tests look good, just need to fix the issue with `mockTCPConn.Close`

pkg/services/command_test.go
lranjbar: Yeah this was originally AnyTimes() in here and both tcp_proxy. I'm reverting this for now

matoval (approved):
lgtm

AaronH88 (approved):
lgtm 2 of 2

PR #1291, opened 2025-03-27
Add more unit tests for config run function in tcp_proxy.go

fincamd:


pkg/services/tcp_proxy_test.go
```diff
@@ -196,3 +197,104 @@ func TestTCPProxyServiceOutbound(t *testing.T) {
 		})
 	}
 }
+
+func TestTCPProxyInboundCfgRun(t *testing.T) {
+	type testCoverageItem struct {
```
matoval, line 201: We should add the specific error we expect to get. This way we know the specific case is getting the error we expect and not a different error.

pkg/services/tcp_proxy_test.go
```diff
@@ -196,3 +197,104 @@ func TestTCPProxyServiceOutbound(t *testing.T) {
...
+			err := tc.configObj.Run()
+			if !tc.expectError && err != nil {
+				t.Errorf("This test case wasn't expected to return an error")
+			}
+		})
+	}
+}
+
+func TestTCPProxyOutboundCfgRun(t *testing.T) {
+	type testCoverageItem struct {
```
matoval, line 258: Same as the previous test, we should add the error we expect to get.

matoval (changes requested):
Tests look good, recommend to check the specific error.

fincamd:
Thanks for the review @matoval, your suggestions are applied in [this commit](https://github.com/ansible/receptor/pull/1291/commits/303573007ce2c0e162a832469298d15c26567a3a)

AaronH88 (approved):
lgtm

1 out of 2

matoval (approved):
lgtm

PR #1293, opened 2025-04-01
Fix coverage of TestTCPProxyServiceOutbound function in tcp_proxy.go

fincamd:
I've added some changes as suggested by @AaronH88 to increase missing coverage in the tcp_proxy.go package. This was made by leveraging a custom Listener object that contains channels we can use to communicate with the goroutine during test execution


fincamd:
Closed in favor of #1294

PR #1297, opened 2025-04-03
Migrate mock generation to go generate

lranjbar:
Migrating the mock generation in receptor so all mocks can be generated using the command `go generate ./...` from the base directory.


pkg/receptortest/net.go
matoval, line 1: What's the purpose of creating receptortest?

lranjbar: To have a package where common test structures (like generalized mocks) can be shared between packages. I know that the tcp_proxy tests and command tests also use a mock for net.Conn in the services packages. 

I used this as an opportunity to create a `receptortest` package but didn't make changes to tcp_proxy or commands tests to limit the change scope in this PR. 

Having a `receptortest` package also gives us a place for things that might be defined in a _test.go file but we want to export to another package. In Golang you cannot export things defined in _test.go files to another package.

matoval (approved):
lgtm

pkg/receptortest/net.go
AaronH88: I too am finding this `receptortest` package at odds with the structure of our codebase. 

I understand the desire to to reduce code repetition, however, readability is equally if not more important to me. 

To that end, I have some suggestions around what you are trying to do. 

The code has a few packages that could be good candidates for something like this, both `pkg/service` and `pkg/utils` come to mind. 

Also to improve readability at a folder level, we have been using the `mock` prefix as well as the `_interface` postfix, both of which you will notice have no impact on their ability to be imported. 

A good example of this can be found in `pkg/services/interfaces/*` 

I suggest re_working this PR to improve readability please 😄

pkg/certificates/mock_certificates/Oser.go
```diff
@@ -1,69 +0,0 @@
-// Code generated by MockGen. DO NOT EDIT.
```
AaronH88, line 1: Im not too sure about combining Oser and Rsaer and into a common mock. They have no connection to each other 🤔

pkg/backends/websockets.go
```diff
@@ -1,3 +1,4 @@
+//go:generate mockgen -source=websockets.go -destination=mock_backends/websockets.go
```
AaronH88: Has this line been auto created using go generate? 

If yes, is their anyway we can get the full dir and not the local dir in this comment?

pkg/controlsvc/mock_controlsvc/io.go
```diff
@@ -1,84 +0,0 @@
-// Code generated by MockGen. DO NOT EDIT.
```
AaronH88, line 1: Nice, 

I love it when we can delete code

pkg/certificates/mock_certificates/Oser.go
lranjbar: Migrated into two different files to separate them in the code. :)

lranjbar:
@PabloHiro Will the changes in your PR #1325 help tox pull in the correct version of click? 

@AaronH88 @matoval [click version 8.2.0](https://click.palletsprojects.com/en/stable/changes/#:~:text=Keep%20stdout%20and%20stderr%20streams%20independent%20in%20CliRunner.%20Always%20collect%20stderr%20output%20and%20never%20raise%20an%20exception.%20Add%20a%20new%20output%20stream%20to%20simulate%20what%20the%20user%20sees%20in%20its%20terminal.%20Removes%20the%20mix_stderr%20parameter%20in%20CliRunner) has made changes to the CliRunner. Looks like we've got more tech debt. 🤔  Still unclear why this PR has made this issue apparent. I would expect this would hit all PRs.

matoval (approved):
lgtm

arrestle (approved):
LGTM

PR #1299, opened 2025-04-11
feat: field ARCH in Makefile shall be optional

PabloHiro:
Making the field ARCH in the Makefile optional so the `make build-package` can be used for non x86_64 architectures.

AAP-43006


AaronH88 (approved):
lgtm

1 of 2

davemulford:
LGTM

PR #1300, opened 2025-04-11
AAP-43201 netceptor logging

arrestle:
This PR is ready to review. 

While it doesn't implement structured logging, it's organized in such a way as to make that easier to add in AAP-44416: Implement dynamic log level changes.

Note that additional diagnostics were added to identify existing connections with initial handshake errors, but does *not* remove them as  that is a bug that is out of scope.  See [PR#1303](https://github.com/ansible/receptor/pull/1303) for details. 

Note that several iterations later, we've decided to leave the logging in netceptor.go the way it is,  just adding a few more log diagnostics and messages. The ability to add suffixes was added the logger with the addition of AddSuffix and UpdateSuffix to the logger itself. All future logs will contain what is in the log suffix.

Example output. See at 15:48:21 I stopped `hop`  and restarted it a few seconds later.
```json
INFO 2025/05/07 15:48:17 Known Connections: {"node_id":"execution","remote_id":"hop"}
INFO 2025/05/07 15:48:17    execution:  {"node_id":"execution","remote_id":"hop"}
INFO 2025/05/07 15:48:17    hop: control(1.00)  {"node_id":"execution","remote_id":"hop"}
INFO 2025/05/07 15:48:17    control: hop(1.00)  {"node_id":"execution","remote_id":"hop"}
INFO 2025/05/07 15:48:17 Routing Table: {"node_id":"execution","remote_id":"hop"}
WARNING 2025/05/07 15:48:21 Backend connection failed (will retry): dial tcp [::1]:2224: connect: connection refused {"remote_id":"hop","node_id":"execution"}
INFO 2025/05/07 15:48:29 Connection established with hop {"node_id":"execution","remote_id":"hop"}
INFO 2025/05/07 15:48:29 Known Connections: {"node_id":"execution","remote_id":"hop"}
INFO 2025/05/07 15:48:29    execution: hop(1.00)  {"node_id":"execution","remote_id":"hop"}
INFO 2025/05/07 15:48:29    hop: control(1.00) execution(1.00)  {"node_id":"execution","remote_id":"hop"}
INFO 2025/05/07 15:48:29    control: hop(1.00)  {"node_id":"execution","remote_id":"hop"}
```

### DIFF
--- a/pkg/backends/tcp.go
+++ b/pkg/backends/tcp.go
@@ -62,6 +62,10 @@ func (b *TCPDialer) Start(ctx context.Context, wg *sync.WaitGroup) (chan netcept
 				return nil, err
 			}
 
+			if b.logger != nil {
+				b.logger.Debug("TCPDialer connected to TCP %s Address %s\n", b.address, conn.RemoteAddr().String())
+			}
+
 			return newTCPSession(conn, closeChan), nil
 		})
 }
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -254,7 +254,7 @@ func (rl *ReceptorLogger) Log(level int, format string, v ...interface{}) {
 	}
 }
 
-func (rl *ReceptorLogger) appendSuffix(format string, v []interface{}) (string, []interface{}) {
+func (rl *ReceptorLogger) appendSuffix(format string, v []string) (string, []string) {
 	if rl.Suffix != nil {
 		rl.m.Lock()
 		defer rl.m.Unlock()
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -99,7 +99,7 @@ func NewReceptorLogger(prefix string) *ReceptorLogger {
 	}
 }
 
-// NewReceptorLoggerWithSuffix to instantiate a new logger object with a new Suffix
+// NewReceptorLoggerWithSuffix to instantiate a new logger object with a new Suffix.
 func NewReceptorLoggerWithSuffix(prefix string, suffix map[string]string) *ReceptorLogger {
 	logger := NewReceptorLogger(prefix)
 	logger.SetSuffix(suffix)
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -88,6 +88,7 @@ type ReceptorLogger struct {
 	log.Logger
 	Prefix string
 	m      sync.Mutex
+	Suffix map[string]string
 }
 
 // NewReceptorLogger to instantiate a new logger object.
@@ -95,6 +96,30 @@ func NewReceptorLogger(prefix string) *ReceptorLogger {
 	return &ReceptorLogger{
 		Logger: *log.New(os.Stdout, prefix, log.LstdFlags),
 		Prefix: prefix,
+	}
+}
+
+func NewReceptorLoggerWithSuffix(prefix string, suffix map[string]string) *ReceptorLogger {
+	logger := NewReceptorLogger(prefix)
+	logger.SetSuffix(suffix)
+
+	return logger
+}
+
+func (rl *ReceptorLogger) SetSuffix(suffix map[string]string) {
+	rl.m.Lock()
+	defer rl.m.Unlock()
+	rl.Suffix = suffix
+}
+
+func (rl *ReceptorLogger) UpdateSuffix(suffix map[string]string) {
+	rl.m.Lock()
+	defer rl.m.Unlock()
+	if rl.Suffix == nil {
+		rl.Suffix = make(map[string]string)
+	}
+	for k, v := range suffix {
+		rl.Suffix[k] = v
 	}
 }
 
@@ -218,8 +243,37 @@ func (rl *ReceptorLogger) Log(level int, format string, v ...interface{}) {
 
 	if logLevel >= level {
 		rl.Logger.SetPrefix(prefix)
+		format, v = rl.appendSuffix(format, v)
 		rl.Logger.Printf(format, v...)
 	}
+}
+
+func (rl *ReceptorLogger) appendSuffix(format string, v []interface{}) (string, []interface{}) {
+	if rl.Suffix != nil {
+		rl.m.Lock()
+		defer rl.m.Unlock()
+
+		var b strings.Builder
+		b.WriteString("{")
+		first := true
+		for k, val := range rl.Suffix {
+			if !first {
+				b.WriteString(",")
+			}
+			first = false
+			b.WriteString(`"`)
+			b.WriteString(k)
+			b.WriteString(`":"`)
+			b.WriteString(val)
+			b.WriteString(`"`)
+		}
+		b.WriteString("}")
+
+		format = strings.ReplaceAll(format, "\n", "") + " %s"
+		v = append(v, b.String())
+	}
+
+	return format, v
 }
 
 // SanitizedLog adds a prefix and prints a given log message.
@@ -245,7 +299,8 @@ func (rl *ReceptorLogger) SanitizedLog(level int, format string, v ...interface{
 	}
 
 	if logLevel >= level {
-		message := fmt.Sprintf(format, v...)
+		message, v := rl.appendSuffix(format, v)
+		message = fmt.Sprintf(message, v...)
 		sanMessage := strings.ReplaceAll(message, "\n", "")
 		rl.Logger.SetPrefix(prefix)
 		rl.Logger.Print(sanMessage)
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -254,7 +254,7 @@ func (rl *ReceptorLogger) Log(level int, format string, v ...interface{}) {
 	}
 }
 
-func (rl *ReceptorLogger) appendSuffix(format string, v []string) (string, []string) {
+func (rl *ReceptorLogger) appendSuffix(format string, v []interface{}) (string, []interface{}) {
 	if rl.Suffix != nil {
 		rl.m.Lock()
 		defer rl.m.Unlock()
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -99,6 +99,7 @@ func NewReceptorLogger(prefix string) *ReceptorLogger {
 	}
 }
 
+// NewReceptorLoggerWithSuffix to instantiate a new logger object with a new Suffix
 func NewReceptorLoggerWithSuffix(prefix string, suffix map[string]string) *ReceptorLogger {
 	logger := NewReceptorLogger(prefix)
 	logger.SetSuffix(suffix)
@@ -106,12 +107,17 @@ func NewReceptorLoggerWithSuffix(prefix string, suffix map[string]string) *Recep
 	return logger
 }
 
+// SetSuffix sets the suffix for the logger, overwriting any existing suffix.
 func (rl *ReceptorLogger) SetSuffix(suffix map[string]string) {
 	rl.m.Lock()
 	defer rl.m.Unlock()
 	rl.Suffix = suffix
 }
 
+// UpdateSuffix allows adding suffix key/value pairs to an existing suffix.
+// If the key already exists, the value will be overwritten.
+// If the suffix is nil, it will be initialized.
+// This is useful for adding additional context to log messages when you want to keep the existing suffix key/value pairs.
 func (rl *ReceptorLogger) UpdateSuffix(suffix map[string]string) {
 	rl.m.Lock()
 	defer rl.m.Unlock()
--- a/pkg/logger/logger_test.go
+++ b/pkg/logger/logger_test.go
@@ -134,7 +134,7 @@ func TestGetLoggerWithSuffix(t *testing.T) {
 		receptorLogger := logger.NewReceptorLoggerWithSuffix("", suffix)
 		receptorLogger.SetOutput(&logBuffer)
 
-		receptorLogger.Error(testname)
+		receptorLogger.Error("%s", testname)
 		if !strings.Contains(logBuffer.String(), testname) {
 			t.Errorf("expected log message %s not found in log: %s", testname, logBuffer.String())
 		}
@@ -154,7 +154,7 @@ func TestGetLoggerWithSuffix(t *testing.T) {
 			"cost": "12",
 		}
 		receptorLogger.UpdateSuffix(updated)
-		receptorLogger.SanitizedError(testname)
+		receptorLogger.SanitizedError("%s", testname)
 
 		if !strings.Contains(logBuffer.String(), testname) {
 			t.Errorf("expected log message %s not found in log: %s", testname, logBuffer.String())
--- a/pkg/logger/logger_test.go
+++ b/pkg/logger/logger_test.go
@@ -3,7 +3,6 @@ package logger_test
 import (
 	"bytes"
 	"fmt"
-	"os"
 	"strings"
 	"testing"
 
@@ -73,14 +72,10 @@ func TestLogLevelToNameWithError(t *testing.T) {
 }
 
 func TestDebugPayload(t *testing.T) {
-	logFilePath := "/tmp/test-output"
+	var logBuffer bytes.Buffer
 	logger.SetGlobalLogLevel(4)
 	receptorLogger := logger.NewReceptorLogger("testDebugPayload")
-	logFile, err := os.OpenFile(logFilePath, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0o600)
-	if err != nil {
-		t.Error("error creating test-output file")
-	}
-
+	receptorLogger.SetOutput(&logBuffer)
 	payload := "Testing debugPayload"
 	workUnitID := "1234"
 	connectionType := "unix socket"
@@ -105,56 +100,67 @@ func TestDebugPayload(t *testing.T) {
 
 	for _, testCase := range debugPayloadTestCases {
 		t.Run(testCase.name, func(t *testing.T) {
-			receptorLogger.SetOutput(logFile)
 			receptorLogger.DebugPayload(testCase.debugPayload, testCase.payload, testCase.workUnitID, testCase.connectionType)
+			testOutput := logBuffer.Bytes()
 
-			testOutput, err := os.ReadFile(logFilePath)
-			if err != nil {
-				t.Error("error reading test-output file")
-			}
-			if !bytes.Contains(testOutput, []byte(testCase.expectedLog)) {
+			if !strings.Contains(string(testOutput), testCase.expectedLog) {
 				t.Errorf("failed to log correctly, expected: %v got %v", testCase.expectedLog, string(testOutput))
 			}
-			if err := os.Truncate(logFilePath, 0); err != nil {
-				t.Errorf("failed to truncate: %v", err)
-			}
+			logBuffer.Reset()
 		})
 	}
 }
 
-func TestGetLoggerWithSuffix(t *testing.T) {
-	var logBuffer bytes.Buffer
+func assertSuffixFieldsPresent(t *testing.T, logLine string, expected map[string]string) {
+	t.Helper()
+	for k, v := range expected {
+		needle := `"` + k + `":"` + v + `"`
+		if !strings.Contains(logLine, needle) {
+			t.Errorf("expected key-value pair %s not found in log: %s", needle, logLine)
+		}
+	}
+}
 
+func TestGetLoggerWithSuffix(t *testing.T) {
 	logger.SetGlobalLogLevel(4)
 
-	suffix := map[string]string{
-		"node_id":   "controller",
-		"remote_id": "hop",
-	}
-	receptorLogger := logger.NewReceptorLoggerWithSuffix("", suffix)
-	receptorLogger.SetOutput(&logBuffer)
-	expectedLog := `example message 1 {"node_id":"controller","remote_id":"hop"}`
-	receptorLogger.Error("example message 1")
-	testOutput := logBuffer.Bytes()
+	t.Run("initial suffix", func(t *testing.T) {
+		var logBuffer bytes.Buffer
+		testname := "Initial Suffix Example"
+		suffix := map[string]string{
+			"node_id":   "controller",
+			"remote_id": "hop",
+		}
+		receptorLogger := logger.NewReceptorLoggerWithSuffix("", suffix)
+		receptorLogger.SetOutput(&logBuffer)
 
-	if !strings.Contains(logBuffer.String(), expectedLog) {
-		t.Errorf("failed to log correctly, expected: %v got %v", expectedLog, string(testOutput))
-	}
+		receptorLogger.Error(testname)
+		if !strings.Contains(logBuffer.String(), testname) {
+			t.Errorf("expected log message %s not found in log: %s", testname, logBuffer.String())
+		}
+		assertSuffixFieldsPresent(t, logBuffer.String(), suffix)
+	})
+	t.Run("updated suffix", func(t *testing.T) {
+		var logBuffer bytes.Buffer
+		testname := "Updated Suffix Example"
+		suffix := map[string]string{
+			"node_id":   "controller",
+			"remote_id": "hop",
+		}
+		receptorLogger := logger.NewReceptorLoggerWithSuffix("", suffix)
+		receptorLogger.SetOutput(&logBuffer)
 
-	logBuffer.Reset()
+		updated := map[string]string{
+			"cost": "12",
+		}
+		receptorLogger.UpdateSuffix(updated)
+		receptorLogger.SanitizedError(testname)
 
-	suffix = map[string]string{
-		"node_id":   "controller",
-		"remote_id": "hop",
-		"cost":      "12",
-	}
-	receptorLogger.UpdateSuffix(suffix)
-	expectedLog = `example message 2 {"node_id":"controller","remote_id":"hop","cost":"12"}` // note fields comes out alphabetically
-	receptorLogger.SanitizedError("example message 2")
-	testOutput = logBuffer.Bytes()
+		if !strings.Contains(logBuffer.String(), testname) {
+			t.Errorf("expected log message %s not found in log: %s", testname, logBuffer.String())
+		}
 
-	if !strings.Contains(logBuffer.String(), expectedLog) {
-		t.Errorf("failed to log correctly, expected: %v got %v", expectedLog, string(testOutput))
-	}
-	logBuffer.Reset()
+		assertSuffixFieldsPresent(t, logBuffer.String(), suffix)
+		assertSuffixFieldsPresent(t, logBuffer.String(), updated)
+	})
 }
--- a/pkg/logger/logger_test.go
+++ b/pkg/logger/logger_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/ansible/receptor/pkg/logger"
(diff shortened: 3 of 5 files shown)


.gitignore
```diff
@@ -29,4 +29,5 @@ venv
 /vendor
 pkg/services/.lock
 pkg/workceptor/status
-pkg/workceptor/status.lock
\ No newline at end of file
+pkg/workceptor/status.lock
+cmd/receptor-cl
```
matoval: Looks like cmd/receptor-cl can't be in the .gitignore

arrestle: Thanks  Matthew!

pkg/logger/logger.go
```diff
@@ -98,6 +100,30 @@ func NewReceptorLogger(prefix string) *ReceptorLogger {
 	}
 }
 
+func NewReceptorLoggerWithPostfix(prefix string, postfix map[string]interface{}) *ReceptorLogger {
```
PabloHiro: `Postfix`-> `Suffix`?

Postfix is a Linux utility to send emails, this naming might induce confusion.

arrestle: Done.

pkg/netceptor/netceptor_test.go
```diff
@@ -894,3 +895,76 @@ func TestTracerDoesNotReturnsNewConnectionTracer(t *testing.T) {
...
+func testNetceptorLogMethod(t *testing.T, logFunc func(*Netceptor, string, ...interface{}), level string) {
+	var capturedOutput strings.Builder
+	nodeID := "testnode"
+	n := New(context.Background(), nodeID)
+	n.Logger.SetOutput(&capturedOutput)
+
+	logFunc(n, "unit test %s", "message")
+
+	logged := capturedOutput.String()
+	expectedPrefix := fmt.Sprintf("unit test message {node_id=%s}", nodeID)
```
matoval: `{node_id=%s}` should be `{node_id:%s}`

lranjbar:
We have some unit tests failing perhaps this needs a rebase?

pkg/netceptor/netceptor.go
```diff
@@ -451,6 +452,56 @@ func (s *Netceptor) GetLogger() *logger.ReceptorLogger {
 	return s.Logger
 }
 
+// Error logs message with nodeID prepended.
+func (s *Netceptor) Error(format string, v ...interface{}) {
```
AaronH88: No that you have moved to a Suffix model, are these helper functions needed?

pkg/netceptor/netceptor.go
```diff
@@ -355,6 +355,7 @@ func NewWithConsts(ctx context.Context, nodeID string,
 		MinVersion: tls.VersionTLS12,
 	}
 	s.AddNameHash(nodeID)
+	s.Logger.SetSuffix(map[string]interface{}{"node_id": nodeID})
```
AaronH88: For testing, it might be easier to call the `GetLogger()` function and then SetSuffix

pkg/netceptor/netceptor.go
```diff
@@ -1951,42 +2014,42 @@ func (s *Netceptor) runProtocol(ctx context.Context, sess BackendSession, bi *Ba
 						}
 						remoteEstablished = true
 						if ok && remoteCost != connectionCost {
-							s.removeConnection(remoteNodeID)
-
-							return s.sendAndLogConnectionRejection(remoteNodeID, ci, "we disagree about the connection cost")
```
AaronH88: Removing this return and moving the `sendAndLogConnectionRejection()` has changed the logic of this flow, any reason for this? 😄

pkg/netceptor/netceptor.go
arrestle: If I take them out they won't match workunitbase.go, which I felt was a standard. That said, they aren't needed with this implementation and I'm open to removing them.

pkg/netceptor/netceptor.go
arrestle: Looks like it was removed by mistake.

pkg/netceptor/netceptor.go
arrestle: reverted netceptor.go back to original

pkg/netceptor/netceptor.go
arrestle: Reverted.

pkg/logger/logger.go
```diff
@@ -218,10 +244,25 @@ func (rl *ReceptorLogger) Log(level int, format string, v ...interface{}) {
...
 		rl.Logger.Printf(format, v...)
 	}
 }
 
+func (rl *ReceptorLogger) appendSuffix(format string, v []interface{}) (string, []interface{}) {
+	if rl.Suffix != nil {
+		rl.m.Lock()
+		defer rl.m.Unlock()
+		suffixJSON, err := json.Marshal(rl.Suffix)
+		if err == nil {
```
davemulford: Should the json.Marshal error be logged if `err != nil`?

pkg/logger/logger_test.go
```diff
@@ -120,3 +120,58 @@ func TestDebugPayload(t *testing.T) {
 		})
 	}
 }
+
+func TestGetLogLevelByNameWithSuffix(t *testing.T) {
+	logFilePath := "/tmp/test-suffix"
```
davemulford: Use `os.MkTempDir` in preparation for having a Windows testing environment?

davemulford (commented):
Added two minor things that I've asked about inline. Overall looks good!

pkg/logger/logger_test.go
arrestle: logFilePath := "/tmp/test-suffix"
	os.MkdirTemp("/tmp", "test-suffix")
	defer os.RemoveAll(logFilePath)

dleehr: When you're using something like `os.MkdirTemp()` to create a temporary directory (which is cross-platform by design) you shouldn't include any platform-specifics (like `"/tmp"`) in that call. Hard-coding `"/tmp"` (even on Linux) is not usually a good idea, and some operating systems don't have `"/tmp"` at all.

So that first parameter should be an empty string:

```
dir, err := os.MkdirTemp("", "test-suffix")
	if err != nil {
		log.Fatal(err)
	}
	defer os.RemoveAll(dir) // clean up
```
See https://pkg.go.dev/os#MkdirTemp, https://gobyexample.com/temporary-files-and-directories

dleehr: Actually, now that I look at more of this code - it looks like you're doing all of this to create a temporary file. So you could possibly just use [os.CreateTemp()](https://pkg.go.dev/os#CreateTemp) instead of the other stuff.

But I'll look at this a little more now, since I think I have some other suggestions.

pkg/logger/logger_test.go
```diff
@@ -120,3 +121,63 @@ func TestDebugPayload(t *testing.T) {
...
+	}
+	receptorLogger := logger.NewReceptorLoggerWithSuffix("", suffix)
+	logFile, err := os.OpenFile(logFilePath, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0o600)
+	if err != nil {
+		t.Error("error creating test-output file")
+	}
+
+	expectedLog := "test-suffix {node_id:controller,node_id_to:hop}"
+	receptorLogger.SetOutput(logFile)
+	receptorLogger.Error("test-suffix")
```
dleehr: What do you think about logging a different error message in the test, like `test-error-message`? I think it's confusing to use `test-suffix` because you're testing that something else appears after it, yet it's called `suffix`.

To be clear, I understand what's happening - you're sending a log message, `test-suffix` to the logger and verifying the configured suffix `{node_id:controller,node_id_to:hop}` is logged after it, and the code is working as designed. And I assume you picked `test-suffix` because you're testing the suffix functionality 😄.

But when I read this code, I see something called `test-suffix` at the beginning of the message and it _looks_ like a bug. I think it would be clearer to read this code if we used a different string for the logging.

pkg/logger/logger_test.go
```diff
@@ -120,3 +121,63 @@ func TestDebugPayload(t *testing.T) {
 		})
 	}
 }
+
+func TestGetLogLevelByNameWithSuffix(t *testing.T) {
+	tmpdir, err := os.MkdirTemp(os.TempDir(), "receptor-test-suffix*")
```
dleehr:
```suggestion
	tmpdir, err := os.MkdirTemp("", "receptor-test-suffix*")
```

When using `os.mkdirTemp` we're telling Go we want a temporary file. We shouldn't need to tell it where. But this may be obsolete (see comment on `io.Writer` and `bytes.Buffer` below so don't rush to implement it.

pkg/logger/logger_test.go
```diff
@@ -120,3 +121,63 @@ func TestDebugPayload(t *testing.T) {
...
+		"node_id_to": "hop",
+	}
+	receptorLogger := logger.NewReceptorLoggerWithSuffix("", suffix)
+	logFile, err := os.OpenFile(logFilePath, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0o600)
+	if err != nil {
+		t.Error("error creating test-output file")
+	}
+
+	expectedLog := "test-suffix {node_id:controller,node_id_to:hop}"
+	receptorLogger.SetOutput(logFile)
```
dleehr: I'm commenting on a single line of interest but this comment relates to this entire test function.

There's quite a bit of code in this Test around creating a temporary directory, creating a temporary file, opening it, configuring the logger to write to it, re-reading the file and searching for the entries, and cleaning up all that stuff.

And I remembered a very similar pattern from https://quii.gitbook.io/learn-go-with-tests/go-fundamentals/dependency-injection where they used a `bytes.Buffer` instead.

Specifically, what we need to pass to `log.SetOutput()` (https://pkg.go.dev/log#SetOutput) doesn't need to be a real file, just something that satisfies `io.Writer`. 

So, instead of creating the temp file/dir and `os.ReadFile` and `os.Truncate` can we do something like this:

```go
func TestGetLogLevelByNameWithSuffix(t *testing.T) {
	logBuffer := &bytes.Buffer{}
	logger.SetGlobalLogLevel(4)

	suffix := map[string]interface{}{
		"node_id":    "controller",
		"node_id_to": "hop",
	}
	receptorLogger := logger.NewReceptorLoggerWithSuffix("", suffix)
	expectedLog := "test-suffix {node_id:controller,node_id_to:hop}"
	receptorLogger.SetOutput(logBuffer)

	receptorLogger.Error("test-suffix")
	testOutput := logBuffer.String()

	if !strings.Contains(testOutput, expectedLog) {
		t.Errorf("failed to log correctly, expected: %v got %v", expectedLog, testOutput)
	}
}
```

So no intermediate file to manage. Just provide a `bytes.Buffer` as the output "file". The logger doesn't care if it's a file, just that it has a `write` method. After we log, check if the result of`.String()` includes the log data you expected.

Edited to provide a working function. It only tests the first case but I think you get the idea.

pkg/logger/logger.go
```diff
@@ -88,6 +89,7 @@ type ReceptorLogger struct {
 	log.Logger
 	Prefix string
 	m      sync.Mutex
+	Suffix map[string]interface{}
```
dleehr: What's the purpose of using `interface{}` as the type here? It looks like everything we're sending is a `string` and it needs to get logged, so why not `map[string]string`?

pkg/logger/logger.go
```diff
@@ -218,10 +244,27 @@ func (rl *ReceptorLogger) Log(level int, format string, v ...interface{}) {
...
+	if rl.Suffix != nil {
+		rl.m.Lock()
+		defer rl.m.Unlock()
+		suffixJSON, err := json.Marshal(rl.Suffix)
+		if err == nil {
+			format = strings.ReplaceAll(format, "\n", "") + " %s"
+			v = append(v, strings.ReplaceAll(string(suffixJSON), "\"", ""))
+		} else {
+			format += " error marshalling suffix " + err.Error()
+		}
```
dleehr: This logic seems fragile - we want the `Suffix` in a one-line JSON-like string (but not JSON because we don't want the `"`) we can use for logging, so we ask Go to produce JSON, then do a bunch of string replacements on it for literals like `\n` (fine, but https://pkg.go.dev/encoding/json#Compact may be a better option) and `\"` (why?). Is JSON marshalling actually helping here?

Then, if the marshalling fails, we return an error message as the suffix. Do we have a test for that?

I think this gets back to the earlier comment about Suffix being a map of `string` to `interface{}`. If it's just a map of string to string, we can write a very simple function here to format that suffix into what we want.

dleehr (commented):
Looks like this will be really helpful for us to review logs, so I'm excited to see it completed.

I did have a fair amount of feedback, let me know what you think!

pkg/logger/logger.go
arrestle: Currently it's only holding node_id and remote_id, but there is no reason why it couldn't be an int, float, or a struct with json mappings. just keeping things flexible so we don't need to change it later.

pkg/logger/logger.go
arrestle: good points [from dan](https://github.com/ansible/receptor/pull/1300/files#r2071667776) and here. While I'm all over a better json marshaller, I still hesitate to restrict a suffix to a map of strings. cost, for example, is an integer, and structs generally marshal well (as long as they don't have callback funcs or bad json mappings) so I dislike the idea of not supporting them. That said if it's the group's consensus that a suffix should only be a map[string][string] that's what we'll do.  Let's decide before I start writing tests for failure to marshal, though ;)

pkg/logger/logger.go
dleehr: Understood. I think there's a balance there for sure. Personally I vote [YANGI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) here. The requirement is to include `nodeID`, which is a string type.

dleehr: > just keeping things flexible so we don't need to change it later.

Are we sure we don't need to change it later? If yes I'd expect to see tests supporting it. I do see we have one that supports an integer, but nothing for the struct with JSON mappings.

pkg/logger/logger.go
dleehr: Certainly worth a broader discussion but if we don't get to that today I want to underscore two points:

1. I believe that logging code should be simple and predictable (or bulletproof if it can't be simple). 
2. The logic here of doing string replacement to remove unwanted characters is a red flag / [code smell](https://en.wikipedia.org/wiki/Code_smell) to me. So I raise the questions: "is `json.Marshal` the right tool for this job?" and "Are there any other characters we should be removing/How else could this fail?"

So, back to the "it should be simple" principle - suppose we did switch to `map[string]string`. That simplifies the formatting logic greatly (and I'd suggest extracting the logic to its own testable function regardless). Then if someone wants to provide a different type as the suffix value (like an `int`), the code doesn't compile. So either they satisfy the compiler or refactor at that point. But until we need that, we'd have simplicity and predictability here.

dleehr: See also https://quii.gitbook.io/learn-go-with-tests/go-fundamentals/reflection. I think if we're going to stick with `map[string]interface{}` we should not use `json.Marshal`

pkg/logger/logger_test.go
```diff
@@ -120,3 +121,63 @@ func TestDebugPayload(t *testing.T) {
 		})
 	}
 }
+
+func TestGetLogLevelByNameWithSuffix(t *testing.T) {
```
dleehr: Didn't catch this before, but can we come up with a better name? We're not testing the ability to get the log level by name. We're testing that a configured suffix appears in the log message.

pkg/logger/logger_test.go
arrestle: Agreed. Changed it to `var logBuffer bytes.Buffer`

pkg/logger/logger_test.go
arrestle: yes agreed.

pkg/logger/logger.go
arrestle: I'm seeing the poll is neck and neck on slack [poll](https://redhat-internal.slack.com/archives/C08CJ87JRG9/p1746201641873809) but adding TestLoggerWithConnStructSuffix to attempt to make my case.  

Cleaned up json marshalling - just leaving in the quotes it'll make it easier to run jq on it later if we need to.

pkg/logger/logger.go
arrestle: Added TestLoggerWithConnStructSuffix to illustrate, but truly a judgement call.

pkg/logger/logger.go
```diff
@@ -98,6 +101,30 @@ func NewReceptorLogger(prefix string) *ReceptorLogger {
...
 }
 
+func NewReceptorLoggerWithSuffix(prefix string, suffix map[string]interface{}) *ReceptorLogger {
+	logger := NewReceptorLogger(prefix)
+	logger.SetSuffix(suffix)
+
+	return logger
+}
+
+func (rl *ReceptorLogger) SetSuffix(suffix map[string]interface{}) {
```
dleehr: Can we add a comment to this function so that developers know when to use `SetSuffix` and `UpdateSuffix`?

pkg/logger/logger.go
```diff
@@ -98,6 +101,30 @@ func NewReceptorLogger(prefix string) *ReceptorLogger {
...
+	rl.m.Lock()
+	defer rl.m.Unlock()
+	rl.Suffix = suffix
+}
+
+func (rl *ReceptorLogger) UpdateSuffix(suffix map[string]interface{}) {
```
dleehr: Can we add a comment to this function so that developers know when to use `SetSuffix` and `UpdateSuffix`?

pkg/logger/logger_test.go
```diff
@@ -120,3 +121,74 @@ func TestDebugPayload(t *testing.T) {
...
+	}
+	receptorLogger.UpdateSuffix(suffix)
+	expectedLog = `example message 2 {"cost":12,"node_id":"controller","remote_id":"hop"}` // note fields comes out alphabetically
+	receptorLogger.SanitizedError("example message 2")
+	testOutput = logBuffer.Bytes()
+
+	if !strings.Contains(logBuffer.String(), expectedLog) {
+		t.Errorf("failed to log correctly, expected: %v got %v", expectedLog, string(testOutput))
+	}
+	logBuffer.Reset()
```
dleehr: Just a style comment, but since we're testing two very similar things, can we use a table like we do in [TestDebugPayload](https://github.com/ansible/receptor/blob/devel/pkg/logger/logger_test.go#L74) (and other general testing practices)?

Related, I see that `TestDebugPayload` is using the hard-coded `logFilePath := "/tmp/test-output"` (I assume that's where the pattern came from). So If you felt like updating that to use a `bytesBuffer` as well, that would be a welcome change! It's certainly not within the scope, but if we cared to change it in the new code, we should probably change it in the existing code - at some point.

dleehr (commented):
Thanks for addressing the feedback, nice improvements! I had a couple other thoughts, since I felt I could understand the changes better now 🥇. But nothing showstopping.

We'll see how the poll goes for how you decide to address the suffix type.

pkg/logger/logger.go
arrestle: we're doing map[stringstring

pkg/logger/logger_test.go
arrestle: Great feedback, but, I tried using a table like in TestDebugPayload, but that required special case logic inside the test itself to handle validating logs both initially and after updating the suffix. The branching made the structure difficult to write and felt [wrong](https://www.reddit.com/r/programming/comments/2c9gum/google_testing_blog_dont_put_logic_in_tests/). As a compromise I created two sub-tests. 

I did go ahead and switch TestDebugPayload over to use logBuffer instead of writing to a file. Not really in scope, but figured if we're doing that in new tests, might as well keep things consistent.

pkg/logger/logger.go
arrestle: done

pkg/logger/logger.go
```diff
@@ -98,6 +99,36 @@ func NewReceptorLogger(prefix string) *ReceptorLogger {
 	}
 }
 
+// NewReceptorLoggerWithSuffix to instantiate a new logger object with a new Suffix
```
dleehr:
```suggestion
// NewReceptorLoggerWithSuffix to instantiate a new logger object with a new Suffix.
```

pkg/logger/logger.go
```diff
@@ -218,10 +249,39 @@ func (rl *ReceptorLogger) Log(level int, format string, v ...interface{}) {
 
 	if logLevel >= level {
 		rl.Logger.SetPrefix(prefix)
+		format, v = rl.appendSuffix(format, v)
 		rl.Logger.Printf(format, v...)
 	}
 }
 
+func (rl *ReceptorLogger) appendSuffix(format string, v []interface{}) (string, []interface{}) {
```
dleehr, line 257: Should we change to `string` from `interface{}` here too?

```suggestion
func (rl *ReceptorLogger) appendSuffix(format string, v []string) (string, []string) {
```

pkg/logger/logger_test.go
```diff
@@ -72,14 +72,10 @@ func TestLogLevelToNameWithError(t *testing.T) {
 }
 
 func TestDebugPayload(t *testing.T) {
-	logFilePath := "/tmp/test-output"
+	var logBuffer bytes.Buffer
```
dleehr, line 75: 👍

pkg/netceptor/netceptor.go
```diff
@@ -2010,11 +2014,21 @@ func (s *Netceptor) runProtocol(ctx context.Context, sess BackendSession, bi *Ba
...
+					var connError error
+					connError = nil
+
+					// Verify that the existing connection is valid
+					if ok && existingConn != nil {
+						connError = existingConn.Context.Err()
+					}
+					if ok && connError != nil {
+						s.Logger.Error("Initial handshake failed: %s", connError)
+					}
```
dleehr: Is this related to logging? I think this is an unrelated change you were exploring when the team was investigating a connectivity issue from a customer.

pkg/logger/logger.go
dleehr: I see this was marked resolved but the question wasn't addressed. Can you please help me understand?

arrestle: github auto-resolved when I committed the suggestion. I reverted the commit and manually unresolved just now. Explanation below:

Existing loggers use something like this: `func (rl *ReceptorLogger) Log(level int, format string, v ...interface{}) ` to handle formatting like "%s" "%i" etc... so private 

`func (rl *ReceptorLogger) appendSuffix(format string, v []interface{}) (string, []interface{}) {` 

needs to as well.

pkg/netceptor/netceptor.go
arrestle: That's correct. The original ticket AAP-43201 `Improve logging in netceptor to have full traceability of network` contained requirements to **_add logs to the netceptor_** to identify edge conditions. This was identified as an edge condition during a recent escalation.

pkg/logger/logger.go
dleehr: Thanks for the explanation!

pkg/netceptor/netceptor.go
dleehr: OK. I understand the relationship, but I'm pretty sure this logic was not present in the PR last week when I first reviewed it (please correct me if I'm wrong, but the force-pushes are making it difficult to follow the history to say for sure 😞).

Regardless, this change introduces new behavior. I don't see any tests for that or evidence to support that it fixes a problem. Therefore, I propose holding off on these changes around checking the existing connection - removing these from the PR. 

This PR has been in review for quite some time and we should prioritize wrapping it up. We've had a lot of the team engaging, providing feedback. It's been a valuable process for me and I think it's been a fantastic process for all of us. We've been doing a great job heading towards convergence - refactoring tests, simplifying some things, etc. So my ultimate feedback is that we should be continuing on that.

arrestle: There  actually was code last week that attempted to fix this handshake, due to some cleanup that went awry (ugh...) that history is now lost. I did remove code that changed functionality in favor of *just* reporting the error. This ticket veered into the _*just add node-id and remote-id to each log message*_ but that is not the overarching goal of the ticket, which is `Improve logging in netceptor to have full traceability of network` . Note I added a description to this PR previously that explains why this was added. Please review that and let me know if you agree.

dleehr: > And yes, as valuable as this process is, it's blocking me from getting to hashicorp tickets.

I agree with that and I understand the frustration. However, my feedback is that the path to wrapping this up is directly addressing the feedback raised by multiple reviewers instead of adding more functionality that's not tested.

dleehr: Caught up on a call with @arrestle and reviewed the changes here. I can see that the logging code here should not have any side effects that need to be tested, so I'm good with the changes, but I would like @AaronH88 to weigh in on this, since he may have more context about the requirement in the referenced ticket.

Outside of the PR, my request is to be more mindful of local git issues earlier so that changes don't get lost (or reappear) with force-pushes. @arrestle can you keep that in mind for the next PR?

dleehr (approved):

pkg/backends/tcp.go
```diff
@@ -62,6 +62,10 @@ func (b *TCPDialer) Start(ctx context.Context, wg *sync.WaitGroup) (chan netcept
 				return nil, err
 			}
 
+			if b.logger != nil {
```
AaronH88, line 65: Nice work here, can you open a new PR for UDP and Websocket connections for something similar please (OUT OF SCOPE FOR THIS PR)

pkg/netceptor/netceptor.go
```diff
@@ -2010,11 +2014,21 @@ func (s *Netceptor) runProtocol(ctx context.Context, sess BackendSession, bi *Ba
...
 					}
+					var connError error
+					connError = nil
+
+					// Verify that the existing connection is valid
+					if ok && existingConn != nil {
+						connError = existingConn.Context.Err()
+					}
+					if ok && connError != nil {
+						s.Logger.Error("Initial handshake failed: %s", connError)
```
AaronH88: So the connError is not a new handshake coming in right? 

This connError, if it happens, is when an existing connection's context is closed. Do I have that correct? 

I think I do, in which case this error is confusing. 
Also, if the context is canceled, and we are just printing an error and not doing anything about it, that seems kinda strange. 
If the context is canceled I would expect the `existingConn` to be removed from the list of `s.connections` . 
And if I have gotten all the logic correct so far, then I believe this section of logging is now a log and a fix, so we should probably have a unit test for it, and BOOM, more scope creep 😂 

So to not tie up this PR anymore, can I please suggest removing this area of logging for now, but adding it to a different PR with appropriate reactions and unit tests? 

Let me know if I got any logic wrong or you would prefer to go in another direction 😄

arrestle: This was explained in the description: _Note that additional diagnostics were added to identify existing connections with initial handshake errors, but does not remove them as that is a bug that is out of scope. See https://github.com/ansible/receptor/pull/1303 for details._ 

Note this change *will* make debugging customer logs with handshake errors easier to diagnose, and the point of the ticket contains "full traceability of network". I feel I'm under-reporting network errors as it is.

arrestle: @AaronH88 requested a change over slack:
`ERROR: DATE Context for existing connection error: context canceled {node_id:x, remote_id:y}` 
change applied.

pkg/backends/tcp.go
arrestle: Done. Referenced in a comment on the original Jira for this PR.

AaronH88 (approved):
lgtm

PR #1302, opened 2025-04-15
Update Dockerfile to include ansible-runner

kdelee:
where this is used in controller/AAP/awx we need ansible-runner for the health check. this brings more in line w/ downstream so we can test downstream AAP w/ upstream receptor image when y'all have new fixes you'd like to try out


rooftopcellist (approved):

matoval (approved):
lgtm!

PR #1304, opened 2025-04-17
update check for connection info to have less time complexity in netceptor.go/runProtocol(...)

lranjbar:
tldr: Changes from using a for loop iterating over all the keys in a map to check for key existence in a map. This change does not make changes to the business logic in this small code block.

longer version:
The code `for conn := range s.connections` will iterate over all the keys in the map `s.connections`. This is a map where the nodeIDs are the keys of the map and the data is a connInfo object. We can simply check to see if the key exists in the map instead of looping over the keys to see if we have a connection to this node. 

even more info:
[In Golang you can use a two-value assignment to test the existence of a key in a map.](https://go.dev/blog/maps) The most common variable to use for existence is "ok".. which I generally refer to as the "if ok" syntax of Golang.

### DIFF
--- a/pkg/netceptor/netceptor.go
+++ b/pkg/netceptor/netceptor.go
@@ -2003,19 +2003,20 @@ func (s *Netceptor) runProtocol(ctx context.Context, sess BackendSession, bi *Ba
 						return s.sendAndLogConnectionRejection(remoteNodeID, ci, "it is not in the allowed peers list")
 					}
 
+					// Check if there is connection cost for this remoteNodeID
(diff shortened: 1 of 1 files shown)


dleehr (approved):
Looks like a good change, and thanks for sharing the Go pattern here. Seems like a win-win. 

As you say, it reduces the time complexity, and for the reader it's less code that's more obvious. We just want to know if our item is in the map, so let's use the idiom!

arrestle (approved):
Code practices!

PR #1308, opened 2025-04-21
Refactor unixsock and add Windows test

davemulford:
Refactors unixsock.go so the build tags can be removed. This let me move the Windows-specific code into the same file as everything else, which made the Windows code testable on non-Windows machines.


matoval (approved):
lgtm

AaronH88:
While I do like the discovery of the OS at run time, I also like the discovery at compile time, it reduces the size of the binary and ensures that the unix code dosent even exist on windows binaries. 

To that end, and as the compile time flag can only be applied to entire files, I propose keeping the existing compile time flags, and instead creating a new unix file that is common to all os versions.

This will ensure that on windows machines, all the unix code dosent exist,
And on unix machines, no windows code exists,
Only a common file to say the error messages.

What do you think?

davemulford:
Thank you and makes sense to me. I'll work on making that happen.

AaronH88:
This is looking perfect, great job!

AaronH88 (approved):
lgtm

PR #1312, opened 2025-04-24
cmd: Add unit tests and fix WebSocket path default

TheRealHaoLiu:
This commit adds unit tests for the cmd package, improving test coverage from 0% to 50.5%. The tests cover configuration parsing, default settings, and command execution flow.

Key changes:
- Add tests for configuration functions and utility methods
- Fix bug in SetWSListenerDefaults where it incorrectly set BindAddr instead of Path for empty paths
- Add test infrastructure including mock implementations
- Skip complex tests that require additional setup for future implementation

The WebSocket path bug fix ensures that when a WebSocket listener is configured without specifying a path, it correctly defaults to '/' instead of incorrectly setting the BindAddr field.

### DIFF
--- a/cmd/config_test.go
+++ b/cmd/config_test.go
@@ -1,0 +1,285 @@
+package cmd
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/ansible/receptor/pkg/backends"
+	"github.com/ansible/receptor/pkg/controlsvc"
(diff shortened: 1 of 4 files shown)


AaronH88 (approved):
lgtm 1 of 2

matoval (approved):
lgtm

PR #1314, opened 2025-04-28
Fix flock_test from interrupting utils package tests

matoval:
Utils package unit tests are stopping after running flock_test because the test was unlocking FD 1 which is stdout.

This change fixes the test by creating a temp file and unlock it instead of FD 1.


davemulford (approved):
LGTM

AaronH88 (approved):
lgtm

PR #1317, opened 2025-05-02
Additional unit tests in netceptor/conn.go

dleehr:


pkg/netceptor/conn_simple_test.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// Skip the tracer test since it's difficult to test without mocking
+// the quic.ConnectionID type
+
+// TestNetceptorListen tests basic functionality of the Listen method
+func TestNetceptorListen(t *testing.T) {
+	// Skip this test in CI environments or when network operations are not possible
+	if os.Getenv("CI") != "" {
+		t.Skip("Skipping test in CI environment")
```
dleehr, line 69: Noting the `t.Skip()` here

pkg/netceptor/conn_simple_test.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	return string(result)
+}
+
+// TestNetceptorListenAndAdvertise tests basic functionality of the ListenAndAdvertise method
+func TestNetceptorListenAndAdvertise(t *testing.T) {
+	// Skip this test in CI environments or when network operations are not possible
+	if os.Getenv("CI") != "" {
+		t.Skip("Skipping test in CI environment")
```
dleehr, line 126: Noting the `t.Skip()` here

pkg/netceptor/conn_simple_test.go
```diff
@@ -0,0 +1,185 @@
...
+	// Verify the result
+	assert.NotNil(t, config)
+	assert.True(t, config.InsecureSkipVerify)
+	assert.NotNil(t, config.VerifyPeerCertificate)
+	assert.Equal(t, []string{"netceptor"}, config.NextProtos)
+	assert.Equal(t, host, config.ServerName)
+}
+
+// Skip the tracer test since it's difficult to test without mocking
+// the quic.ConnectionID type
```
PabloHiro, line 63: nit: I would not put the comment here. It would force us to maintain the comment itself if we ever refactor, and nothing checks the comments validity. I think something like below is more maintainable

```
func TestTracer(t *testing.T) {
    t.Skip("Skip the tracer test since it's difficult to test without mocking the quic.ConnectionID type")
}
```

pkg/netceptor/conn.go
```diff
@@ -564,7 +578,8 @@ func generateServerTLSConfig() *tls.Config {
 	}
 }
 
-func verifyServerCertificate(rawCerts [][]byte, _ [][]*x509.Certificate) error {
+// VerifyServerCertificate verifies that the server certificate has the expected common name.
+func VerifyServerCertificate(rawCerts [][]byte, _ [][]*x509.Certificate) error {
```
PabloHiro, line 582: Not for this PR:

Just so this is in your radar Dan, I think this function overlaps partly with:
https://github.com/ansible/receptor/blob/devel/pkg/netceptor/netceptor.go#L1013

It would be nice to consolidate at some point.

pkg/netceptor/conn.go
```diff
@@ -45,12 +45,12 @@ type AcceptResult struct {
 
 // Listener implements the net.Listener interface via the Receptor network.
 type Listener struct {
-	s          *Netceptor
-	pc         PacketConner
-	ql         *quic.Listener
+	S          *Netceptor
```
AaronH88, line 48: I think I would prefer creating some `Getters()` and `Setters()` rather than exposing the private data fields, 

I can see you have also created a new constructor, so its just continuing the same idea

pkg/netceptor/conn.go
```diff
@@ -313,7 +315,18 @@ func (li *Listener) Close() error {
 
 // Addr returns the local address of this listener.
 func (li *Listener) Addr() net.Addr {
-	return li.pc.LocalAddr()
+	return li.PC.LocalAddr()
+}
+
+// NewTestListener creates a Listener for testing purposes.
+func NewTestListener(s *Netceptor, pc PacketConner, addr net.Addr, acceptChan chan *AcceptResult, doneChan chan struct{}, doneOnce *sync.Once) *Listener {
```
AaronH88, line 322: Just for readability of the file, having the constructor close to the struct definition would be nice 😄

pkg/netceptor/conn_simple_test.go
```diff
@@ -0,0 +1,185 @@
+package netceptor
```
AaronH88, line 1: Im not quite sure why there is a need for both a conn_simple_test as well as a con_tls_test, I would expect these to be combined into one file 😄

pkg/netceptor/conn_tls_test.go
```diff
@@ -0,0 +1,125 @@
+package netceptor_test
```
AaronH88, line 1: these tests can be moved into the netceptor package, perhaps then you would need to expose so many functions

pkg/netceptor/conn_simple_test.go
```diff
@@ -0,0 +1,185 @@
...
+	"crypto/x509"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestServerTLSConfig tests the GenerateServerTLSConfig function.
+func TestServerTLSConfig(t *testing.T) {
```
AaronH88, line 14: opportunity for table driven testing here?

dleehr:
Thanks for the feedback. I'm replacing this with #1320 which adds tests in the `netceptor` package so we don't need to modify the source files. I'm still checking a few things, so that's in draft for now.

PR #1320, opened 2025-05-09
Additional tests for conn.go

dleehr:
Kept in same package netceptor

Replaces #1317


pkg/netceptor/conn_internal_test.go
```diff
@@ -0,0 +1,870 @@
...
+	return 0
+}
+
+func (m *mockNetceptorForListen) AddLocalServiceAdvertisement(service string, connType byte, tags map[string]string) {
+	// Mock implementation - do nothing
+}
+
+func (m *mockNetceptorForListen) GetLogger() *logger.ReceptorLogger {
+	return m.logger
+}
```
dleehr, line 221: The pattern for these mocks is to generate them into  `mock_netceptor` rather than writing them in-line in the test file.

pkg/netceptor/conn_internal_test.go
```diff
@@ -0,0 +1,870 @@
...
+	maxForwardingHops byte
+	logger            *logger.ReceptorLogger
+	context           context.Context
+	cancelFunc        context.CancelFunc
+	listenerLock      sync.RWMutex
+	getEphemeralFunc  func() string
+}
+
+// listen is a simplified implementation of Netceptor.listen for testing.
+func (m *mockNetceptorForListen) listen(ctx context.Context, service string, tlscfg *tls.Config, advertise bool, adTags map[string]string) (*Listener, error) {
```
dleehr, line 156: But if we mock the `listen` method, we won't be running the real one from `TestListen`, right?

PR #1321, opened 2025-05-12
Sanity check - do not review or merge

dleehr:
Seeing test failures over in #1320 but haven't made code changes around the things that are failing. Just checking if a new branch off `devel` is passing or not.

### DIFF
--- a/receptorctl/setup.cfg
+++ b/receptorctl/setup.cfg
@@ -13,7 +13,7 @@ version = file: .VERSION
 packages = find:
 install_requires =
     python-dateutil
-    click
+    click < 8.2.0
     pyyaml
 
 [options.entry_points]


dleehr:
Seems to be related to https://github.com/pallets/click/issues/2908. A change in `click` in the 8.2.0 release is causing this ImportError.

Since we're not pinning `click` in https://github.com/ansible/receptor/blob/devel/receptorctl/setup.cfg#L16, we're getting 8.2.0. But this release requires Python 3.9 or later, so in the 3.8/3.9 environments we're still getting click 8.1.8 and that continues to pass.

In https://github.com/ansible/receptor/pull/1321/commits/78c44d84ee2cb29fda14523bda263ed8a72f43cb I'm testing if limiting the version prevents the failures.

PR #1325, opened 2025-05-20
feat: Use pyproject.toml instead of setup.cfg + pinned dependencies + ruff as linter.

PabloHiro:
AAP-45792

Changes:

* Use pyproject.toml instead of setup.cfg for receptorctl and receptor-python-worker [PEP518](https://peps.python.org/pep-0518/)
* Set minimum version of all project dependencies, leave loosen test (`optional-dependencies`) dependencies as they will be constrained by project dependencies as per [PEP 633](https://peps.python.org/pep-0633/#optional-dependencies)
* Use ruff as linter. Flake8 does not have native support for pyproject.toml and ruff has modern support.
* Remove usage of the file `.VERSION` and instead use standard tooling. For linting or test report, default to a `fallback_version`. When building, use `SETUPTOOLS_SCM_PRETEND_VERSION` variables. [Documentation](https://setuptools-scm.readthedocs.io/en/stable/config/)
* Linted .py files with ruff.

All nox sessions are kept, so CI does not need to be changed.


docs/source/conf.py
```diff
@@ -28,7 +28,6 @@
 # ones.
 extensions = [
     "sphinx.ext.autosectionlabel",
-    "pbr.sphinxext",
```
PabloHiro, line 31: This needs to be removed because it uses `setup.cfg` as source of truth for the version. Not the case anymore.

receptorctl/pyproject.toml
```diff
@@ -1,6 +1,51 @@
+[project]
+name = "receptorctl"
+authors = [{name = "Red Hat", email = "info@ansible.com"}]
+description = "Receptorctl is a front-end CLI and importable Python library that interacts with Receptor over its control socket interface."
+readme = "README.md"
+dynamic = ["version"]
+dependencies = [
+  "python-dateutil==2.9.0",
+  "click==8.1.8",
+  "PyYAML==6.0.2",
```
simaishi: Pin to exact versions will cause conflicts, especially because the same receptor version is used in multiple AAP releases with different dependent rpm versions. Can you set lower and/or upper version constraint instead?

davemulford (approved):
Adding approval, knowing that this won't be merged until the comment by @simaishi about version pinning is resolved.

receptorctl/requirements/requirements.txt
```diff
@@ -0,0 +1,14 @@
...
+#
+#    pip-compile --allow-unsafe --output-file=requirements/requirements.txt --strip-extras pyproject.toml
+#
+click==8.1.8
+    # via receptorctl (pyproject.toml)
+python-dateutil==2.9.0
+    # via receptorctl (pyproject.toml)
+pyyaml==6.0.2
+    # via receptorctl (pyproject.toml)
+six==1.17.0
```
AaronH88, line 13: is it ok to pin to exact versions here?

PabloHiro: These are auto-generated when the versions in `pyproject.toml` get resolved. Think of it as a lock file, not a source of truth

receptorctl/pyproject.toml
PabloHiro: You are right. Testing this with different versions is a bit challenging though. Will update the PR once I manage to do that.

AaronH88 (approved):
lgtm

simaishi (approved):

PR #1332, opened 2025-06-02
AAP-46484 Capture k8s pod diagnostics

arrestle:
Detect a pod failure caused by OOMKilled or other infrastructure related error.

This one outputs:

```
ERROR: Remote unit failed: Pod diagnostics failed: pod default/kubeit-btlm7 failed due to OOMKilled:  reason OOMKilled

INFO 2025/06/03 17:06:57 Detected EOF for pod default/kubeit-btlm7. {"node_id":"kube-node"}
WARNING 2025/06/03 17:06:59 Pod diagnostics failed: pod default/kubeit-btlm7 failed due to OOMKilled:  reason OOMKilled {"node_id":"kube-node"}
```

Note we have a new [PR 1368](https://github.com/ansible/receptor/pull/1368) for CapturePodStatus


pkg/workceptor/kubernetes.go
```diff
@@ -113,6 +117,14 @@ func (ku KubeAPIWrapper) Get(ctx context.Context, clientset *kubernetes.Clientse
...
 
 func (ku KubeAPIWrapper) Create(ctx context.Context, clientset *kubernetes.Clientset, namespace string, pod *corev1.Pod, opts metav1.CreateOptions) (*corev1.Pod, error) {
+	if pod.Annotations == nil {
+		pod.Annotations = make(map[string]string)
+	}
+
+	// Add default system annotations
+	pod.Annotations["receptor.redhat.com/created-by"] = "receptor"
+	pod.Annotations["receptor.redhat.com/creation-timestamp"] = time.Now().UTC().Format(time.RFC3339)
+	pod.Annotations["receptor.redhat.com/request_id"] = "xxx-xxxx-xxxx-xxxx" // Placeholder, replace with actual request ID if available
```
lranjbar: I like it, but instead of "receptor" I think our official product name is "automation-mesh" but maybe this is fine 🤔 

How would the creation timestamp be different than the kubernetes creation timestamp on the pod? Perhaps we can drop this one

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +236,106 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+	for _, cs := range pod.Status.ContainerStatuses {
+		if cs.State.Terminated == nil {
+			return true, "container has not terminated", nil
+		}
+		if cs.State.Terminated.ExitCode != 0 {
+			return false, fmt.Sprintf("container exited with code %d: %s", cs.State.Terminated.ExitCode, cs.State.Terminated.Reason), nil
+		}
+		if cs.State.Terminated.Reason == "Error" {
+			return false, "container terminated with 'Error'", nil
+		}
```
lranjbar:
```suggestion
		if cs.State.Terminated.ExitCode != 0 {
			return false, fmt.Sprintf("container exited with code %d: %s", cs.State.Terminated.ExitCode, cs.State.Terminated.Reason), nil
		}
```

`ExitCode` and `Reason` are returned at the same time `cs.State.Terminated.Reason == "Error"` is included in the `cs.State.Terminated.ExitCode != 0` statement so lines 281-283 is unreachable code.

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +236,106 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+		return false, "pod is nil", fmt.Errorf("pod is nil")
+	}
+
+	switch pod.Status.Phase {
+	case corev1.PodFailed:
+		for _, cs := range pod.Status.ContainerStatuses {
+			if cs.State.Terminated != nil {
+				switch cs.State.Terminated.Reason {
+				case "OOMKilled", "Evicted":
+					return false, cs.State.Terminated.Reason, nil
```
lranjbar:
```suggestion
		for _, cs := range pod.Status.ContainerStatuses {
			if cs.State.Terminated != nil {
				switch cs.State.Terminated.ExitCode {
				case 0:
				        break
				default:
					return false, cs.State.Terminated.Reason, nil
```

We can simplify on `cs.State.Terminated.ExitCode` because an exit code of 0 is success and everything else is failure.

pkg/workceptor/kubernetes.go
```diff
@@ -902,7 +1014,32 @@ func (kw *KubeUnit) runWorkUsingLogger() {
...
+			reason := fmt.Sprintf("Pod diagnostics failed: %v reason %s", err, reason)
+			kw.GetWorkceptor().nc.GetLogger().Warning(reason)
+			kw.UpdateBasicStatus(WorkStateFailed, reason, stdout.Size())
+			return
+		} else if !ok {
+			kw.GetWorkceptor().nc.GetLogger().Warning("Pod did not succeed: %s", reason)
+			kw.UpdateBasicStatus(WorkStateFailed, reason, stdout.Size())
+			return
+		}
+	}
```
lranjbar: I had originally put my code here too. The issue I had that sometimes we detect EOF and cut the stream before we get here. That's why I moved it to after EOF detection. We will always detect EOF but we don't always get to the stream closing out completely before the pod/container goes belly up.

lranjbar (changes requested):
Looking good! Just some feedback on these :)

pkg/workceptor/kubernetes.go
arrestle: agreed. updated. would love to have that request_id, maybe the work_unit_id too? @AaronH88 please weigh in.

pkg/workceptor/kubernetes.go
arrestle: Ohhhhh.... Gooooood catch!  Created a new `func (kw *KubeUnit) capturePodStatus` so we can call it in both places.

matoval:
When a pod fails because the ephemeral-storage is over the limit I'm getting these error logs:
`ERROR: Remote unit failed: Error` in the terminal output and this in the logs:
````
WARNING 2025/06/09 14:34:02 Pod did not succeed: Error {"node_id":"kube-node"}
WARNING 2025/06/09 14:34:02 Pod did not succeed: Error {"node_id":"kube-node"}
````

I do see OOMKilled when the pod fails because of an OOMKill

File: pkg/workceptor/mock_workceptor/workunitbase.go
Comment (PabloHiro) [inline pkg/workceptor/mock_workceptor/workunitbase.go:1]: Why are we empyting this file?

File: pkg/workceptor/kubernetes.go [hunk 5]
```diff
@@ -902,12 +1014,49 @@ func (kw *KubeUnit) runWorkUsingLogger() {
...
+	ok, reason, err := kw.KubeAPIWrapperInstance.GetPodStatus(pod)
+	if err != nil {
+		reason := fmt.Sprintf("Pod diagnostics failed: %v reason %s", err, reason)
+		kw.GetWorkceptor().nc.GetLogger().Warning(reason)
+		kw.UpdateBasicStatus(WorkStateFailed, reason, stdoutSize)
+		return false, err
+	} else if !ok {
+		kw.GetWorkceptor().nc.GetLogger().Warning("Pod did not succeed: %s", reason)
+		kw.UpdateBasicStatus(WorkStateFailed, reason, stdoutSize)
+		return false, fmt.Errorf("pod did not succeed: %s", reason)
```
PabloHiro: Would it be possible to include the name of the namespace and name of the pod in the log message?

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +235,105 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+				}
+			}
+		}
+		return true, "", nil
+
+	case corev1.PodSucceeded:
+		return true, "", nil
+
+	case corev1.PodRunning, corev1.PodPending:
+		return true, fmt.Sprintf("pod ohase: %s", pod.Status.Phase), fmt.Errorf("pod not in terminal state")
```
PabloHiro:
```suggestion
		return true, fmt.Sprintf("pod phase: %s", pod.Status.Phase), fmt.Errorf("pod not in terminal state")
```

PabloHiro (changes requested):

pkg/workceptor/mock_workceptor/workunitbase.go
arrestle: no idea. I ran `make generate`,  will research.

pkg/workceptor/kubernetes.go
arrestle: Good idea. Adding container name too for app failures.

pkg/workceptor/kubernetes.go
arrestle: fixed.

pkg/workceptor/mock_workceptor/workunitbase.go
arrestle: Fixed but will regroup with Lisa later to find out why make generate cleared this file.

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +234,105 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+	interval := 1 * time.Second
+	latestPod := pod
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return latestPod, fmt.Errorf("timeout: pod did not complete within %s", timeout)
```
PabloHiro:
```suggestion
			return latestPod, fmt.Errorf("timeout: pod %s/%s did not complete within %s", pod.Namespace, pod.Name, timeout)
```

We are going to need this multiple times. Would it make sense to make this a function that accepts pod and a string as arguments and it logs the error with the pod context?

arrestle: Cleaned this up a slightly different way. take a look please!

pkg/workceptor/mock_workceptor/workunitbase.go
arrestle: Fixed by pr.

.gitignore
```diff
@@ -29,4 +29,6 @@ venv
 /vendor
 pkg/services/.lock
 pkg/workceptor/status
-pkg/workceptor/status.lock
\ No newline at end of file
+pkg/workceptor/status.lock
+demo
+**/__debug_*
```
matoval, line 33: Are these two necessary to ignore in git?

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +237,103 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+		return true, "", nil
+
+	case corev1.PodSucceeded:
+		return true, "", nil
+
+	case corev1.PodRunning, corev1.PodPending:
+		return true, fmt.Sprintf("pod phase: %s", pod.Status.Phase), fmt.Errorf("pod not in terminal state")
+
+	default:
+		return false, fmt.Sprintf("unknown phase: %s", pod.Status.Phase), fmt.Errorf("invalid pod phase")
```
PabloHiro: Since we don't want to create new functions for now, can you add the pod name and pod namespaces for both of these lines as well?

pkg/workceptor/kubernetes.go
PabloHiro: Since we are not having a dedicated logging function, can we add the context of the pod name and pod namespace here?

pkg/workceptor/kubernetes.go
```diff
@@ -902,12 +1016,49 @@ func (kw *KubeUnit) runWorkUsingLogger() {
...
+func (kw *KubeUnit) CapturePodStatus(pod *corev1.Pod, stdoutSize int64, timeout time.Duration) (ok bool, err error) {
+
+	pod, err = kw.KubeAPIWrapperInstance.WaitForPodCompleted(pod, kw.clientset, timeout)
+	if err != nil {
+		kw.GetWorkceptor().nc.GetLogger().Debug("Job complete and pod is still running: %v", err)
+	}
+
+	ok, reason, err := kw.KubeAPIWrapperInstance.GetPodStatus(pod)
+	if err != nil {
+		reason := fmt.Sprintf("Pod diagnostics failed: %v reason %s", err, reason)
```
PabloHiro: Pod information here too please :pray:

pkg/workceptor/kubernetes.go
```diff
@@ -902,12 +1016,49 @@ func (kw *KubeUnit) runWorkUsingLogger() {
...
+		kw.GetWorkceptor().nc.GetLogger().Warning(reason)
+		kw.UpdateBasicStatus(WorkStateFailed, reason, stdoutSize)
+		return false, err
+	} else if !ok {
+		kw.GetWorkceptor().nc.GetLogger().Warning("Pod did not succeed: %s", reason)
+		kw.UpdateBasicStatus(WorkStateFailed, reason, stdoutSize)
+		return false, fmt.Errorf("pod did not succeed: %s", reason)
```
PabloHiro: Pod information here too please 🙏

PabloHiro (changes requested):

pkg/workceptor/kubernetes.go
arrestle: Hi Pablo, agreed. The container name is added here, the pod namespace and name is added in GetPodStatus with the podRef := ... command.

pkg/workceptor/kubernetes.go
arrestle: OK. I don't get it, on my branch I have:

```
// GetPodStatus checks if the pod has successfully completed its application logic and infrastructure is healthy.
func (ku KubeAPIWrapper) GetPodStatus(pod *corev1.Pod) (bool, string, error) {

	if pod == nil {
		return false, "pod is nil", fmt.Errorf("pod is nil")
	}

	podRef := fmt.Sprintf("pod %s/%s", pod.Namespace, pod.Name)

	infraOK, reason, err := ku.podInfrastructureSuccess(pod)

	if !infraOK || err != nil {
		return infraOK, fmt.Sprintf("%s infrastructure %s", podRef, reason), err
	}

	appOK, reason, err := ku.podApplicationSuccess(pod)
	if !appOK || err != nil {
		return appOK, fmt.Sprintf("%s application %s", podRef, reason), err
	}

	return true, "", nil
}

```

But it doesn't show up here.

pkg/workceptor/kubernetes.go
arrestle: See above.

pkg/workceptor/kubernetes.go
PabloHiro: Ah sorry, yeah, GetPodStatus should be passing that info in the `err` already. My bad

.gitignore
arrestle: __debug_ was bringing in binaries, so I would say yes. I will remove demo though.

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +237,101 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
 	return inner
 }
 
+// podInfrastructureFailure checks if the pod has failed to start due to infrastructure issues.
+func (ku KubeAPIWrapper) podInfrastructureSuccess(pod *corev1.Pod) (bool, string, error) {
```
davemulford: The func name and comment don't seem to match here. Is there a function named `podInfrastructureFailure`?

Also, the linter isn't happy with the blank line after the function definition.

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +237,101 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+
+	default:
+		return false, fmt.Sprintf("unknown phase: %s", pod.Status.Phase), fmt.Errorf("invalid pod phase")
+	}
+}
+
+// podApplicationSuccess checks if the pod has successfully completed its application logic.
+func (ku KubeAPIWrapper) podApplicationSuccess(pod *corev1.Pod) (bool, string, error) {
+
+	for _, cs := range pod.Status.ContainerStatuses {
```
davemulford: Maybe this loop should check to make sure the status is for a specific container. This [code in CreatePod](https://github.com/arrestle/receptor/blob/capture-k8s-pod-diagnostics/pkg/workceptor/kubernetes.go#L590-L597) (shown below) suggests it's possible to have multiple containers in the Pod.

```Go
		for i := range spec.Containers {
			if spec.Containers[i].Name == "worker" {
				spec.Containers[i].Stdin = true
				spec.Containers[i].StdinOnce = true
				foundWorker = true

				break
			}
		}
```

davemulford (changes requested):

.gitignore
arrestle: Removed demo.

pkg/workceptor/kubernetes.go
arrestle: Discussed on slack https://redhat-internal.slack.com/archives/C08CJ87JRG9/p1749741250597779

pkg/workceptor/kubernetes.go
```diff
@@ -902,12 +1015,49 @@ func (kw *KubeUnit) runWorkUsingLogger() {
...
+
+	pod, err = kw.KubeAPIWrapperInstance.WaitForPodCompleted(pod, kw.clientset, timeout)
+	if err != nil {
+		kw.GetWorkceptor().nc.GetLogger().Debug("Job complete and pod is still running: %v", err)
+	}
+
+	ok, reason, err := kw.KubeAPIWrapperInstance.GetPodStatus(pod)
+	if err != nil {
+		reason := fmt.Sprintf("Pod diagnostics failed: %v reason %s", err, reason)
+		kw.GetWorkceptor().nc.GetLogger().Warning(reason)
```
matoval:
```suggestion
		kw.GetWorkceptor().nc.GetLogger().Warning("%s", reason)
```
This will fix the lint issue by providing a constant format string.

pkg/workceptor/kubernetes.go
```diff
@@ -902,12 +1015,49 @@ func (kw *KubeUnit) runWorkUsingLogger() {
...
+	// Only transition to WorkStateSucceeded if the work unit is still running
+	// and has not already failed due to diagnostic or streaming errors.
+	// This ensures we don't override a previously set WorkStateFailed.
 	if kw.GetContext().Err() != context.Canceled && kw.Status().State == WorkStateRunning {
 		kw.UpdateBasicStatus(WorkStateSucceeded, "Finished", stdout.Size())
 	}
 }
 
+func (kw *KubeUnit) CapturePodStatus(pod *corev1.Pod, stdoutSize int64, timeout time.Duration) (ok bool, err error) {
+
```
matoval: Lint is complaining about this new line

pkg/workceptor/kubernetes.go
```diff
@@ -902,12 +1015,49 @@ func (kw *KubeUnit) runWorkUsingLogger() {
...
+	if err != nil {
+		kw.GetWorkceptor().nc.GetLogger().Debug("Job complete and pod is still running: %v", err)
+	}
+
+	ok, reason, err := kw.KubeAPIWrapperInstance.GetPodStatus(pod)
+	if err != nil {
+		reason := fmt.Sprintf("Pod diagnostics failed: %v reason %s", err, reason)
+		kw.GetWorkceptor().nc.GetLogger().Warning(reason)
+		kw.UpdateBasicStatus(WorkStateFailed, reason, stdoutSize)
+		return false, err
```
matoval: Expected a new line before this return

pkg/workceptor/kubernetes.go
```diff
@@ -902,12 +1015,49 @@ func (kw *KubeUnit) runWorkUsingLogger() {
...
+	} else if !ok {
+		kw.GetWorkceptor().nc.GetLogger().Warning("Pod did not succeed: %s", reason)
+		kw.UpdateBasicStatus(WorkStateFailed, reason, stdoutSize)
+		return false, fmt.Errorf("pod did not succeed: %s", reason)
```
matoval: Expected a new line before this return

pkg/workceptor/kubernetes_test.go
```diff
@@ -562,3 +563,241 @@ func TestKubeLoggingWithReconnect(t *testing.T) {
...
+			ok, reason, err := realImplementation.GetPodStatus(tt.pod)
+			if ok != tt.wantOk || reason != tt.wantReason || (err != nil) != tt.wantErr {
+				t.Errorf("Failed %s case: ok=%v reason=%q err=%v", tt.name, ok, reason, err)
+			}
+			if err != nil && !strings.Contains(err.Error(), tt.wantReason) {
+				t.Errorf("Expected error message '%s', got '%s'", tt.wantReason, err.Error())
+			}
+		})
+	}
+
```
matoval: Lint is complaining about this new line

pkg/workceptor/kubernetes_test.go
```diff
@@ -562,3 +563,241 @@ func TestKubeLoggingWithReconnect(t *testing.T) {
...
+}
+
+func TestWaitForPodCompleted(t *testing.T) {
+
```
matoval: Lint is complaining about this new line

matoval (changes requested):
Lint is complaining about these issues:

pkg/workceptor/kubernetes_test.go
```diff
@@ -1252,3 +1253,240 @@ func TestKubeUnitRestart(t *testing.T) {
...
+				}
+			}
+			if tt.wantErr && err != nil {
+				if !strings.Contains(err.Error(), tt.wantError) {
+					t.Errorf("Expected error message '%s', got '%s'", tt.wantError, err.Error())
+				}
+			}
+		})
+	}
+}
```
matoval: Lint wants a new line at the end of the file

matoval (changes requested):
Lint fix

pkg/workceptor/kubernetes.go
```diff
@@ -25,7 +25,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
+	schema "k8s.io/apimachinery/pkg/runtime/schema"
```
matoval: Explicitly setting schema isn't necessary here.

matoval (changes requested):
Looks good, I just had one comment.

matoval (approved):
lgtm

pkg/workceptor/kubernetes_test.go
```diff
@@ -1252,3 +1253,240 @@ func TestKubeUnitRestart(t *testing.T) {
 		assert.Contains(t, err.Error(), "restart not implemented for streammethod tcp")
 	})
 }
+
+func createTestPods() (*corev1.Pod, *corev1.Pod, *corev1.Pod, *corev1.Pod) {
```
lranjbar: nit: Test pods could be just declared as variables in the `_test.go` file to reduce future memory usage. 

(If I were to call `createTestPods()` in a second function it would allocate another set of memory for the same objects.)

lranjbar (dismissed):

PabloHiro (approved):

pkg/workceptor/kubernetes.go
```diff
@@ -88,9 +88,15 @@ type KubeAPIer interface {
 	UntilWithSync(context.Context, cache.ListerWatcher, runtime.Object, watch2.PreconditionFunc, ...watch2.ConditionFunc) (*watch.Event, error)
 	NewFakeNeverRateLimiter() flowcontrol.RateLimiter
 	NewFakeAlwaysRateLimiter() flowcontrol.RateLimiter
+
+	// Pod status helpers
+	GetPodStatus(pod *corev1.Pod) (bool, string, error)
```
AaronH88: These two helper functions should not be added to `KubeAPIer`

`KubeAPIer` is an interface to wrap the kube api, and make mocking possible. 

Adding more functions that are not part of Kube Api will break this intention, and I am surprised the calls to the real kube are working. 

Instead let us separate our concerns. Create a new interface, something like `KubeHelperFunctions interface` 
Then have KubeAPIWrapper implement these functions like you have below. 

This will work because a struct can satisfy multiple interfaces in go 😄

pkg/workceptor/kubernetes.go
```diff
@@ -88,9 +88,15 @@ type KubeAPIer interface {
...
+	WaitForPodCompleted(pod *corev1.Pod, clientset kubernetes.Interface, timeout time.Duration) (*corev1.Pod, error)
 }
 
-type KubeAPIWrapper struct{}
+type KubeAPIWrapper struct {
+	KubeAPIer
```
AaronH88: Remove this, 

We satisfy the interface, not instantiate it directly

pkg/workceptor/kubernetes.go
```diff
@@ -113,6 +119,13 @@ func (ku KubeAPIWrapper) Get(ctx context.Context, clientset *kubernetes.Clientse
 }
 
 func (ku KubeAPIWrapper) Create(ctx context.Context, clientset *kubernetes.Clientset, namespace string, pod *corev1.Pod, opts metav1.CreateOptions) (*corev1.Pod, error) {
+	if pod.Annotations == nil {
+		pod.Annotations = make(map[string]string)
+	}
+
+	// Add default system annotations
+	pod.Annotations["receptor.redhat.com/created-by"] = "automation-mesh"
```
AaronH88, line 127: Magic strings,

Lets define constants at the top of the file please

AaronH88: Your defining pod.Annotations here but they are never being re referenced in this PR,

Why are they here? 😄 
Is this something left over from an old idea?

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +237,102 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
 	return inner
 }
 
+// podInfrastructureSuccess checks if the pod has either successfully started, is pending or is running, or has is successfully terminated.
+// Any other state is considered an infrastructure failure.
+func (ku KubeAPIWrapper) podInfrastructureSuccess(pod *corev1.Pod, containerName string) (bool, string, error) {
```
AaronH88: Their already is a function called `podRunningAndReady` 
How is this function different to `podRunningAndReady` and could `podRunningAndReady` be used instead?

AaronH88: If we are continuing with this function (depending on reply) please export this function and add it to the new interface I suggested earlier

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +237,102 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+	switch pod.Status.Phase {
+	case corev1.PodFailed:
+		for _, cs := range pod.Status.ContainerStatuses {
+			if cs.State.Terminated != nil && cs.Name == containerName {
+				switch cs.State.Terminated.ExitCode {
```
AaronH88: This is complex, 
something like `if exitCode != 0` will work just as well

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +237,102 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+				case 0:
+					break
+				default:
+					return false, fmt.Sprintf("pod reason %s container %s %s", pod.Status.Reason, cs.Name, cs.State.Terminated.Reason), nil
+				}
+			}
+		}
+
+		return true, "", nil
```
AaronH88: So in the case of `ExitCode == 0` we return true, "", nil ?

Thats not obvious by the name or description of the function 🤔

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +237,102 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
 	return inner
 }
 
+// podInfrastructureSuccess checks if the pod has either successfully started, is pending or is running, or has is successfully terminated.
+// Any other state is considered an infrastructure failure.
+func (ku KubeAPIWrapper) podInfrastructureSuccess(pod *corev1.Pod, containerName string) (bool, string, error) {
+	switch pod.Status.Phase {
+	case corev1.PodFailed:
+		for _, cs := range pod.Status.ContainerStatuses {
+			if cs.State.Terminated != nil && cs.Name == containerName {
```
AaronH88: If the containerName is not found we `return true, "", nil` 

I dont think this is what you meant

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +237,102 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+			}
+		}
+
+		return true, "", nil
+
+	case corev1.PodSucceeded:
+		return true, "", nil
+
+	case corev1.PodRunning, corev1.PodPending:
+		return true, fmt.Sprintf("pod phase: %s", pod.Status.Phase), fmt.Errorf("pod not in terminal state")
```
AaronH88: Its very odd to return true along with an error, 

can you think of a different approach here please?

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +237,102 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
 	return inner
 }
 
+// podInfrastructureSuccess checks if the pod has either successfully started, is pending or is running, or has is successfully terminated.
+// Any other state is considered an infrastructure failure.
+func (ku KubeAPIWrapper) podInfrastructureSuccess(pod *corev1.Pod, containerName string) (bool, string, error) {
+	switch pod.Status.Phase {
+	case corev1.PodFailed:
+		for _, cs := range pod.Status.ContainerStatuses {
```
AaronH88: I recommend doing this loop at the top of the function, finding or not finding the container first, before you go into any of this conditional logic, 

All this case, loop, if, case flows are tricky to read

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +237,102 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+		return false, fmt.Sprintf("unknown phase: %s", pod.Status.Phase), fmt.Errorf("invalid pod phase")
+	}
+}
+
+// podApplicationSuccess checks if the pod has successfully completed its application logic.
+// this is called after podInfrastructureSuccess has confirmed the pod is in a terminal state.
+func (ku KubeAPIWrapper) podApplicationSuccess(pod *corev1.Pod, containerName string) (bool, string, error) {
+	for _, cs := range pod.Status.ContainerStatuses {
+		if cs.Name == containerName {
+			if cs.State.Waiting != nil {
```
AaronH88: You check cs.State.Waiting != nil then immediately test for State.Terminated. In K8s, a container can only be in one state at a time—terminated containers have Waiting == nil. So this branch is dead/unreachable.

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +237,102 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+	default:
+		return false, fmt.Sprintf("unknown phase: %s", pod.Status.Phase), fmt.Errorf("invalid pod phase")
+	}
+}
+
+// podApplicationSuccess checks if the pod has successfully completed its application logic.
+// this is called after podInfrastructureSuccess has confirmed the pod is in a terminal state.
+func (ku KubeAPIWrapper) podApplicationSuccess(pod *corev1.Pod, containerName string) (bool, string, error) {
+	for _, cs := range pod.Status.ContainerStatuses {
+		if cs.Name == containerName {
```
AaronH88: Container not found returns (true, "", nil)

I dont think you meant this

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +237,102 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+		return true, fmt.Sprintf("pod phase: %s", pod.Status.Phase), fmt.Errorf("pod not in terminal state")
+
+	default:
+		return false, fmt.Sprintf("unknown phase: %s", pod.Status.Phase), fmt.Errorf("invalid pod phase")
+	}
+}
+
+// podApplicationSuccess checks if the pod has successfully completed its application logic.
+// this is called after podInfrastructureSuccess has confirmed the pod is in a terminal state.
+func (ku KubeAPIWrapper) podApplicationSuccess(pod *corev1.Pod, containerName string) (bool, string, error) {
```
AaronH88: error was defined as a return type, but this function only returns nil for it.

AaronH88: You describe in the comments that this function should only be used once the pod is in a terminal state, I recommend defensive programming and checking for this state first before continuing

AaronH88: I also recommend moving this function to the new interface i recommended earlier, and also exporting to allow for easy mocking.

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +237,102 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+					return false, fmt.Sprintf("container %s exited with code %d: %s", cs.Name, cs.State.Terminated.ExitCode, cs.State.Terminated.Reason), nil
+				}
+			}
+		}
+	}
+
+	return true, "", nil
+}
+
+func (ku KubeAPIWrapper) WaitForPodCompleted(pod *corev1.Pod, clientset kubernetes.Interface, timeout time.Duration) (*corev1.Pod, error) {
```
AaronH88: Lets also take in a context so we can exit early, 
Very important in looping code.

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +237,102 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+	interval := 1 * time.Second
```
AaronH88: Polling like this can open us up for API rate limiting and other unexpected results,
Kube API has a watch specifically for these cases.
https://pkg.go.dev/k8s.io/apimachinery/pkg/watch 

Please refactor this function to watch instead of poll please

pkg/workceptor/kubernetes.go
```diff
@@ -224,6 +237,102 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+
+// GetPodStatus checks if the pod has successfully completed its application logic and infrastructure is healthy.
+func (ku KubeAPIWrapper) GetPodStatus(pod *corev1.Pod) (bool, string, error) {
+	if pod == nil {
+		return false, "pod is nil", fmt.Errorf("pod is nil")
+	}
+
+	podRef := fmt.Sprintf("pod %s/%s", pod.Namespace, pod.Name)
+
+	infraOK, reason, err := ku.podInfrastructureSuccess(pod, "worker")
```
AaronH88: Magic "worker" string

pkg/workceptor/kubernetes.go
```diff
@@ -386,6 +495,10 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
 						podName,
 					)
 
+					if false { // deactivate new code TODO write integration tests
```
AaronH88: I have never seen an approach like this to adding dead/uncalled code, 
It isnt sitting too well with me, 
Can we delete it please, just leave the function uncalled

pkg/workceptor/kubernetes.go
```diff
@@ -902,12 +1015,50 @@ func (kw *KubeUnit) runWorkUsingLogger() {
 		return
 	}
 
-	// only transition from WorkStateRunning to WorkStateSucceeded if WorkStateFailed is set we do not override
+	pod, err := kw.KubeAPIWrapperInstance.Get(kw.GetContext(), kw.clientset, podNamespace, podName, metav1.GetOptions{})
+	if err != nil {
+		kw.GetWorkceptor().nc.GetLogger().Warning("Failed to retrieve pod for diagnostics: %v", err)
+	} else if false { // deactivate new code TODO write integration tests
```
AaronH88: same as above, please remove everything in the false block

pkg/workceptor/kubernetes.go
```diff
@@ -902,12 +1015,50 @@ func (kw *KubeUnit) runWorkUsingLogger() {
...
 
+func (kw *KubeUnit) CapturePodStatus(pod *corev1.Pod, stdoutSize int64, timeout time.Duration) (ok bool, err error) {
+	pod, err = kw.KubeAPIWrapperInstance.WaitForPodCompleted(pod, kw.clientset, timeout)
+	if err != nil {
+		kw.GetWorkceptor().nc.GetLogger().Debug("Job complete and pod is still running: %v", err)
+	}
+
+	ok, reason, err := kw.KubeAPIWrapperInstance.GetPodStatus(pod)
+	if err != nil {
+		reason := fmt.Sprintf("Pod diagnostics failed: %v reason %s", err, reason)
```
AaronH88: re using the same var name here is tricky to follow, please use a new var name

AaronH88 (changes requested):
mostly small syntax stuff, but one larger request to use a watcher instead of poll

pkg/workceptor/kubernetes.go
lranjbar: `ExitCode == 0` is always Success/Completed and `ExitCode != 0` is always a failure of some kind.

pkg/workceptor/kubernetes.go
lranjbar: This complexity comes from the Kube API unfortunately

If the `Container.State.Terminated == nil` then the exit code is also nil. Once `Container.State.Terminated` is populated so is the exit code. This check is to avoid panics on exit code.

The object looks like:

```
ContainerStates {
Pending: nil,
Running: nil,
Terminated: 
      ContainerState {
      Exit Code: 0
      Reason: "Success" }
}
```

or

```
ContainerStates {
Pending: nil,
Running: nil,
Terminated: 
    ContainerState {
    Exit Code: 1
    Reason: "Error" }
}
```

pkg/workceptor/kubernetes.go
lranjbar: 🤔 It's a good call out, we are working through the current hero function while avoiding changing `podRunningAndReady`. I think we should add it to our new interface.

pkg/workceptor/kubernetes.go
lranjbar: Adding some pod.Annotations was mostly to help out with troubleshooting bugs. At this point we weren't looking to act upon an annotation.

pkg/workceptor/kubernetes.go
lranjbar: I think we can drop the containerName check here

pkg/workceptor/kubernetes.go
lranjbar: Agreed this is probably more a case of an info log:

`return true, fmt.Infof("pod phase: %s", pod.Status.Phase), nil`

pkg/workceptor/kubernetes.go
lranjbar: I think we should name the new interface around pods `KubePodHelper` or something like this. It seems that we are trying to make an interface for the Pod and Container checks we need to do.

pkg/workceptor/kubernetes.go
arrestle: I had originally implemented this as a watch but couldn't get it to work, hence the loop. That said Lisa and I are taking another look at this.

pkg/workceptor/kubernetes.go
AaronH88: And the If statement right about this check `if cs.State.Terminated != nil` this looks to me like it will catch any panic states?

So that just leaves the `ExitCode`, which is a primitive and defaults to 0 not nil. So I still think because of the if condition above we can move this to `if cs.State.Terminated.ExitCode != 0` 

Am I missing something? 😄

lranjbar: Looks like I just needed to look at this with fresh eyes on Monday morning.

pkg/workceptor/kubernetes.go
```diff
@@ -397,6 +401,15 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
 
+					// timeout := int64(10)
+					// _, err = kw.CapturePodStatus(kw.Pod, stdout.Size(), &timeout)
+					// if err != nil {
+					// 	kw.GetWorkceptor().nc.GetLogger().Info("Detected error while retrieving pod status for %s/%s.",
+					// 		podNamespace,
+					// 		podName,
+					// 	)
+					// }
+
```
PabloHiro, line 414: Thoughts on removing comments from the code? If we are afraid it to lose it, we can leave it in a separate draft PR

pkg/workceptor/kubernetes.go
```diff
@@ -913,7 +926,21 @@ func (kw *KubeUnit) runWorkUsingLogger() {
...
+	// if err != nil {
+	// 	kw.GetWorkceptor().nc.GetLogger().Warning("Failed to retrieve pod for diagnostics: %v", err)
+	// }
+	// timeout := int64(10)
+	// ok, _ := kw.CapturePodStatus(pod, stdout.Size(), &timeout)
+	// if !ok {
+	// 	// If the pod did not succeed, we already updated the status to WorkStateFailed
+	// 	// and we can return early.
+	// 	return
+	// }
```
PabloHiro, line 937: Same as above

pkg/workceptor/pod.go
```diff
@@ -0,0 +1,139 @@
...
+
+func (kw *KubeUnit) CapturePodStatus(pod *corev1.Pod, stdoutSize int64, timeoutSeconds *int64) (bool, error) {
+	if pod == nil {
+		return false, fmt.Errorf("pod is nil")
+	}
+
+	if pod.Status.Phase == corev1.PodRunning || pod.Status.Phase == corev1.PodPending {
+		err := kw.WaitForPodCompleted(kw.GetContext(), pod, kw.clientset, timeoutSeconds)
+		if err != nil {
+			kw.GetWorkceptor().nc.GetLogger().Debug("Pod error detected while waiting for completion: %v", err)
```
PabloHiro: I would keep the pod name + pod namespace only in the logging calls of this function which is the one that calls the rest. 

That way we only need to ensure that we are consistent here.

pkg/workceptor/pod.go
```diff
@@ -0,0 +1,139 @@
...
+	return true, nil
+}
+
+// GetPodStatus checks if the pod has successfully completed its application logic and infrastructure is healthy.
+func (ku KubeUnit) GetPodStatus(pod *corev1.Pod) (bool, error) {
+	if pod == nil {
+		return false, fmt.Errorf("pod is nil")
+	}
+
+	podRef := fmt.Sprintf("pod %s/%s", pod.Namespace, pod.Name)
```
PabloHiro: I would move this one to `CapturePodStatus`, so the formatting logic of namespace+podname in the error is in a single place

pkg/workceptor/pod.go
```diff
@@ -0,0 +1,139 @@
...
+
+			if cs.State.Terminated.ExitCode != 0 { // exit code of 0 means success
+				return false, fmt.Errorf("container %s exited with code %d: %s", cs.Name, cs.State.Terminated.ExitCode, cs.State.Terminated.Reason)
+			}
+
+			return true, nil // container terminated with exit code of 0
+		}
+	}
+
+	return false, fmt.Errorf("pod %s/%s does not contain container %s", pod.Namespace, pod.Name, containerName)
```
PabloHiro: Same here. Line 76 does not report pod information, but line 83 does. Tracking which log lines have pod information or not makes the code hard to read.

pkg/workceptor/pod.go
```diff
@@ -0,0 +1,139 @@
...
+				return true, nil
+			case corev1.PodRunning, corev1.PodPending:
+				return true, nil
+			default:
+				return false, fmt.Errorf("unknown phase: %s", pod.Status.Phase)
+			}
+		}
+	}
+
+	return false, fmt.Errorf("pod %s/%s does not contain container %s", pod.Namespace, pod.Name, containerName)
```
PabloHiro: Same here. Line 110 does not report pod information, but line 115 does. Tracking which log lines have pod information or not makes the code hard to read.

pkg/workceptor/pod.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return err
+	}
+
+	for event := range watcher.ResultChan() {
+		if event.Type == watch.Error {
+			return event.Object.(error)
+		}
+		pod = event.Object.(*corev1.Pod)
+		fmt.Printf("%s: %s/%s (Phase: %s)\n", event.Type, pod.Namespace, pod.Name, pod.Status.Phase)
```
PabloHiro:
```suggestion
		return fmt.Printf("%s: %s/%s (Phase: %s)\n", event.Type, pod.Namespace, pod.Name, pod.Status.Phase)
```
Same here, would move the pod information formatting into a single place. Also this should be returned?

PabloHiro (changes requested):

pkg/workceptor/kubernetes.go
lranjbar: This is turning off the implementation so that it can be addressed in the following PR. @AaronH88 was not a fan of the `if false` so I commented out the code.

PabloHiro (approved):
LGTM, thank you for the hard work!

davemulford (approved):
Reviewed during a 1:1 call, talked through some of the testing methods. Looks great!

AaronH88 (changes requested):
Looking right now

pkg/workceptor/kubernetes.go
```diff
@@ -400,6 +403,15 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
 						podName,
 					)
 
+					// timeout := int64(10)
```
AaronH88, line 406: Take out commented out code please,

The tests are enough for now

pkg/workceptor/kubernetes.go
```diff
@@ -912,7 +924,21 @@ func (kw *KubeUnit) runWorkUsingLogger() {
 		return
 	}
 
-	// only transition from WorkStateRunning to WorkStateSucceeded if WorkStateFailed is set we do not override
+	// pod, err := kw.KubeAPIWrapperInstance.Get(kw.GetContext(), kw.clientset, podNamespace, podName, metav1.GetOptions{})
```
AaronH88, line 927: same as above
Take out commented out code please,

The tests are enough for now

pkg/workceptor/kubernetes.go
arrestle: @AaronH88  I have a conflict here because @lranjbar would like to keep the comment in the code so we know where it goes in the future.

arrestle:
Close this was merged because it's a duplicate of #1371 and #1372

PR #1333, opened 2025-06-04
Add a pod status check after EOF detection

lranjbar:
In researching the bug: [AAP-46484](https://issues.redhat.com/browse/AAP-46484) 
Jobs are in failed status with message "Receptor detail: Finished"

Receptor is returning a WorkStateSuccessful with a truncated stdout. This causes parsing JSON errors by ansible-runner.

The reproducer we have so far is shows that when a container exits with a failure exit code (non-zero exit code) receptor will pass the truncated stdout as a success. We can force this path by creating a container with low memory that causes k8s to OOMKill the container and pod.

This adds a check to grab the terminated state of the container and sends an error message if the container exited with a failure of some kind.

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -386,6 +386,27 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout 
 						podName,
 					)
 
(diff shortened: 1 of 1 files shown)


lranjbar:
Changes have been rolled into #1332

PR #1335, opened 2025-06-09
Update to Golang 1.23.9

lranjbar:
Updates to Golang version 1.23.9.


.github/workflows/pull_request.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: ["1.22"]
+        go-version: ["1.23"]
```
matoval: We should have both 1.22 and 1.23 here if we want to test both 1.22 and 1.23 in the PR checks

arrestle (approved):

matoval (approved):
lgtm

PR #1339, opened 2025-06-10
Remove unreachable code in workceptor/remote_work.go

lranjbar:
The variable `err` is always nil here because it is checked against `err != nil` on lines 347-351 and returned.

There is not another function populating `err` here. I think there was a Save() function that was called in the past but this is no longer called.


AaronH88 (approved):
LGTM

matoval (approved):
lgtm, this check was left behind in a previous PR

PR #1340, opened 2025-06-10
Remove fsnotify from TestMonitorLocalStatus

lranjbar:
This test is calling fsnotify mocks... so we should refactor this test. (or maybe even remove?)


PabloHiro (approved):
LGTM

AaronH88 (approved):
lgtm

PR #1342, opened 2025-06-13
Add get_sanitized_results function

davemulford:
This new function will print a warning and allow regular processing to continue. Warnings are printed under the following conditions.

- The work unit results string is empty.
- The work unit results string looks like JSON but doesn't parse correctly.
- The work unit results string is a mix of regular text and JSON fragments.

[AAP-47152](https://issues.redhat.com/browse/AAP-47152)

### DIFF
--- a/receptorctl/receptorctl/socket_interface.py
+++ b/receptorctl/receptorctl/socket_interface.py
@@ -19,11 +19,11 @@ def contains_json_like_fragments(s):
     """Determines if the input contains substrings that resemble JSON."""
 
     # Pattern to match balanced {...} or [...] in a rough and non-recursive way
-    object_pattern = r'({[^{}[\]]*})'
-    array_pattern = r'(\[[^\[\]{}]*\])'
+    object_pattern = r"({[^{}[\]]*})"
+    array_pattern = r"(\[[^\[\]{}]*\])"
 
     # Find all non-overlapping matches
-    matches = re.findall(f'{object_pattern}|{array_pattern}', s)
+    matches = re.findall(f"{object_pattern}|{array_pattern}", s)
 
     for match in matches:
         # match is a tuple because of regex alternation. Get the first non-empty match.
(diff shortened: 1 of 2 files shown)


receptorctl/receptorctl/socket_interface.py
```diff
@@ -277,3 +308,21 @@ def get_work_results(self, unit_id, startpos=0, return_socket=False, return_sock
...
+            return
+
+        try:
+            if text in ['', b'', '\n']:
+                print(f'WARNING: results from work unit were empty: {text}')
+            elif looks_like_json(text):
+                try:
+                    _ = json.loads(text)
+                except (ValueError, TypeError):
+                    print(f'WARNING: results appear to be valid JSON but did not parse correctly: {text}')
```
PabloHiro: The `get_work_results` function is called in this file of AWX:
https://github.com/ansible/awx/blob/3db2e04efe00c8fd0265de1f2a1f7388da95fd34/awx/main/tasks/receptor.py

It seems like all calls are captured in exceptions, so I would simply raise exceptions and not print anything.

receptorctl/receptorctl/socket_interface.py
```diff
@@ -277,3 +308,21 @@ def get_work_results(self, unit_id, startpos=0, return_socket=False, return_sock
             return sockfile
         else:
             return
+
+    def get_sanitized_results(self, text):
```
PabloHiro: I think @AaronH88 suggestion was to create a new function similar to `get_work_results`, instead of extending what it does, but your approach is fine as well imho.

PabloHiro (changes requested):

receptorctl/receptorctl/socket_interface.py
davemulford: Thanks Pablo! This clears up the logging question I had.

receptorctl/receptorctl/socket_interface.py
davemulford: Sure that makes sense. Since it was Aaron's idea, I'll wait for him to tell me if I should be recreating the existing function that can be called conditionally, or if the current approach is okay.

davemulford: Spoke with Aaron about this. The goal is to make a new function that callers can migrate to when they are ready, leaving the existing function untouched.

davemulford:
Closing this pull request as the `get_work_results` function won't know what the result is, as the caller is responsible for `read()`ing from the socket file returned by this function.

PR #1344, opened 2025-06-17
Fix a test that fails on go 1.24 due to crypto/rsa changes

dleehr:
Golang 1.24 returns a different error message than 1.23 when attempting to initialize a key with invalid bit length. This change updates the test to consider both error messages since we support 1.23 and 1.24. It was developed on a collaboration call.

### DIFF
--- a/pkg/certificates/ca_test.go
+++ b/pkg/certificates/ca_test.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"net"
 	"reflect"
+	"slices"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


matoval (approved):
lgtm, tested with go1.23.9 and go1.24.0

arrestle:
> lgtm, tested with go1.23.9 and go1.24.0

Me too. tested with go1.24.4 and go1.23.9 with a go clean -testcache in between. Don't have permissions to approve tho.

arrestle (commented):
LGTM

lranjbar (approved):

arrestle (approved):

PR #1345, opened 2025-06-17
AAP-46609 cli go unit test

arrestle:
Increases test cover of cli.go to 85.5%
Does not test Run() due to issues mocking /tmp files which will be discussed separately,
Including original tests (with additional mock expectations) as well as new ones.


pkg/certificates/ca_test.go
```diff
@@ -518,7 +517,14 @@ sa6wJArEJueVGTZbXePe5zaBUWAHdPX7DYeVISHO4WkOvEquwT1BwJ/v8KO77aKv
 2HcWw8NEqjF9Enb+ieyBD9Ds04NFFAaPDHA/8eWL+PXuMLtN2ALbomG1UcfsrnEL
 qI0vTc9/cZNK/A3W8rUFjat6jPUW1CsZm8yvJ3ORU01V4xVfZMRH3JzwB0WqRGVo
 FYeY7rWtiZX43Wq4szS4xWy7nlk0Jg==
------END PRIVATE KEY-----`)
+-----END PRIVATE KEY-----`
```
PabloHiro, line 521:
```suggestion
-----END PRIVATE KEY-----` // NOSONAR 
```
To make sonarqube happy, we will refactor this at some point so these are generated at test time, but for now...

PabloHiro (changes requested):

pkg/certificates/ca_test.go
PabloHiro: Dang, it did not work :( 

I am not sure in which line the comment needs to be added, maybe in the first one instead?

arrestle: I think it just needed a `make fmt`. it seems happier now.

PabloHiro: We are still getting:
image
:(

arrestle: Yup. trying something else now. Trying to wrap it in a function. That worked.

PabloHiro (approved):

lranjbar (approved):

PR #1346, opened 2025-06-18
feat: tests for cli.go SignReq

PabloHiro:
Changes:
1. Make `SignReq` mockable
2. Move the logic of `SignReqConfig.Run()` into a separate method named `ValidateAndSign`

New coverage can be seen here:
https://app.codecov.io/gh/ansible/receptor/pull/1346/blob/pkg/certificates/cli.go?dropdown=coverage

```
github.com/ansible/receptor/pkg/certificates/cli.go:173:        SignReq                 89.4%
github.com/ansible/receptor/pkg/certificates/cli.go:247:        ValidateAndSign         100.0%
github.com/ansible/receptor/pkg/certificates/cli.go:267:        Run                     0.0%
```

Run is 0% but does not contain any logic.


pkg/certificates/cli_test.go
```diff
@@ -314,10 +409,14 @@ func TestSignReq(t *testing.T) {
 					EXPECT().
 					ReadFile(gomock.Eq(positiveCaCrtPath)).
 					Return(setupGoodCaCertificatePEMData(), nil).
-					Times(1)
+					AnyTimes()
```
AaronH88, line 412: Can I ask, why the move to `AnyTimes()` ?

AaronH88:
Overall looking good, 
One question around the use of AnyTimes vs Times(1)

pkg/certificates/cli_test.go
PabloHiro: Because in certain test scenarios, the path is not even reached. Example:
```
		{
			name: "Error in CA Path",
			args: args{
				opts:      &positiveCertOptions,
				caCrtPath: invalidPath,
				caKeyPath: positiveCaKeyPath,
				reqPath:   positiveReqPath,
				certOut:   positiveCertOut,
				verify:    true,
			},
			wantErr: true,
		},
``` 
In this scenario, we don't get to read from `caKeyPath` or `reqPath`, because we already failed at `caCrtPath`. We could mock conditionally based on the scenario, but it adds complexity to the test and makes it really hard to read and extend it with new test cases, I tried it before ending up using `AnyTimes()`.

The trade-off between being strict vs having an easy to read and maintain tests, is not worth it imho.

AaronH88 (approved):
LGTM 

Thanks for this 🎉

arrestle (approved):
YES! Came to the same conclusion on the AnyTimes()

PR #1347, opened 2025-06-19
Add more tests to types

AaronH88:


pkg/types/main_test.go
```diff
@@ -37,3 +43,172 @@ func TestMainInitNodeID(t *testing.T) {
...
 	}
 }
+
+func TestNodeCfgInitEmptyID(t *testing.T) {
+	originalHostname := ""
+	defer func() {
+		if originalHostname != "" {
+			os.Setenv("HOSTNAME", originalHostname)
+		}
+	}()
```
matoval: originalHostname var and defer func doesn't seem to needed. I might be missing something but I removed it and the test functions with no issues.

matoval (changes requested):
I left one requested change, also it's not in the code you changed but on line 36 in the error message there is a typo. exected should be expected

lranjbar (approved):
This looks good after the removal of the `originalHostname`. I think that it makes sense to do the localhost cases in a follow up PR. That's going to be involved from what I'm seeing here.

pkg/types/main_test.go
AaronH88: removed, ty

PabloHiro (approved):

matoval (approved):
lgtm

PR #1348, opened 2025-06-19
Add tests in the runProtocol to prove that the connection is not remo…

AaronH88:
…ved in the event of a canceled context


matoval (approved):
lgtm

PR #1357, opened 2025-06-21
Handle EOF by check if pod is ready, Update TestKubeLoggingWithReconn…

matoval:
Handle EOF by check if pod is ready, Update TestKubeLoggingWithReconnect to test edge cases and increase coverage


AaronH88 (approved):
LGTM

I have no concerns around the extra kube call. It is at most 1 extra call per pod, so it should not hit any rate limiting. OK to merge

davemulford (approved):
LGTM

PR #1358, opened 2025-06-23
Tests for udp_proxy.go

PabloHiro:
AAP-45827

Changes:

* Move the logic inside the loop of `runNetceptorToUDPInbound` into `processInboundPacket`
* Move the logic inside the loop of `runUDPToNetceptorOutbound` into `processOutboundPacket`
* Make both `runNetceptorToUDPInbound` and `runUDPToNetceptorOutbound` mockable (for follow up PRs)
* Removed `*logger.ReceptorLogger` from the function arguments, as it is not being used.

Also changed:
```
if addr != expectedAddr { // First is net.Addr, second is netceptor.Addr
```

Into:
```
if addr.String() != expectedAddr.String() {
```

I suspect this function is not called in production environments, or otherwise it would have logged:
```
Received packet from unexpected source
```
every time it is called.

### DIFF
--- a/pkg/services/udp_proxy_test.go
+++ b/pkg/services/udp_proxy_test.go
@@ -2,6 +2,8 @@ package services
 
 import (
 	"errors"
+	"fmt"
+	"net"
 	"testing"
 
 	"github.com/ansible/receptor/pkg/logger"
@@ -9,6 +11,7 @@ import (
 	"github.com/ansible/receptor/pkg/netceptor/mock_netceptor"
 	mock_net_interface "github.com/ansible/receptor/pkg/services/interfaces/mock_interfaces"
 	"github.com/ansible/receptor/pkg/services/mock_services"
+	"github.com/ansible/receptor/pkg/utils"
 	"go.uber.org/mock/gomock"
 )
 
@@ -24,8 +27,6 @@ func setUpMocks(ctrl *gomock.Controller) (*mock_services.MockNetcForUDPProxy, *m
 }
 
 func TestUDPProxyServiceInbound(t *testing.T) {
-	ctrl := gomock.NewController(t)
-	defer ctrl.Finish()
 	var mockNetceptor *mock_services.MockNetcForUDPProxy
 	var mockNetter *mock_net_interface.MockNetterUDP
 	var mockUDPConn *mock_net_interface.MockUDPConnInterface
@@ -77,8 +78,12 @@ func TestUDPProxyServiceInbound(t *testing.T) {
 	}
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
 			mockNetceptor, mockNetter, mockUDPConn, mockPacketCon = setUpMocks(ctrl)
-			tc.calls()
+			if tc.calls != nil {
+				tc.calls()
+			}
 			err := UDPProxyServiceInbound(mockNetceptor, tc.host, tc.port, tc.node, tc.service, mockNetter)
 			if tc.expectErr {
 				if err == nil {
@@ -94,8 +99,6 @@ func TestUDPProxyServiceInbound(t *testing.T) {
 }
 
 func TestUDPProxyServiceOutbound(t *testing.T) {
-	ctrl := gomock.NewController(t)
-	defer ctrl.Finish()
 	var mockNetceptor *mock_services.MockNetcForUDPProxy
 	var mockNetter *mock_net_interface.MockNetterUDP
 	var mockPacketCon *mock_netceptor.MockPacketConner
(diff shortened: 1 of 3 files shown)


pkg/services/udp_proxy_test.go
```diff
@@ -150,3 +153,111 @@ func TestUDPProxyServiceOutbound(t *testing.T) {
...
+	expectedAddr := netceptor.Addr{}
+	expectedAddr.SetNetwork("tcp")
+	expectedAddr.SetNode("127.0.0.1")
+	expectedAddr.SetService("2222")
+	expectedUDPAddr := &net.UDPAddr{IP: net.ParseIP("127.0.0.1"), Port: 2222}
+	unexpectedUDPAddr := &net.UDPAddr{IP: net.ParseIP("127.0.0.1"), Port: 3333}
+	tests := []struct {
+		name           string
+		expectedAddr   netceptor.Addr
+		calls          func(mockPacketConner *mock_netceptor.MockPacketConner, mockUDPConnInterface *mock_net_interface.MockUDPConnInterface)
```
lranjbar:
```suggestion
		calls          func()
```
This can be simplified to just `func()` and you can declare these as variables in the test. Example is on lines 99-101

Then the following test cases just have the func() heading makes it easier to read.

matoval (approved):
lgtm

pkg/services/udp_proxy_test.go
```diff
@@ -150,3 +153,111 @@ func TestUDPProxyServiceOutbound(t *testing.T) {
...
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+			mockPacketConner := mock_netceptor.NewMockPacketConner(ctrl)
+			mockUDPConnInterface := mock_net_interface.NewMockUDPConnInterface(ctrl)
```
lranjbar: Let's use the `setUpMocks()` function here

pkg/services/udp_proxy_test.go
```diff
@@ -150,3 +153,111 @@ func TestUDPProxyServiceOutbound(t *testing.T) {
...
+		})
+	}
+}
+
+func TestProcessOutboundPacket(t *testing.T) {
+	destinationAddr := &net.UDPAddr{IP: net.ParseIP("192.168.0.1"), Port: 2222}
+	tests := []struct {
+		name           string
+		Addr           netceptor.Addr
+		calls          func(mockPacketConner *mock_netceptor.MockPacketConner, mockUDPConnInterface *mock_net_interface.MockUDPConnInterface)
```
lranjbar:
```suggestion
		calls          func()
```
Same as above. Let's simplify to `func()` here.

pkg/services/udp_proxy_test.go
```diff
@@ -150,3 +153,111 @@ func TestUDPProxyServiceOutbound(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+			mockPacketConner := mock_netceptor.NewMockPacketConner(ctrl)
+			mockUDPConnInterface := mock_net_interface.NewMockUDPConnInterface(ctrl)
+			tt.calls(mockPacketConner, mockUDPConnInterface)
```
lranjbar:
```suggestion
			if tt.calls != nil {
			     tt.calls()
			}
```
Need to add a check if `tt.calls()` is nil. This is to safeguard the tests against panics when we are adding new test cases into the slice.

pkg/services/udp_proxy_test.go
```diff
@@ -150,3 +153,111 @@ func TestUDPProxyServiceOutbound(t *testing.T) {
...
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+			mockPacketConner := mock_netceptor.NewMockPacketConner(ctrl)
+			mockUDPConnInterface := mock_net_interface.NewMockUDPConnInterface(ctrl)
```
lranjbar: Lets use `setUpMocks()` here

lranjbar (changes requested):
Handful of things:
- We should check if tt.calls() is nil before calling it.
- Lets use our setUpMocks() function
- Lets simplify the func() signature in the tests

😃

pkg/services/udp_proxy_test.go
PabloHiro: Yeah... I changed this along with moving this inside the test loop:
```
	ctrl := gomock.NewController(t)
	defer ctrl.Finish()
```
Because I was getting less flakiness this way. You are right and defining the pointers at test scope is the way to go.

pkg/services/udp_proxy_test.go
PabloHiro: Done, thank you!

pkg/services/udp_proxy_test.go
PabloHiro: Done, thank you!

pkg/services/udp_proxy_test.go
PabloHiro: Done, thank you!

pkg/services/udp_proxy_test.go
PabloHiro: Done, and updated all tests in this file with that. Thank you!

lranjbar (approved):

PR #1360, opened 2025-06-23
Add new mocks to generate.go

lranjbar:
When adding new mocks we need to add them to the `generate.go` file so that they can be generated using `make generate`.

I added the missing line and ran `make generate-clean && make generate` using mockgen `0.5.2`


PabloHiro (approved):

arrestle (approved):
GO generate! LGTM.

PR #1366, opened 2025-06-26
Test LoglevelCfgInit & SanitizedLog

arrestle:


pkg/logger/logger_test.go
```diff
@@ -71,6 +71,151 @@ func TestLogLevelToNameWithError(t *testing.T) {
...
+				return
+			}
+			if err != nil {
+				t.Fatalf("unexpected error for level %q: %v", tt.level, err)
+			}
+			got := logger.GetLogLevel()
+			if got != tt.wantLevel {
+				t.Errorf("expected log level %d, got %d", tt.wantLevel, got)
+			}
+		})
```
matoval, line 108: Since logLevel is a global var, we should set the default logLevel once each test case is done.

pkg/logger/logger_test.go
```diff
@@ -71,6 +71,151 @@ func TestLogLevelToNameWithError(t *testing.T) {
...
+					t.Errorf("logger delegate called with wrong args: %v %v %v", level, format, v)
+				}
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			oldLogger := logger.GetRegisteredLogger()
+			defer logger.RegisterLogger(oldLogger)
```
matoval, line 179: Can you explain why we are setting the oldLogger after each test?

matoval (changes requested):
I have a question and a suggested change

PabloHiro (approved):
LGTM

pkg/logger/logger_test.go
arrestle: Yup. Added. `logger.SetGlobalLogLevel(logger.InfoLevel) // Reset to default after test`

pkg/logger/logger_test.go
arrestle: Short answer: Because otherwise it panics.

Longer:
-  The logger uses a global variable `var logger MessageFunc)` that maintains state across tests.
-   If one test modifies this state  via `logger.RegisterLogger(...)` and doesn't reset it, subsequent tests inherit the modified state.
-  Example: Test A registers a custom logger; Test B then unintentionally uses Test A's logger, causing erratic behavior. and, of course, a panic when the MessageFunc is nil, as happened in my case.

davemulford (approved):
LGTM

fincamd (approved):
LGTM

pkg/logger/logger_test.go
matoval: Makes sense, thank you for explaining it in detail!

matoval (approved):
lgtm

PR #1370, opened 2025-07-03
Test for CreatePod and runWorkUsingLogger in kubernetes.go

AaronH88:


pkg/workceptor/kubernetes_test.go
```diff
@@ -2171,3 +2171,683 @@ func TestKubeUnit_SetFromParams(t *testing.T) {
...
+				assert.NoError(t, err, tt.description)
+				if tt.validateResult != nil {
+					tt.validateResult(t, kubeUnit)
+				}
+			}
+		})
+	}
+}
+
+func TestKubeUnit_LoggingMethodSelection(t *testing.T) {
```
PabloHiro: If we ever refactor `runWorkUsingLogger` to split the logic this test is left dangling because it does not call any line of real code. I would honestly remove it or move the logic out of `runWorkUsingLogger` into a separate function and test it here.

We should not be afraid of modifying the source code. We know it needs to be done at some point to make it healthier. Better do it now since you have the context fresh in your head.

PabloHiro (changes requested):
LGTM except for that test. 

Nit: I would also ask you to title the PR with something a little bit more precise since the title is what is going to be used as a commit later. Something like: "Test for CreatePod and runWorkUsingLogger in kubernetes.go"

pkg/workceptor/kubernetes_test.go
AaronH88: ok, ill remove this test for now.

I agree about not being afraid to modify source code, with a caveat, this area of the code base had little to no coverage. Adding tests will give any refactor a lot more confidence, so I want to get as much coverage in as possible before refactoring to give that confidence of no regressions 😄

PabloHiro (approved):
LGTM!

arrestle:
> LGTM except for that test.
> 
> Nit: I would also ask you to title the PR with something a little bit more precise since the title is what is going to be used as a commit later. Something like: "Test for CreatePod and runWorkUsingLogger in kubernetes.go"

Can we also add the Jira AAP-xxxxx number? Not the link, just the number. Trying to determine if this is for AAP-47456 or another ticket.

arrestle (changes requested):
Please also add the Jira AAP-xxxxx number. Not the link, just the number. Trying to determine if this is for AAP-47456 or another ticket.

AaronH88:
> Please also add the Jira AAP-xxxxx number. Not the link, just the number. Trying to determine if this is for AAP-47456 or another ticket.

Nope, nothing to do with any of that work, just the unit test epic, AAP-49244

arrestle (approved):
LGTM.

matoval (approved):

PR #1371, opened 2025-07-03
AAP-46619 Get pod container status

arrestle:
This pull request addresses concerns about the size and commit history of [PR #1332](https://github.com/ansible/receptor/pull/1332). We are consolidating these helper functions in this PR and moving the WaitForPodCompleted function to a separate pull request.

1. PodHealthy - Determine if the Pod and Container are both healthy. Eventually to be called after WaitForBodCompleted function determines the Pod has "settled" into a state that has diagnostic information.

1. PodContainerHealthy - Typically called by PodHealthy to make sure the "worker" container is healthy but can also be invoked externally.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,6 +33,11 @@ require (
 )
 
 require (
+	github.com/pkg/errors v0.9.1 // indirect
+	gopkg.in/evanphx/json-patch.v4 v4.12.0 // indirect
+)
+
+require (
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
 	github.com/stretchr/testify v1.10.0
 )
--- a/go.mod
+++ b/go.mod
@@ -33,11 +33,6 @@ require (
 )
 
 require (
-	github.com/pkg/errors v0.9.1 // indirect
-	gopkg.in/evanphx/json-patch.v4 v4.12.0 // indirect
-)
-
-require (
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
 	github.com/stretchr/testify v1.10.0
 )
--- a/pkg/workceptor/pod.go
+++ b/pkg/workceptor/pod.go
@@ -7,7 +7,6 @@ import (
 )
 
 type KubePodStateHelper interface {
-	GetPodStatus(pod *corev1.Pod) (bool, error)
 	PodHealthy(pod *corev1.Pod, containerName string) (bool, error)
 	PodContainerHealthy(pod *corev1.Pod, containerName string) (bool, error)
 }
--- a/pkg/workceptor/pod.go
+++ b/pkg/workceptor/pod.go
@@ -30,7 +30,7 @@ func (kw KubeUnit) PodContainerHealthy(pod *corev1.Pod, containerName string) (b
 	if foundContainer == nil {
 		return false, fmt.Errorf("pod does not contain container %s", containerName)
 	}
-	if foundContainer.State.Waiting != nil { // means it is waiting, so application logic has not completed yet. Normal behavior when job completes successfully.
+	if foundContainer.State.Waiting != nil { // means it is waiting, so application logic has not completed yet.
 		return false, fmt.Errorf("container %s is waiting: %s %s", containerName, foundContainer.State.Waiting.Reason, foundContainer.State.Waiting.Message)
 	}
 	if foundContainer.State.Terminated == nil { // means it is waiting or running, so application logic has not completed yet. Normal behavior when job completes successfully.
--- a/pkg/workceptor/pod.go
+++ b/pkg/workceptor/pod.go
@@ -51,10 +51,10 @@ func (kw KubeUnit) PodHealthy(pod *corev1.Pod, containerName string) (bool, erro
 		return false, fmt.Errorf("pod is nil")
 	}
 
-	containerDiag := fmt.Sprintf("container %s is healthy", containerName)
+	var containerDiag string = ""
 	containerOk, containerError := kw.PodContainerHealthy(pod, containerName)
 	if containerError != nil {
-		containerDiag = fmt.Sprintf("%v", containerError)
+		containerDiag = fmt.Sprintf(" %v", containerError)
 	}
 
 	switch pod.Status.Phase {
@@ -64,11 +64,11 @@ func (kw KubeUnit) PodHealthy(pod *corev1.Pod, containerName string) (bool, erro
 			podError = fmt.Errorf("%s message: %s", podError, pod.Status.Message)
 		}
 
-		return false, fmt.Errorf("%s %v", podError, containerDiag)
+		return false, fmt.Errorf("%s%s", podError, containerDiag)
 
 	case corev1.PodSucceeded, corev1.PodRunning, corev1.PodPending:
 		return containerOk, containerError
 	default:
-		return false, fmt.Errorf("unknown phase: %s %s", pod.Status.Phase, containerDiag)
+		return false, fmt.Errorf("unknown phase: %s%s", pod.Status.Phase, containerDiag)
 	}
 }
--- a/pkg/workceptor/pod.go
+++ b/pkg/workceptor/pod.go
@@ -43,45 +43,24 @@ func (kw KubeUnit) PodHealthy(pod *corev1.Pod, containerName string) (bool, erro
 		return false, fmt.Errorf("pod is nil")
 	}
 
-	for _, cs := range pod.Status.ContainerStatuses {
-		// Check if this is the container we care about first
-		if cs.Name == containerName {
-			switch pod.Status.Phase {
-			case corev1.PodFailed:
-				podError := fmt.Errorf("pod failed with reason: %s", pod.Status.Reason)
-				if pod.Status.Message != "" {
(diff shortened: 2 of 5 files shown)


PabloHiro (approved):

pkg/workceptor/pod.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// PodContainerHealthy checks if the pod has successfully completed its application logic.
+// this is called after podInfrastructureSuccess has confirmed the pod is in a terminal state.
+func (kw KubeUnit) PodContainerHealthy(pod *corev1.Pod, containerName string) (bool, error) {
+	if pod == nil {
+		return false, fmt.Errorf("pod is nil")
+	}
+
+	for _, cs := range pod.Status.ContainerStatuses {
```
AaronH88: Lets make this logic a little easier,

First lets find the container with the correct name, then exit the for loop, 
This will reduce the branching complexity and make testing easier.

pkg/workceptor/pod.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// PodInfrastructureSuccess checks if the pod has either successfully started, is pending or is running, or has is successfully terminated.
+// Any other state is considered an infrastructure failure.
+func (kw KubeUnit) PodHealthy(pod *corev1.Pod, containerName string) (bool, error) {
+	if pod == nil {
+		return false, fmt.Errorf("pod is nil")
+	}
+
+	for _, cs := range pod.Status.ContainerStatuses {
```
AaronH88: Like above, lets reduce the branching complexity,
We have a switch case, in a if statement in a loop. 
If we just find the container by name first, then exit the loop, this function will be a lot simpler and easier to test.

pkg/workceptor/pod.go
```diff
@@ -0,0 +1,87 @@
...
+				return false, fmt.Errorf("unknown phase: %s", pod.Status.Phase)
+			}
+		}
+	}
+
+	return false, fmt.Errorf("pod does not contain container %s", containerName)
+}
+
+// GetPodStatus checks if the pod has successfully completed its application logic and infrastructure is healthy.
+func (kw KubeUnit) GetPodStatus(pod *corev1.Pod) (bool, error) {
```
AaronH88: I am struggling to see the functionality of this function. 
It seems to just wrap `PodHealthy` rather than add anything to it. 

The `pod == nil` check already exists in `PodHealthy`
And line 83 just returns what `PodHealthy` returns, 
Can you expand on why this is needed please? Or even if it can be removed/modify `PodHealthy` for this to be removed?

AaronH88:
Mostly this looks great and thanks so much for all the hard work on it so far. 

I think some complexity could be reduced and I question 1 function. But thank you so much for reducing the size of this change and making it possible to do a good code review!

pkg/workceptor/pod_test.go
```diff
@@ -0,0 +1,390 @@
...
+	Status: corev1.PodStatus{
+		Phase:  corev1.PodFailed,
+		Reason: "Pod OOMKilled",
+		ContainerStatuses: []corev1.ContainerStatus{
+			{
+				Name: "worker",
+				State: corev1.ContainerState{
+					Terminated: &corev1.ContainerStateTerminated{
+						ExitCode: 137,
+						Reason:   "Container OOMKill",
```
lranjbar:
```suggestion
		Reason: "OOMKilled",
		ContainerStatuses: []corev1.ContainerStatus{
			{
				Name: "worker",
				State: corev1.ContainerState{
					Terminated: &corev1.ContainerStateTerminated{
						ExitCode: 137,
						Reason:   "OOMKill",
```
This object matches the actual K8s objects that are returned. While I understand the want to change these it would be confusing in a real debugging scenario. K8s does not include Pod/Container in the returned reason so we should not expect it.

lranjbar (changes requested):

pkg/workceptor/pod.go
lranjbar: The difference was that `PodHealthy` and `ContainerHealthy` used to filter on the container name provided and `GetPodStatus` did not. 

Perhaps `GetPodStatus` is not needed in the future though? 🤔

arrestle: Removing for now. I believe it'll make more sense when we add the WaitForPodComplete.

pkg/workceptor/pod.go
arrestle: See what you think of this version!

pkg/workceptor/pod.go
```diff
@@ -0,0 +1,72 @@
+package workceptor
+
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+type KubePodStateHelper interface {
+	GetPodStatus(pod *corev1.Pod) (bool, error)
```
AaronH88: No longer needed in this PR 😄

pkg/workceptor/pod.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	if foundContainer.State.Terminated.ExitCode != 0 { // exit code of 0 means success
+		return false, fmt.Errorf("container %s exited with code %d: %s", containerName, foundContainer.State.Terminated.ExitCode, foundContainer.State.Terminated.Reason)
+	}
+
+	return true, nil // container terminated with exit code of 0
+}
+
+// PodInfrastructureSuccess checks if the pod has either successfully started, is pending or is running, or has is successfully terminated.
```
AaronH88: Seems the comment is left over from an old iteration 😄

go.mod
```diff
@@ -32,6 +32,11 @@ require (
 	k8s.io/client-go v0.31.3
 )
 
+require (
+	github.com/pkg/errors v0.9.1 // indirect
```
AaronH88: Please run `go mod tidy` and `go mod vendor`

pkg/workceptor/pod.go
```diff
@@ -0,0 +1,72 @@
...
+		if cs.Name == containerName {
+			foundContainer = &pod.Status.ContainerStatuses[i]
+
+			break
+		}
+	}
+	if foundContainer == nil {
+		return false, fmt.Errorf("pod does not contain container %s", containerName)
+	}
+	if foundContainer.State.Terminated == nil { // means it is waiting or running, so application logic has not completed yet. Normal behavior when job completes successfully.
```
AaronH88: This assumes non-terminated containers are healthy, but containers in `Waiting` state with `CrashLoopBackOff`, `ImagePullBackOff`, or similar errors would incorrectly return true. This could mask important failure conditions 😄

pkg/workceptor/pod.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// PodInfrastructureSuccess checks if the pod has either successfully started, is pending or is running, or has is successfully terminated.
+// Any other state is considered an infrastructure failure.
+func (kw KubeUnit) PodHealthy(pod *corev1.Pod, containerName string) (bool, error) {
+	if pod == nil {
+		return false, fmt.Errorf("pod is nil")
+	}
+
+	containerDiag := fmt.Sprintf("container %s is healthy", containerName)
```
AaronH88: Building these error messages is tricky to follow, 
using two different vars, and 6 possible assignments is making me need more coffee 😂 

Can we simplify this please and just create a specific message per return

pkg/workceptor/pod_test.go
```diff
@@ -0,0 +1,351 @@
...
+					Running: &corev1.ContainerStateRunning{
+						StartedAt: metav1.Now(),
+					},
+				},
+			},
+		},
+	},
+}
+
+func TestPodHeathy(t *testing.T) {
```
AaronH88, line 278: Can we add some waiting, and imagepullback off error tests here please 😄

AaronH88 (changes requested):
Really close, 🕺

lranjbar (approved):

pkg/workceptor/pod.go
matoval: I agree, presetting the containerDiag here is confusing and hard to track what message is in containerDiag when it's used

pkg/workceptor/pod.go
arrestle: see foundContainer = &pod.Status.ContainerStatuses[i]

pkg/workceptor/pod.go
arrestle: Resolved with some help from lisa

pkg/workceptor/pod_test.go
arrestle: Added these per Lisa and Aaron

1. podImagePullBackOff 
2. podErrImagePull 
3. podCrashLoopBackOff

pkg/workceptor/pod.go
AaronH88: Can you remove this one please 😄

pkg/workceptor/pod.go
arrestle: It's used below so I  need to define it.  Changed to `var containerDiag string = ""`

pkg/workceptor/pod.go
arrestle: Added 

1. podImagePullBackOff
2. podErrImagePull
3. podCrashLoopBackOff

AaronH88 (approved):
lgtm, 

as soon as everything passes feel free to hit the button 🕺

PR #1376, opened 2025-07-07
Additional unit tests for netceptor/conn.go

lranjbar:

PR #1378, opened 2025-07-16
Add unit tests for netceptor.Listener and removes panic from Listener.Accept()

lranjbar:
Partial completion of AAP-47012. 
Since #1376 is getting quite big I'm breaking these changes out.

Adds unit tests for `netceptor.Listener`.
- `Listener.Addr()`
- `Listener.Accept()`
- `Listener.Close()`

`Listener.Accept()` would panic if the accept channel was closed. This is because `AcceptChan` uses pointers for AcceptResult (ie: `AcceptChan: make(chan *AcceptResult)`). When the `AcceptChan` is closed the pointer for the `AcceptResult` becomes `nil`.

### DIFF
--- a/pkg/netceptor/conn.go
+++ b/pkg/netceptor/conn.go
@@ -38,6 +38,12 @@ type QuicConnectionForConn interface {
 	quic.Connection
 }
 
+type QuicListenerForListener interface {
+	Accept(ctx context.Context) (quic.Connection, error)
+	Addr() net.Addr
+	Close() error
+}
+
 type AcceptResult struct {
 	Conn net.Conn
 	Err  error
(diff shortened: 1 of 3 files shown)


pkg/netceptor/conn.go
```diff
@@ -292,7 +304,11 @@ func (li *Listener) acceptLoop(ctx context.Context) {
 func (li *Listener) Accept() (net.Conn, error) {
 	select {
 	case ar := <-li.AcceptChan:
-		return ar.Conn, ar.Err
+		if ar == nil {
+			return nil, fmt.Errorf("listener closed")
```
matoval: I suggested we differentiate this closed AcceptChan error message from the DoneChan error message.
Something like "listener accept channel closed"

matoval (changes requested):
Looks good, I just have one suggestion

matoval (approved):
lgtm

davemulford (approved):
Looks great!

PR #1379, opened 2025-07-16
[TRIVIAL] Move Netceptor interfaces to pkg/netceptor/interfaces.go

lranjbar:
# Current direction

We've decided to move interfaces in a package into a file called interfaces.go for each package. I've repurposed this PR for the Netceptor package.

**Changes**
- [‎pkg/netceptor/interfaces.go](https://github.com/ansible/receptor/pull/1379/files#diff-285d9c3e06cf594a470c125ad164be24f17dc67c8d6dbd2995141f80e3787a73) Created this file and moved all existing package interfaces to it.
- [generate.go](https://github.com/ansible/receptor/pull/1379/files#diff-daddc66424f674dc57b411e63fce1f1bd231a27a77e39d25c781b4a8ca6f6062) Added interfaces.go to this file and removed the old entries.
- Removed interfaces from their original files
- Regenerated mocks 

# Original direction

This PR adds an interface and mocks for Netceptor. As we are adding more unit tests we are creating multiple interfaces for Netceptor in various folders to make mocks. It might be time to have one mock for Netceptor. This PR does not swap out the existing interfaces for the new one because that change is very large.

I have found interfaces for Neceptor in the following files:
- https://github.com/ansible/receptor/blob/devel/pkg/controlsvc/interfaces.go#L20-L29 
- https://github.com/ansible/receptor/blob/devel/pkg/controlsvc/controlsvc.go#L45-L48
- https://github.com/ansible/receptor/blob/devel/pkg/netceptor/ping.go#L11-L16
- https://github.com/ansible/receptor/blob/devel/pkg/netceptor/packetconn.go#L33-L47
- https://github.com/ansible/receptor/blob/devel/pkg/services/command.go#L20-L23
- https://github.com/ansible/receptor/blob/devel/pkg/services/tcp_proxy.go#L17-L21
- https://github.com/ansible/receptor/blob/devel/pkg/services/udp_proxy.go#L15-L20
- https://github.com/ansible/receptor/blob/devel/pkg/workceptor/workceptor.go#L32-L38

These interfaces could be replaced with this new netceptor interface. 

I also cast the Netceptor struct to the new interface. This is to make sure that Netceptor implements all the functions in the Netceptor interface as we expect. At the moment it is a no-op. In the future if we make unexpected changes to existing functions to Netceptor this will cause Golang to catch it for us.


AaronH88:
I get your idea here, 

Can you explain in your own words why we chose to go down the route of making a smaller interface for the specific Netceptor functions used rather than this one interface you are proposing? 

Maybe some Pro's and Con's 😄

lranjbar:
I don't know the history of the decision but here is my take.

**Smaller interfaces**:
Pros:
- Each specific interface is easier to understand
- Ability to expose specific functions for mocks

Cons:
- Multiple interfaces in different folders is hard to keep track of.
- When to use each interface is unclear.
- Allows for inconsistent use of the interfaces in the codebase.
- No interface casting available.

**One interface**:
Pros:
- One interface for Netceptor, easy to keep track of.
- Everybody would be using the same interface for Netceptor so it is clear when to use.
- Interface casting available.

Cons:
- Larger interfaces can be unwieldy.

Starting with smaller interfaces makes sense. But over time there has been inconsistent usage across the code base for these smaller Netceptor interfaces. It's not really clear when to use one or make a new one.

lranjbar:
Direction from team discussion is to move our interfaces into `interfaces.go` file in each folder.

pkg/netceptor/interface.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"context"
+	"crypto/tls"
+	"sync"
+	"time"
+
+	"github.com/ansible/receptor/pkg/logger"
+	"github.com/ansible/receptor/pkg/utils"
+)
+
```
davemulford: Can doc comments be written here to explain why the interface exists and what problem it solves? Knowing the historical reason for why the project ended up with a single, large interface will make it easy for newcomers to understand it. A short comment for each function would also help, if that's not too much.

I see this as a task where AI can do most of the work.

davemulford (changes requested):
I really like what you're doing here. One (hopefully) small ask for doc comments. Feel free to not do that if (A) it's asking too much, or (B) you feel it's unnecessary.

Great work!

AaronH88:
> Direction from team discussion is to move our interfaces into `interfaces.go` file in each folder.

Do you want to do that in this PR or open a new one? 

A new PR will give you the opportunity to make smaller PR's

lranjbar:
> > Direction from team discussion is to move our interfaces into `interfaces.go` file in each folder.
> 
> Do you want to do that in this PR or open a new one?
> 
> A new PR will give you the opportunity to make smaller PR's

For this PR. I'm going to repurpose for the Netceptor package changes

pkg/netceptor/interface.go
lranjbar: Let me know what you think of the comments. 😃

pkg/netceptor/interfaces.go
```diff
@@ -0,0 +1,280 @@
...
+// These interfaces follow the Interface Segregation Principle by providing only
+// the methods that specific components actually need from the main Netceptor instance,
+// making the code more testable and reducing coupling.
+
+// NetcForPacketConn defines the subset of Netceptor methods needed by PacketConn.
+//
+// Implemented by:
+//   - Netceptor (main Netceptor struct implements this through its methods)
+//
+// Mock:
```
AaronH88: I like most of the comments, up until Mock, I am thinking from Mock onwards this becomes a pain to maintain with diminishing returns, 
Most ide's we use can go to references, definitions etc 😄

AaronH88 (dismissed):
Mostly good, just the comments in interface are too much in my opinion 😄

pkg/netceptor/interfaces.go
davemulford: I agree that most of the comments are great. A couple of things:

1. I agree with Aaron that the mock references can be removed.
2. Some of the comments list specific line numbers. Those may get outdated over time as the code changes. I'd be for removing the line numbers to prevent that.

The "Used by" and "Tests" sections are super useful in helping to navigate a code base of this size.

Great job and thank you!

davemulford (approved):
The new and simpler comments look great. Thank you!

arrestle (approved):
Nice work @lranjbar !

PR #1385, opened 2025-07-21
Add unit tests for Listener.SendResult()

lranjbar:
# PR Changes
- Made SendResult() public
- Adding tests for SendResult()

Some of these tests are flaky due to the behavior of SendResult():

The current implementation of SendResult() has non-deterministic behavior. During context cancellation it can either send nothing or an AcceptResult(Conn: nil, Err: nil). It's random on which it will pick.

Update: It was decided in team discussion to go with option 1 because this behavior is easier to test.

# Team Discussion Starter:
## **Non-deterministic Behavior of `select`**

When multiple cases in a `select` statement are ready simultaneously, Go randomly chooses one. This means:

- If `AcceptChan` has buffer space AND `ctx.Done()` is closed, Go might choose either case
- The behavior becomes race-dependent and non-deterministic

## The Core Problem

The `select` statement as written doesn't prioritize context cancellation or done channel over sending to `AcceptChan`. This creates unpredictable behavior where:

1. **Sometimes** it respects cancellation/done signals
2. **Sometimes** it sends the result anyway
3. The behavior depends on timing and Go's random case selection

## Current Behavior Analysis

The method attempts to:
1. Send an `AcceptResult` to the `AcceptChan`
2. But give up immediately if the context is cancelled
3. But give up immediately if the listener is done

However, due to `select` semantics, if multiple cases are ready, any one of them might be chosen.

### Option 1: Check conditions before select (Predictable)
```go
func (li *Listener) SendResult(ctx context.Context, conn net.Conn, err error) {
	// Check cancellation first
	select {
	case <-ctx.Done():
		return
	case <-li.DoneChan:
		return
	default:
	}
	
	// Then try to send
	select {
	case li.AcceptChan <- &AcceptResult{Conn: conn, Err: err}:
	case <-ctx.Done():
	case <-li.DoneChan:
	}
}
```

**Pros:** Predictable behavior - always respects cancellation/done signals
**Cons:** Two select statements, slightly more complex

### Option 2: Accept current behavior (Non-deterministic but reasonable)
```go
// Keep current implementation but document the behavior
func (li *Listener) SendResult(ctx context.Context, conn net.Conn, err error) {
	// Note: This method attempts to send to AcceptChan but may be cancelled
	// by context or done channel. The behavior is non-deterministic when
	// multiple cases are ready simultaneously.
	select {
	case <-ctx.Done():
		return
	case li.AcceptChan <- &AcceptResult{Conn: conn, Err: err}:
	case <-li.DoneChan:
	}
}
```

**Pros:** Simple, current behavior is actually reasonable for most use cases
**Cons:** Non-deterministic, harder to test reliably

### Option 3: Priority-based approach (Context first)
```go
func (li *Listener) SendResult(ctx context.Context, conn net.Conn, err error) {
	// Always check cancellation first
	if ctx.Err() != nil {
		return
	}
	
	select {
	case li.AcceptChan <- &AcceptResult{Conn: conn, Err: err}:
	case <-ctx.Done():
	case <-li.DoneChan:
	}
}
```

**Pros:** Context cancellation always respected
**Cons:** Doesn't handle done channel with same priority

## Impact on Production Code

The current behavior is actually reasonable for production use:
- In normal operation, `AcceptChan` is consumed by `Accept()` calls
- Context cancellation and done channel closure are exceptional cases
- The non-deterministic behavior only matters in edge cases during shutdown

## Recommendations for Team Discussion

1. **Current behavior acceptable?** The existing implementation works well in practice
2. **Test strategy:** Should we test the actual behavior (non-deterministic) or change the implementation?
3. **Documentation:** Should we document the non-deterministic behavior?
4. **Consistency:** Are there similar patterns elsewhere in the codebase?

## Test Strategy Options

1. **Accept non-determinism:** Test that the method doesn't block and sometimes sends results
2. **Force determinism:** Use unbuffered channels to force blocking and test cancellation behavior
3. **Change implementation:** Make behavior predictable and test accordingly

### DIFF
--- a/pkg/netceptor/conn.go
+++ b/pkg/netceptor/conn.go
@@ -204,19 +204,13 @@ func (s *Netceptor) ListenAndAdvertise(service string, tlscfg *tls.Config, tags 
 }
 
 func (li *Listener) SendResult(ctx context.Context, conn net.Conn, err error) {
-	// Check cancellation first
 	select {
 	case <-ctx.Done():
 		return
-	case <-li.DoneChan:
-		return
-	default:
-	}
-
-	// Then try to send
-	select {
-	case li.AcceptChan <- &AcceptResult{Conn: conn, Err: err}:
-	case <-ctx.Done():
+	case li.AcceptChan <- &AcceptResult{
+		Conn: conn,
+		Err:  err,
+	}:
 	case <-li.DoneChan:
 	}
 }
--- a/pkg/netceptor/conn.go
+++ b/pkg/netceptor/conn.go
@@ -204,13 +204,19 @@ func (s *Netceptor) ListenAndAdvertise(service string, tlscfg *tls.Config, tags 
 }
 
 func (li *Listener) SendResult(ctx context.Context, conn net.Conn, err error) {
+	// Check cancellation first
 	select {
 	case <-ctx.Done():
 		return
-	case li.AcceptChan <- &AcceptResult{
-		Conn: conn,
-		Err:  err,
-	}:
+	case <-li.DoneChan:
+		return
+	default:
+	}
+
+	// Then try to send
+	select {
+	case li.AcceptChan <- &AcceptResult{Conn: conn, Err: err}:
+	case <-ctx.Done():
 	case <-li.DoneChan:
 	}
 }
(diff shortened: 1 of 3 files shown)


davemulford (approved):
LGTM. The tests are great!

arrestle (approved):
We've talked about this one. I'm approving.

PR #1399, opened 2025-08-06
Fix Kubernetes Retry Counter Resetting

lranjbar:
# Kubernetes Retry Counter Fix

## Original Problem
The Kubernetes retry logic had a bug where the retry counter was not being reset properly. Users reported seeing logs like:
```
2025-08-05T01:00:44 Detected Error: EOF for pod. Will retry 5 more times.
2025-08-05T05:00:45 Detected Error: EOF for pod. Will retry 5 more times.  
2025-08-05T09:00:45 Detected Error: EOF for pod. Will retry 5 more times.
2025-08-05T13:00:45 Detected Error: EOF for pod. Will retry 4 more times.  ← Should be 5
2025-08-05T17:00:45 Detected Error: EOF for pod. Will retry 3 more times.  ← Should be 5
```

## Solution

- Rename retries variables to be more clear: 
  - remainingRetries -> retryGetLogStream
  - retries -> retryGetPod
- Reset retryGetLogStream counter on successful write to the stream
- Add defensive programming checks on line 387 so that we don't panic on checking conditions on a nil erroredPod.
- Instead of breaking loop immediately continue so that retryGetLogStream can decrement
- After retries are exhausted check if we have an EOF and return
- Pass the error to stdoutErr if its not EOF and break to the outer loop (to allow for reconnect)
- Updated existing tests to have validation on log messages to decrement the retry counters
- Added new test **retry_counter_resets_after_write_succeeds** that simulates retries with 2 connections.


arrestle (approved):

pkg/workceptor/workunitbase_test.go
```diff
@@ -86,12 +86,9 @@ func TestIsPending(t *testing.T) {
 
 func setUp(t *testing.T) (*gomock.Controller, workceptor.BaseWorkUnit, *workceptor.Workceptor, *mock_workceptor.MockNetceptorForWorkceptor, *logger.ReceptorLogger) {
 	ctrl := gomock.NewController(t)
-
 	mockNetceptor := mock_workceptor.NewMockNetceptorForWorkceptor(ctrl)
-
-	// attach logger to the mock netceptor and return any number of times
-	logger.SetGlobalLogLevel(4)
```
matoval, line 93: I'm curious why this was removed? Was it unnecessary or was it causing an issue with the tests?

AaronH88:
Can we add tests for this please to prove that the fix works,

I have renamed the retry vars to improve readability, removed the old redundant reset, and moved the set to above the connection attempt

arrestle (changes requested):
looks like some logic to resolve flakey unit tests has worked it's way in here.

pkg/workceptor/workunitbase_test.go
arrestle: I believe this is related to flakey unit tests related to a shared logger.

lranjbar: The unit tests are creating a data race over setting the log level.

lranjbar:
> Can we add tests for this please to prove that the fix works,
> 
> I have renamed the retry vars to improve readability, removed the old redundant reset, and moved the set to above the connection attempt

I have added log validation for the existing tests that trigger retry. 😄

pkg/workceptor/kubernetes.go
```diff
@@ -410,39 +418,44 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
 						}
 					}
+
+					return // EOF means we're done, no need for new connection cycle
 				}
 
-				return
+				*stdoutErr = err
+
+				break // Non-EOF error - break to outer loop for new connection cycle
```
matoval, line 458: Is it possible that we get an err and line isn't an empty string? If we hit that scenario we would want to read the last line instead of break.

lranjbar:
Closing in lieu of #1400

PR #1400, opened 2025-08-07
Update KubeLoggingWithReconnect to better handle EOF and other possible errors, pod/container disconnections

matoval:
- Added test for RunWorkUsingLogger setting Finished status on EOF from Kubernetes API log stream
- Added container state awareness: Now checks if containers are Running/Waiting/Terminated when stream errors occur, enabling smarter retry decisions
  - Implemented EOF + pod-not-ready recovery: When EOF occurs but pod isn't ready, the function retries with exponential backoff to allow temporary pod issues to resolve
  - Fixed safety issues: Added nil pointer checks for pod status and eliminated duplicate final line processing that could cause data corruption
  - Enhanced retry management: Added proper retry counter decrements with bounds checking to prevent infinite loops while preserving retry attempts for different error scenarios
  - Improved control flow: Used labeled breaks to cleanly transition between retry strategies (stream-level retries vs. full reconnection)


AaronH88:
Good first shot at this! 

So the bit thats missing for me is checking that the status is set to `Finished` 

This will require more of an integration style test, so I am currently looking at how that might work

AaronH88:
I think we want some way to prove that this test results in setting the status to finished on line 935 in kubernetes.go, 

So ill update this test to also check for stdinErr, 

Then we need to expand our tests to look at RunWorkUsingLogger and try to hit 935 under the conditions outlined in this test

pkg/workceptor/kubernetes.go
```diff
@@ -315,12 +315,15 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
 	var sinceTime time.Time
 	var err error
+	var retryGetLogStream int
 	podNamespace := kw.Pod.Namespace
 	podName := kw.Pod.Name
 
 	retries := 5
+	pervDelay, curDelay := 0, 1
+	pervPodDelay, curPodDelay := 0, 1
+	pervContDelay, curContDelay := 0, 1
```
dleehr: Are these for previous and current delays? Might want to fix that typo.

pkg/workceptor/kubernetes.go
```diff
@@ -378,47 +385,75 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
-							if err != nil {
-								*stdoutErr = fmt.Errorf("writing final line to stdout: %s", err)
-								kw.GetWorkceptor().nc.GetLogger().Error("Error writing final line to stdout: %s", err)
-
-								return
+					for _, containerStatus := range erroredPod.Status.ContainerStatuses {
+						if containerStatus.Name == WorkerContainerName {
+							switch {
+							case containerStatus.State.Running != nil:
+								break
```
dleehr: When we have a `Running` state do we need a delay before breaking?

I think we discussed this scenario, and there's two cases when there's a stream reading error and the container state is `Running`

1. Process is **actually** still running and the stream read error was an EOF due to timeout (e.g. the default 4 hour timeout of [`streaming-connection-idle-timeout`](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/)) or network interruption
2. Process is done and the EOF was the signal that the output is done. Here we expect the container state will be `Terminated` very soon.

In the first case, if there was some network issue, we should probably delay before attempting to reconnect so that we improve our chances.

In the second case, we probably want the delay too. I did some tests with a local cluster and found that sometimes the container state changes to `Terminated` before the EOF comes, and sometimes the final EOF comes first. When the EOF came first, it was about 1 second:

## EOF comes first
```
...
2025/08/09 00:03:47 [2025-08-09T00:03:47.291593-04:00] Creating pod: hello
2025/08/09 00:03:47 [2025-08-09T00:03:47.304638-04:00] Pod created successfully: hello
2025/08/09 00:03:47 [2025-08-09T00:03:47.312672-04:00] Pod Event - Type: MODIFIED, Phase: Pending, Message: 
2025/08/09 00:03:47 [2025-08-09T00:03:47.312672-04:00] Container Status - hello: Waiting (ContainerCreating)
2025/08/09 00:03:48 [2025-08-09T00:03:48.614041-04:00] Pod Event - Type: MODIFIED, Phase: Running, Message: 
2025/08/09 00:03:48 [2025-08-09T00:03:48.614041-04:00] *** CONTAINER STATUS CHANGE *** hello: Waiting (ContainerCreating) -> Running
2025/08/09 00:03:48 [2025-08-09T00:03:48.816054-04:00] Starting log stream for container: hello
2025/08/09 00:03:48 [2025-08-09T00:03:48.81915-04:00] Log - hello: 2025-08-09T04:03:47.786548962Z Hello, Kubernetes!
2025/08/09 00:03:57 [2025-08-09T00:03:57.837171-04:00] *** LOG STREAM EOF *** Container: hello (clean EOF)
2025/08/09 00:03:58 [2025-08-09T00:03:58.647058-04:00] Pod Event - Type: MODIFIED, Phase: Running, Message: 
2025/08/09 00:03:58 [2025-08-09T00:03:58.647058-04:00] *** CONTAINER STATUS CHANGE *** hello: Running -> Terminated (exit code: 0, reason: Completed)
2025/08/09 00:03:59 [2025-08-09T00:03:59.70546-04:00] Pod Event - Type: MODIFIED, Phase: Succeeded, Message: 
2025/08/09 00:04:00 [2025-08-09T00:04:00.367385-04:00] Pod completed with phase: Succeeded
```

## Terminated comes first

```
2025/08/09 00:01:30 [2025-08-09T00:01:30.15744-04:00] Creating pod: hello
2025/08/09 00:01:30 [2025-08-09T00:01:30.173662-04:00] Pod created successfully: hello
2025/08/09 00:01:30 [2025-08-09T00:01:30.194424-04:00] Pod Event - Type: MODIFIED, Phase: Pending, Message: 
2025/08/09 00:01:30 [2025-08-09T00:01:30.194424-04:00] Container Status - hello: Waiting (ContainerCreating)
2025/08/09 00:01:31 [2025-08-09T00:01:31.139974-04:00] Pod Event - Type: MODIFIED, Phase: Running, Message: 
2025/08/09 00:01:31 [2025-08-09T00:01:31.139974-04:00] *** CONTAINER STATUS CHANGE *** hello: Waiting (ContainerCreating) -> Running
2025/08/09 00:01:31 [2025-08-09T00:01:31.183557-04:00] Starting log stream for container: hello
2025/08/09 00:01:31 [2025-08-09T00:01:31.189312-04:00] Log - hello: 2025-08-09T04:01:30.648530348Z Hello, Kubernetes!
2025/08/09 00:01:41 [2025-08-09T00:01:41.178954-04:00] Pod Event - Type: MODIFIED, Phase: Running, Message: 
2025/08/09 00:01:41 [2025-08-09T00:01:41.178954-04:00] *** CONTAINER STATUS CHANGE *** hello: Running -> Terminated (exit code: 0, reason: Completed)
2025/08/09 00:01:41 [2025-08-09T00:01:41.207129-04:00] *** LOG STREAM EOF *** Container: hello (clean EOF)
2025/08/09 00:01:42 [2025-08-09T00:01:42.262184-04:00] Pod Event - Type: MODIFIED, Phase: Succeeded, Message: 
2025/08/09 00:01:43 [2025-08-09T00:01:43.245523-04:00] Pod completed with phase: Succeeded
```

So the feedback here is I think we should delay in the `Running` case. Either it's due to an interruption of some kind or it's a timing issue, and a delay is good for both.

pkg/workceptor/kubernetes.go
```diff
@@ -378,47 +385,75 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
+										containerStatus.State.Terminated.ExitCode,
+									)
+
+									return
+								}
+							default:
+								kw.GetWorkceptor().nc.GetLogger().Info("%s pod could not get container state", podNamespace)
 							}
+
+							break
```
dleehr: Can we extract a helper function out of this code that returns the container state and exit code (if available)? I think that would make this a lot easier to read.

pkg/workceptor/kubernetes.go
```diff
@@ -378,47 +385,75 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
 					kw.GetWorkceptor().nc.GetLogger().Debug("Error getting pod after reading stream: '%s'", kubeErr)
 				}
-				for _, condition := range erroredPod.Status.Conditions {
-					if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionTrue {
-						podConditionReady = true
+
+				if erroredPod != nil {
+					for _, condition := range erroredPod.Status.Conditions {
+						if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionTrue {
+							podConditionReady = true
```
dleehr: I think extracting this to helper function would make this logic easier to understand (similar feedback on the container statuses loop)

pkg/workceptor/kubernetes.go
dleehr: If we hit an EOF, see that the container is still running, do we restart the log stream from the beginning? Looks like we have a `sinceTime` variable that keeps that from happening.

pkg/workceptor/kubernetes.go
```diff
@@ -378,47 +385,75 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
+							switch {
+							case containerStatus.State.Running != nil:
+								break
+							case containerStatus.State.Waiting != nil:
+								kw.GetWorkceptor().nc.GetLogger().Debug("Container in %s pod is in waiting state, with waiting reason: %s", containerStatus.Name, containerStatus.State.Waiting.Reason)
+
+								time.Sleep(time.Second * time.Duration(curContDelay))
+								pervContDelay, curContDelay = curContDelay, pervContDelay+curContDelay
+
+								break streamLoop
```
dleehr: What's the difference between breaking back to `streamLoop` and breaking with no label?  Does breaking to `streamLoop` assume we can keep using the existing stream?

pkg/workceptor/kubernetes.go
```diff
@@ -378,47 +385,75 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
+						if containerStatus.Name == WorkerContainerName {
+							switch {
+							case containerStatus.State.Running != nil:
+								break
+							case containerStatus.State.Waiting != nil:
+								kw.GetWorkceptor().nc.GetLogger().Debug("Container in %s pod is in waiting state, with waiting reason: %s", containerStatus.Name, containerStatus.State.Waiting.Reason)
+
+								time.Sleep(time.Second * time.Duration(curContDelay))
+								pervContDelay, curContDelay = curContDelay, pervContDelay+curContDelay
+
```
dleehr: Handling a waiting container in this loop interesting. I think it would make more sense to handle that earlier - there's no reason to stream logs from a container until it starts running, and it won't go from running back to waiting, but I think that's best addressed in a future refactor.

We actually _do_ check the container state earlier in `podRunningAndReady`:

https://github.com/ansible/receptor/blob/ef6d9e50d5804ae1ec9bad363d877bbdc4ee3baf/pkg/workceptor/kubernetes.go#L205-L218

but it looks like we only delay or fail if that wait is because of an image pull issue. So I think it's fine to check it here. I wonder how often we'd get an EOF from logs while state is waiting.

pkg/workceptor/kubernetes.go
```diff
@@ -378,47 +385,75 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
+								if containerStatus.State.Terminated.ExitCode != 0 {
+									kw.GetWorkceptor().nc.GetLogger().Debug("Container in %s pod has terminated, with terminated exit code: %v", containerStatus.Name, containerStatus.State.Terminated.ExitCode)
+									*stdoutErr = fmt.Errorf("detected Error: %s for pod %s/%s. Pod has terminated, with terminated exit code: %v", err,
+										podNamespace,
+										podName,
+										containerStatus.State.Terminated.ExitCode,
+									)
+
+									return
+								}
```
dleehr: So this will handle a terminated state with a failure exit code and `return`

How do we handle a terminated state with a successful exit code? It sounds like that's the happy path (container is finished, process exited successfully, log stream is EOF). Can we do that here?

dleehr (changes requested):
Awesome progress 🎉 . I have some specific feedback to extract a couple things to helper functions, a typo, and a couple places where I'm asking for a bit more info. But I think this is going to help a lot 💯 

I've also been trying to reproduce some of these cases with an actual kubernetes cluster so we can see what happens, will share that on Monday.

pkg/workceptor/kubernetes.go
matoval: For the Fibonacci back off, in order to add the previous and current together.

pkg/workceptor/kubernetes.go
matoval: If we have a `Running` container and get EOF we break out of the `ContainerStatuses` loop and go to the `podConditionReady` check if the pod isn't ready we delay there or if it is ready we continue and  if `retryGetLogStream`  is greater than 0 we delay on 453.

pkg/workceptor/kubernetes.go
matoval: breaking to `streamLoop` will read the next line in the stream, now that I think about it could a container with a `waiting` state have a payload to read? I was assuming it wouldn't but if it could we should read the value in `line`.

pkg/workceptor/kubernetes.go
matoval: If we get `Terminated` state and a successful exit code we continue the function and check if the pod is ready, in the happy paths I tested the pod stays ready and we retry 5 times.
The out currently looks like this:
```
INFO 2025/08/09 17:10:01 Detected EOF for pod Test_Namespace/Test_Name.
INFO 2025/08/09 17:10:02 Detected EOF for pod Test_Namespace/Test_Name.
INFO 2025/08/09 17:10:03 Detected EOF for pod Test_Namespace/Test_Name.
INFO 2025/08/09 17:10:05 Detected EOF for pod Test_Namespace/Test_Name.
INFO 2025/08/09 17:10:08 Detected EOF for pod Test_Namespace/Test_Name.
INFO 2025/08/09 17:10:08 Detected Error: EOF for pod Test_Namespace/Test_Name. Will retry 1 more times.
ERROR 2025/08/09 17:10:38 Error reading from pod Test_Namespace/Test_Name: EOF
```

pkg/workceptor/kubernetes.go
```diff
@@ -315,12 +315,15 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
 	defer streamWait.Done()
 	var sinceTime time.Time
 	var err error
+	var retryGetLogStream int
 	podNamespace := kw.Pod.Namespace
 	podName := kw.Pod.Name
 
 	retries := 5
+	prevDelay, curDelay := 0, 1
```
AaronH88, line 324: Lets start at 1, 1 on all these so we increase right away 😄

pkg/workceptor/kubernetes.go
```diff
@@ -360,6 +366,7 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
 
 		// read from logstream
 		streamReader := bufio.NewReader(logStream)
+	streamLoop:
 		for *stdinErr == nil { // check between every line read to see if we need to stop reading
```
AaronH88: Due to all this refactoring, is seems like *stdinErr is no longer a good var to use for our loops

pkg/workceptor/kubernetes.go
```diff
@@ -360,6 +366,7 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
 			line, err := streamReader.ReadString('\n')
 			if err != nil {
```
AaronH88, line 376: After this check, we also check the context, 

I think that needs to be reworked a little, as the error message and logic for finishing no longer make sense

pkg/workceptor/kubernetes.go
```diff
@@ -378,47 +385,75 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
-								*stdoutErr = fmt.Errorf("writing final line to stdout: %s", err)
-								kw.GetWorkceptor().nc.GetLogger().Error("Error writing final line to stdout: %s", err)
-
-								return
+					for _, containerStatus := range erroredPod.Status.ContainerStatuses {
+						if containerStatus.Name == WorkerContainerName {
+							switch {
+							case containerStatus.State.Running != nil:
+								break
+							case containerStatus.State.Waiting != nil:
```
AaronH88: Will we ever be in a waiting state?

pkg/workceptor/kubernetes.go
AaronH88: If we know the pod is terminated, why would we continue trying to read the logs? 

To me, it makes more sense to just double check we got the last line and then exit. What do you think?

pkg/workceptor/kubernetes.go
```diff
@@ -378,47 +385,75 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
 				if kubeErr != nil {
 					kw.GetWorkceptor().nc.GetLogger().Debug("Error getting pod after reading stream: '%s'", kubeErr)
 				}
-				for _, condition := range erroredPod.Status.Conditions {
-					if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionTrue {
-						podConditionReady = true
+
+				if erroredPod != nil {
```
AaronH88: Due to all these refactors, the var name `erroredPod` no longer makes sense and is confusing, lets change it 😄

pkg/workceptor/kubernetes.go
AaronH88: Do we now care about the pod ready status? 

Or should we instead focus only on the container status?

I am inclined to remove this pod status check and focus just on the container checks, what do you think?

pkg/workceptor/kubernetes.go
lranjbar: Yes containers stay in waiting state until they have launched. The most common way this happens in practice is being unable to pull the image for the container. (ErrImagePull)

pkg/workceptor/kubernetes.go
lranjbar: A container in waiting state should not have a payload to read because it hasn't fully launched yet. :)

pkg/workceptor/kubernetes.go
dleehr: > in the happy paths I tested the pod stays ready and we retry 5 times.

Hmm. I ran a check locally and I saw the pod Ready change to `False` just before the container's `Terminated` state, so that's a a different pattern.

But I think that's beside the point, I like Aaron's suggestion. Once we have an `EOF` and the container state is `Terminated`, that should be the signal to get the last line and exit. 

If we get an `EOF` and see the container is `Terminated`, that means our container is done running in kubernetes and we have reached the end of the stream. So let's check the exit code to determine if it's success or failure and exit. I don't think there's any reason to keep looping (or checking other things)

pkg/workceptor/kubernetes.go
dleehr: I called this out in my [review](https://github.com/ansible/receptor/pull/1400#discussion_r2265040277), asking the same question - Once we're in this loop, is it even possible for the container state to be `Waiting`? . Sounds like it is possible (since our check for `Waiting` only delays for `ErrImagePull`, see link).

So it's OK to check it, since we weren't before, and the backoff is a good idea. Logging should help us understand next time if we're getting into this loop prematurely.

pkg/workceptor/kubernetes.go
dleehr: @matoval understood, but let me provide a bit more context to the question.

Here in the code, we (A) previously opened the stream connection successfully (`kw.kubeLoggingConnectionHandler()` above) and (B) encountered an error trying to read the next line from that stream (Possibly `EOF`, possibly something else)

My understanding of your changes is that if we `break` (no label), we jump back and establish a new stream by calling `kw.kubeLoggingConnectionHandler()` again. That makes sense for the `Running` state - we got an error which might signal the work is done. So first we check if the work is actually done. If see it's still running, we interpret the error as a temporary connection interruption, discard our errored connection, and reconnect.

But for containers in `Waiting`, we `break streamLoop`, so we jump to a point in the code that reuses the existing `streamReader` and connection to retry the same read. Is there something about a `Waiting` state that suggests we should be able to reuse the existing connection after an error? Is it even possible to get a streamReader error from a container that's still `Waiting`?

The risk I see here is that we don't know how often we get into this loop (it might be never) with a log stream failure on a container in the `Waiting` state. So if we add logic to handle the `Waiting`  that retries with the same errored connection, that might never recover and lead to more early errors.

pkg/workceptor/kubernetes.go
```diff
@@ -351,94 +356,114 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
-		for *stdinErr == nil { // check between every line read to see if we need to stop reading
+		// TODO: removed stdinErr, not set in the loop, doesn't change after first check above
+		for { // check between every line read to see if we need to stop reading
 			line, err := streamReader.ReadString('\n')
 			if err != nil {
-				if kw.GetContext().Err() == context.Canceled {
-					kw.GetWorkceptor().nc.GetLogger().Info(
-						"Context was canceled while reading logs for pod %s/%s. Assuming pod has finished",
+				// First check if the error is not EOF, if error is not EOF set error and mark the job as failed.
+				if err != io.EOF {
```
dleehr: I do not think we should immediately fail the job if we get an error (non-EOF) attempting to read a line from the log stream.

We now know that we'll EOFs for jobs that are still running, but I think it's a mistake to interpret any non-EOF error as a failed job. I think we were retrying those before, right?

pkg/workceptor/kubernetes.go
dleehr: Follow-up on the `Waiting` check. I see in the most recent commit we're not checking for `Waiting` anymore, and I think that's fine.

I had some questions myself of it was actually possible to get into this state (log stream did connect successfully but returned an error on read), and it does not seem to be possible.

This check in the kubernetes source code https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_pods.go#L1561-L1570 shows that the call to get logs from a `Waiting` container will not succeed.

So it's not possible for this case to happen. If we established the log stream connection, the container must not be `Waiting`.

pkg/workceptor/kubernetes.go
dleehr: Updated the other thread and checked the kubernetes source code: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_pods.go#L1561-L1570. Basically, if the container is still in `Waiting` when we try to connect the logs, that call will fail. This logic (now deleted) would be after we already successfully connected to the log stream, which means the container must not have been `Waiting`.

pkg/workceptor/kubernetes.go
matoval: Yes, we were retrying before. The question is, if a non EOF error at this point in the code is recoverable. If it's not recoverable then failing the job makes sense if it is recoverable then we should retry.

I'll try and find a recoverable non EOF error.

dleehr: The (non-EOF) error is from an attempt to read a log line over an open connection. I can imagine many reasons that may fail (it's a network connection 😉) that are recoverable, which are not related to the state of the container.

matoval: If we get a non EOF error is the assumption that is a network connection issue? Retry back up to the mainLoop 5 times for the connection issue to resolve, otherwise fail the job.

dleehr: I think that's a good strategy for encountering a non-`EOF` error here. Honestly I wouldn't *assume* it's always network error, but I would expect network errors to be common.

pkg/workceptor/kubernetes.go
```diff
@@ -338,115 +343,174 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
 			if err != nil {
-				if kw.GetContext().Err() == context.Canceled {
-					kw.GetWorkceptor().nc.GetLogger().Info(
-						"Context was canceled while reading logs for pod %s/%s. Assuming pod has finished",
+				// First check if the error is not EOF, if error is not EOF retry 5 times if error persists set error and mark the job as failed.
+				if err != io.EOF {
+					retryGetLogStream--
+					if retryGetLogStream > 0 {
+						kw.GetWorkceptor().nc.GetLogger().Info(
+							"Detected Error: %s for pod %s/%s that is in a running state. Will retry %d more times.",
```
dleehr:
```suggestion
							"Detected non-EOF Error: %s for pod %s/%s. Will retry %d more times.",
```

At this point we haven't checked that the pod (or container) is in a running state, so I think logging that would be confusing. I suggest we report that this is a non-EOF error.

pkg/workceptor/kubernetes.go
```diff
@@ -338,115 +343,174 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
+
+						time.Sleep(time.Second * time.Duration(curDelay))
+						prevDelay, curDelay = curDelay, prevDelay+curDelay
+
+						continue mainLoop
+					}
+
+					*stdoutErr = err
+					kw.GetWorkceptor().nc.GetLogger().Error(
+						"Unexpected error while reading logs for pod %s/%s. Error: %s",
```
dleehr: Same feedback as above to align the error message that it's a `non-EOF` error. I can see myself grepping for `non-EOF` 🙂 .

It would also  be really helpful when reviewing logs if the error we log in `Info()` above (during retries) had the exact same format as this Error when we give up. Specifically, I'd like to see this one explicitly indicate that we've exhausted all retries and not continuing.

pkg/workceptor/kubernetes.go
```diff
@@ -338,115 +343,174 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
+				// This is a recoverable EOF, so we attempt to reconnect with a back-off.
+				// In BOTH cases, we have a simular approach, wait 1-2 seconds and check if the container status has changed.
+
+				podDetails, kubeErr := kw.KubeAPIWrapperInstance.Get(kw.GetContext(), kw.clientset, podNamespace, podName, metav1.GetOptions{})
 				if kubeErr != nil {
-					kw.GetWorkceptor().nc.GetLogger().Debug("Error getting pod after reading stream: '%s'", kubeErr)
+					// Their are many reasons why the kube api might not be able to get the pod,
+					// This does not mean their is a problem just yet.
+					// Lets try to get the pod again, max 5 times, and decide.
+					kw.GetWorkceptor().nc.GetLogger().Debug("Error getting pod after reading stream: '%s' , continuing try to get pod up to 5 more times.", kubeErr)
```
dleehr: Is there a reason this is `Debug()`? The log for when we retry a `non-EOF` is `Info()`. Seems like these should be consistent.

pkg/workceptor/kubernetes.go
```diff
@@ -338,115 +343,174 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
+				}
+
+				switch {
+				case containerState.Running != nil:
+					// We got EOF but pod is running, is this because we checked too fast? Will it turn into a terminated state soon or are we hitting the 4 hour log stream kube error. We will attempt to reconnect a max of 5 times in order to cover both cases
+					// If we can't get reconnect without an EOF we will error and mark the job as failed.
+					retryGetLogStream--
+					if retryGetLogStream > 0 {
+						kw.GetWorkceptor().nc.GetLogger().Info(
+							"Detected Error: %s for pod %s/%s. Will retry %d more times.",
```
dleehr:
```suggestion
							"Detected EOF Error: %s for pod %s/%s in with container state: Running. Job may not be complete. Will retry %d more times.",
```

It will be really helpful for the log messages to explicitly report the state. Here we know we encountered an `EOF` error and we fetched the container but it is still in `Running` state.

What do you think of including the `curContainerDelay` (the time we're about to Sleep) in these log messages too? Just so we can see in the logs that delay?

pkg/workceptor/kubernetes.go
```diff
@@ -338,115 +343,174 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
+							retryGetLogStream,
+						)
+
+						time.Sleep(time.Second * time.Duration(curContainerDelay))
+						prevContainerDelay, curContainerDelay = curContainerDelay, prevContainerDelay+curContainerDelay
+
+						continue mainLoop
+					}
+					// Retrying hasn't worked we will error and mark the job as failed
+					kw.GetWorkceptor().nc.GetLogger().Debug("Container in %s pod is running but unable to attach to the log stream", WorkerContainerName)
```
dleehr: Since we're returning with failure here, this should be `Error()` and not `Debug()`, right?

Also, as above it would help a lot for tracing problems if we can make the log messages consistent from the retry attempts and giving up.

lranjbar (approved):

pkg/workceptor/kubernetes.go
dleehr: Also at this point, we did successfully **attach** to the log stream but we keep getting EOFs when we try to read it. Let's be clear on that.

pkg/workceptor/kubernetes.go
```diff
@@ -338,115 +343,174 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
+					*stdoutErr = fmt.Errorf("detected Error: %s for pod %s/%s. Pod is running but unable to attach to the log stream", err,
```
dleehr: Same as above, we did successfully **attach** to the log stream but we keep getting EOFs when we try to read it. Let's be clear on that.

pkg/workceptor/kubernetes.go
```diff
@@ -338,115 +343,174 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
+						podNamespace,
+						podName,
+					)
+
+					return
+				case containerState.Terminated != nil:
+					// We got EOF and the pod terminated in a bad state, we will return the error and mark the job as failed.
+					if containerState.Terminated.ExitCode != 0 {
+						kw.GetWorkceptor().nc.GetLogger().Debug("Container in %s pod has terminated, with terminated exit code: %v", WorkerContainerName, containerState.Terminated.ExitCode)
```
dleehr: Do we want `Debug()` here?
```suggestion
						kw.GetWorkceptor().nc.GetLogger().Info("Container in %s pod has terminated, with nonzero exit code: %v", WorkerContainerName, containerState.Terminated.ExitCode)
```

Minor suggestion - In this case I think receptor did its job so I would log it as Info (not Error), but I think putting a word like `nonzero` will make it easier to search for these.

pkg/workceptor/kubernetes.go
```diff
@@ -338,115 +343,174 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
+						*stdoutErr = fmt.Errorf("detected Error: %s for pod %s/%s. Pod has terminated, with terminated exit code: %v", err,
```
dleehr, line 471: In this case (EOF and container is terminated and exit code is nonzero), don't we still need to process the last line? 

🤔 I'm concerned this change is a regression. We don't handle exit codes today (on `devel`), so it's hard to be sure. But, I'm thinking `ansible-runner` is still expected to produce that last line of JSON even if the playbook fails (which is a common cause of non-zero, right?). Is that your understanding or am I imagining things?

pkg/workceptor/kubernetes.go
```diff
@@ -338,115 +343,174 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
-						_, err = stdout.Write([]byte(msg + "\n"))
-						if err != nil {
-							*stdoutErr = fmt.Errorf("writing to stdout: %s", err)
-							kw.GetWorkceptor().nc.GetLogger().Error("Error writing to stdout: %s", err)
-
-							return
-						}
-					}
-				}
+				kw.GetWorkceptor().nc.GetLogger().Error("received EOF on log stream for pod %sand container state is not valid %s, failing and markingthe job as failed", podName, containerState)
```
dleehr:
```suggestion
				kw.GetWorkceptor().nc.GetLogger().Error("received EOF on log stream for pod %s and container state is not valid %s, failing and marking the job as failed", podName, containerState)
```

pkg/workceptor/kubernetes.go
```diff
@@ -338,115 +343,174 @@ func (kw *KubeUnit) KubeLoggingWithReconnect(streamWait *sync.WaitGroup, stdout
...
+				*stdoutErr = fmt.Errorf("received EOF on log stream for pod %sand container state is not valid %s, failing and markingthe job as failed", podName, containerState)
```
dleehr:
```suggestion
				*stdoutErr = fmt.Errorf("received EOF on log stream for pod %s and container state is not valid %s, failing and marking the job as failed", podName, containerState)
```

dleehr (changes requested):
Great work. I really like this direction and think it will help a lot that we're setting `stdoutErr` in cases where we fail to process stdout and cannot recover.

The code looks great functionally. I raised one important question around handling of `containerState.Terminated.ExitCode` when nonzero. I don't know for sure what the behavior is today (since we don't check exit codes), but I am concerned this is a regression. Basically - does this mean we start throwing away the JSON object for failed jobs? Do we have any existing tests for that?

Other than that, all of my feedback is around log polishing (to help our future selves) and not on functionality.

pkg/workceptor/kubernetes.go
matoval: I agree, I will make this log `Info`

pkg/workceptor/kubernetes.go
matoval: `Error` makes sense over Debug, I thought `RunWorkUsingLogger` returned an `Error` log but looks like it just adds the error message to the status.

pkg/workceptor/kubernetes.go
matoval: The issue here is we could have a partial line and if we process the last line it would cause a JSON parse error.

If the container failed while it was streaming a long line we wouldn't have the whole JSON payload.

pkg/workceptor/kubernetes.go
matoval: I'm not sure about adding the delay time in the log. We already have the count so we know where the sleep happen. 🤔

pkg/workceptor/kubernetes.go
dleehr: > If the container failed while it was streaming a long line we wouldn't have the whole JSON payload.

Not sure I follow - we're here because we got an `EOF` and we confirmed the container state is terminated. Sure, it's possible the stream is truncated and we don't have a full JSON payload. But isn't the more common case that it was just a failed playbook? How are we handling that?

I guess my confusion that on `devel` where we don't check the exit code, this logic would still apply whether or not the exitcode was zero:

```go
if err != io.EOF {
	*stdoutErr = err
} else if line != "" && err == io.EOF {
	msg, _, _ := kw.ProcessLogLine(line, sinceTime, successfulWrite)
	if msg != "" {
		_, err = stdout.Write([]byte(msg + "\n"))
		if err != nil {
			*stdoutErr = fmt.Errorf("writing to stdout: %s", err)
			kw.GetWorkceptor().nc.GetLogger().Error("Error writing to stdout: %s", err)

			return
		}
	}
}
```

In this PR, we're only doing that for exit code 0:

```go
// EOF and exit code is 0, this is a good state, we need to check if last line has data
if line != "" {
  msg, _, _ := kw.ProcessLogLine(line, sinceTime, successfulWrite)
  if msg != "" {
    _, err = stdout.Write([]byte(msg + "\n"))
    if err != nil {
      *stdoutErr = fmt.Errorf("error writing last line to stdout: %s", err)
      kw.GetWorkceptor().nc.GetLogger().Error("Error writing last line to stdout: %s", err)

      return
    }
  }
}
```

Why don't we need to do that for nonzero? How do we make sure that last non-empty line is written for Nonzero? Do we need to?

pkg/workceptor/kubernetes.go
dleehr: 👍

pkg/workceptor/kubernetes.go
matoval: If the playbook fails would the container fail as well? I would have to test that case but my assumption is that if a playbook fails the container would return a 0 exit code because the playbook finished running and the ansible-runner process exited.

My worry here is causing JSON parse errors if we write the last line that is a partial line caused by the container terminating while reading a long line.

Let me test a playbook that fails and see what happens to the container.

dleehr: I think I follow a bit better now.  I honestly don't know if we expect ansible-runner to always exit 0 even if a playbook fails, so I'll defer to your expertise there. 

I do see that we write each line as we handle it, but if we get an EOF before \n, we'd have a partial line AND an EOF. 

So in this case a nonzero exit code with a partial line, we're dropping the partial line, and maybe that's OK.

dleehr (approved):

pkg/workceptor/kubernetes.go
dleehr: Aaron reached out to ansible-runner team for clarity and recommended we go ahead. I agree. Even in my questions about this case, it's an edge case that's not a blocker.

dleehr: Update: we did reach out to ansible-runner team and it sounds like this might introduce a bug. @matoval  is investigating, but the understanding is that even if the container exits nonzero, since it's terminated and we have reached EOF, we should still interpret that as receptor finishing its work. so in that case we would not want to set `stdoutErr`, nor would we want to (potentially) discard the partial line in the buffer that was fetched with the EOF.

PR #1411, opened 2025-08-13
Don't fail job when non zero terminated exit code

matoval:
Don't fail job when non zero terminated exit code, update tests to ensure last line is read if not empty

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -467,14 +467,7 @@ mainLoop:
 				case containerState.Terminated != nil:
 					// We got EOF and the pod terminated in a bad state, we will return the error and mark the job as failed.
 					if containerState.Terminated.ExitCode != 0 {
-						kw.GetWorkceptor().nc.GetLogger().Info("Container in %s pod has terminated, with nonzero exit code: %v", WorkerContainerName, containerState.Terminated.ExitCode)
-						*stdoutErr = fmt.Errorf("detected Error: %s for pod %s/%s. Pod has terminated, with terminated exit code: %v", err,
-							podNamespace,
-							podName,
-							containerState.Terminated.ExitCode,
-						)
-
-						return
+						kw.GetWorkceptor().nc.GetLogger().Info("Container in %s pod has terminated, with nonzero exit code: %v, treminated reason: %v and terminated message: %v", WorkerContainerName, containerState.Terminated.ExitCode, containerState.Terminated.Reason, containerState.Terminated.Message)
 					}
 
 					// EOF and exit code is 0, this is a good state, we need to check if last line has data
@@ -490,7 +483,7 @@ mainLoop:
 							}
 						}
 					}
-					// Got EOF, terminaled code 0 and ensured we captured last line then return
+					// Got EOF, terminated and ensured we captured last line then return
 					return
 				default:
 					// We dont expect to ever get here, However, beinging in an unknown state will not have a negative effect so we will log and ignore.
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -465,12 +465,12 @@ mainLoop:
 
 					return
 				case containerState.Terminated != nil:
-					// We got EOF and the pod terminated in a bad state, we will return the error and mark the job as failed.
+					// We got EOF and the pod terminated, we will log the terminated information
 					if containerState.Terminated.ExitCode != 0 {
 						kw.GetWorkceptor().nc.GetLogger().Info("Container in %s pod has terminated, with nonzero exit code: %v, terminated reason: %v and terminated message: %v", WorkerContainerName, containerState.Terminated.ExitCode, containerState.Terminated.Reason, containerState.Terminated.Message)
 					}
 
-					// EOF and exit code is 0, this is a good state, we need to check if last line has data
+					// We need to check if last line has data
 					if line != "" {
 						msg, _, _ := kw.ProcessLogLine(line, sinceTime, successfulWrite)
 						if msg != "" {
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -467,7 +467,7 @@ mainLoop:
 				case containerState.Terminated != nil:
 					// We got EOF and the pod terminated in a bad state, we will return the error and mark the job as failed.
 					if containerState.Terminated.ExitCode != 0 {
-						kw.GetWorkceptor().nc.GetLogger().Info("Container in %s pod has terminated, with nonzero exit code: %v, treminated reason: %v and terminated message: %v", WorkerContainerName, containerState.Terminated.ExitCode, containerState.Terminated.Reason, containerState.Terminated.Message)
+						kw.GetWorkceptor().nc.GetLogger().Info("Container in %s pod has terminated, with nonzero exit code: %v, terminated reason: %v and terminated message: %v", WorkerContainerName, containerState.Terminated.ExitCode, containerState.Terminated.Reason, containerState.Terminated.Message)
 					}
 
 					// EOF and exit code is 0, this is a good state, we need to check if last line has data
--- a/pkg/workceptor/kubernetes_test.go
+++ b/pkg/workceptor/kubernetes_test.go
@@ -524,7 +524,7 @@ func (e *eofReadCloser) Read(p []byte) (n int, err error) {
 		n = copy(p, []byte(e.content))
 		e.hasRead = true
 
-		return n, nil
+		return n, io.EOF
 	}
 
 	return 0, io.EOF
@@ -4046,9 +4046,9 @@ func TestKubeUnit_RunWorkUsingLogger_ExitCode1SetsFinished(t *testing.T) {
 					Namespace: testNamespace,
 				},
 				Status: corev1.PodStatus{
-					Phase: corev1.PodRunning,
+					Phase: corev1.PodFailed,
 					Conditions: []corev1.PodCondition{
-						{Type: corev1.PodReady, Status: corev1.ConditionTrue},
+						{Type: corev1.PodReady, Status: corev1.ConditionFalse},
 					},
 					ContainerStatuses: []corev1.ContainerStatus{
 						{
--- a/pkg/workceptor/kubernetes_test.go
+++ b/pkg/workceptor/kubernetes_test.go
@@ -12,6 +12,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
 	"reflect"
 	"strings"
 	"sync"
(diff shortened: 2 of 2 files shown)


pkg/workceptor/kubernetes.go
```diff
@@ -467,14 +467,7 @@ mainLoop:
...
 					if containerState.Terminated.ExitCode != 0 {
-						kw.GetWorkceptor().nc.GetLogger().Info("Container in %s pod has terminated, with nonzero exit code: %v", WorkerContainerName, containerState.Terminated.ExitCode)
-						*stdoutErr = fmt.Errorf("detected Error: %s for pod %s/%s. Pod has terminated, with terminated exit code: %v", err,
-							podNamespace,
-							podName,
-							containerState.Terminated.ExitCode,
-						)
-
-						return
+						kw.GetWorkceptor().nc.GetLogger().Info("Container in %s pod has terminated, with nonzero exit code: %v, treminated reason: %v and terminated message: %v", WorkerContainerName, containerState.Terminated.ExitCode, containerState.Terminated.Reason, containerState.Terminated.Message)
```
dleehr:
```suggestion
						kw.GetWorkceptor().nc.GetLogger().Info("Container in %s pod has terminated, with nonzero exit code: %v, terminated reason: %v and terminated message: %v", WorkerContainerName, containerState.Terminated.ExitCode, containerState.Terminated.Reason, containerState.Terminated.Message)
```

pkg/workceptor/kubernetes.go
```diff
@@ -467,14 +467,7 @@ mainLoop:
 				case containerState.Terminated != nil:
 					// We got EOF and the pod terminated in a bad state, we will return the error and mark the job as failed.
```
dleehr: Iz this comment out of date?

pkg/workceptor/kubernetes.go
davemulford: "treminated" should be "terminated"

pkg/workceptor/kubernetes.go
```diff
@@ -467,14 +467,7 @@ mainLoop:
...
-							podNamespace,
-							podName,
-							containerState.Terminated.ExitCode,
-						)
-
-						return
+						kw.GetWorkceptor().nc.GetLogger().Info("Container in %s pod has terminated, with nonzero exit code: %v, treminated reason: %v and terminated message: %v", WorkerContainerName, containerState.Terminated.ExitCode, containerState.Terminated.Reason, containerState.Terminated.Message)
 					}
 
 					// EOF and exit code is 0, this is a good state, we need to check if last line has data
```
dleehr: I think this comment is out of date - we have EOF and terminated, don't care what the exit code was now.

pkg/workceptor/kubernetes.go
```diff
@@ -490,7 +483,7 @@ mainLoop:
 							}
 						}
 					}
-					// Got EOF, terminaled code 0 and ensured we captured last line then return
+					// Got EOF, terminated and ensured we captured last line then return
```
dleehr, line 486: 👍

davemulford (changes requested):
Great tests and I really like the comment explaining what's happening there. Minor misspelling. Looks good otherwise!

pkg/workceptor/kubernetes_test.go
```diff
@@ -3950,118 +3951,159 @@ func TestKubeWorkerCfg_Run(t *testing.T) {
...
-				ContainerStatuses: []corev1.ContainerStatus{
-					{
-						Name: workceptor.WorkerContainerName,
-						State: corev1.ContainerState{
-							Terminated: &corev1.ContainerStateTerminated{
-								ExitCode: 1,
-								Reason:   "Error",
-								Message:  "Process completed with errors",
+				Status: corev1.PodStatus{
+					Phase: corev1.PodRunning,
```
dleehr: Nitpick: the actual phase would probably be `.Failed`, but since we're not checking pod phases I don't think it matters, right?

pkg/workceptor/kubernetes_test.go
```diff
@@ -3950,118 +3951,159 @@ func TestKubeWorkerCfg_Run(t *testing.T) {
...
+					Conditions: []corev1.PodCondition{
+						{Type: corev1.PodReady, Status: corev1.ConditionTrue},
```
dleehr: similar note here - Once everything is done I think the ready condition is False. But probably not worth the cycles to change this 😁

davemulford (approved):
LGTM!

pkg/workceptor/kubernetes_test.go
```diff
@@ -3950,118 +3951,159 @@ func TestKubeWorkerCfg_Run(t *testing.T) {
...
+			expectContentWrite: false,
+			expectedContent:    "",
+		},
+		{
+			name:               "Container with exit code 1 and log data should write content to stdout",
+			podName:            "failed-pod-with-data-123",
+			logContent:         "2024-12-09T00:31:18.823849250Z Process failed with exit code 1",
+			description:        "Exit code 1 with log data - should write last line content to stdout",
+			expectContentWrite: true,
+			expectedContent:    "Process failed with exit code 1",
```
dleehr, line 3990: I like this - the `logContent` has a timestamp and a message. The test logic wraps that in an `eofReadCloser` which is designed to return the message without a `\n`. We confirm that partial message shows up as expectedContent.

However, I'm looking at eofReadCloser and I'm not sure it matches what https://pkg.go.dev/bufio#Reader.ReadString does

> If ReadString encounters an error before finding a delimiter, it returns the data read before the error and the error itself (often io.EOF)

eofReadCloser:

https://github.com/ansible/receptor/blob/e72f23accaedcc6b550300406cc2250e668c3915/pkg/workceptor/kubernetes_test.go#L520-L530

So to match the documented behavior of `ReadString` (and trigger our last-line logic), shouldn't we `return n, io.EOF`?

dleehr (commented):
Great work, love the new tests. Had a couple of suggestions about comments that don't match the logic anymore and a question around `eofReadCloser` behavior.

I'm currently testing my suggested change to `eofReadCloser`, will sync on that.

pkg/workceptor/kubernetes_test.go
matoval: Doesn't change the outcome of the test but I'll change it to PodFailed since that is the correct phase in this case

pkg/workceptor/kubernetes_test.go
dleehr: Good news. I made that change locally to `eofReadcloser` and ran the new test. It still succeeds:

```
➜  receptor git:(dont-fail-job-non-zero-exit) ✗ git diff | cat
diff --git a/pkg/workceptor/kubernetes_test.go b/pkg/workceptor/kubernetes_test.go
index ad77143..ebb8ed6 100644
--- a/pkg/workceptor/kubernetes_test.go
+++ b/pkg/workceptor/kubernetes_test.go
@@ -524,7 +524,7 @@ func (e *eofReadCloser) Read(p []byte) (n int, err error) {
 		n = copy(p, []byte(e.content))
 		e.hasRead = true
 
-		return n, nil
+		return n, io.EOF
 	}

➜  receptor git:(dont-fail-job-non-zero-exit) ✗ go test -run ^TestKubeUnit_RunWorkUsingLogger_ExitCode1SetsFinished$ github.com/ansible/receptor/pkg/workceptor
ok  	github.com/ansible/receptor/pkg/workceptor	0.514s
```

I checked the others where we set exitcode - `TestKubeLoggingWithReconnectSimple` and `TestKubeLoggingWithReconnectDuplicateDetection` also pass with my change.

But the bad news is that `TestKubeLoggingWithReconnect` (the big one) fails. I'm not sure if my assumption is wrong,  if the test has a bug, or there's a bug in the code.

dleehr: Just for reference:

```
Running tool: /opt/homebrew/bin/go test -timeout 30s -run ^TestKubeLoggingWithReconnect$ github.com/ansible/receptor/pkg/workceptor

INFO 2025/08/13 16:28:00 Detected EOF Error: EOF for pod Test_Namespace/Test_Name in with container state: Running. Job may not be complete. Will retry 4 more times.
WARNING 2025/08/13 16:28:01 Error opening log stream for pod Test_Namespace/Test_Name. Will retry 5 more times. Error: context canceled
WARNING 2025/08/13 16:28:02 Error opening log stream for pod Test_Namespace/Test_Name. Will retry 4 more times. Error: context canceled
WARNING 2025/08/13 16:28:03 Error opening log stream for pod Test_Namespace/Test_Name. Will retry 3 more times. Error: context canceled
WARNING 2025/08/13 16:28:04 Error opening log stream for pod Test_Namespace/Test_Name. Will retry 2 more times. Error: context canceled
WARNING 2025/08/13 16:28:05 Error opening log stream for pod Test_Namespace/Test_Name. Will retry 1 more times. Error: context canceled
ERROR 2025/08/13 16:28:06 Error opening log stream for pod Test_Namespace/Test_Name. Error: context canceled
panic: test timed out after 30s
	running tests:
		TestKubeLoggingWithReconnect (30s)
		TestKubeLoggingWithReconnect/retry_counter_resets_after_write_succeeds (7s)
```

dleehr: Just to summarize my position here.

I think we have a bug in our eofReadCloser (assuming it should align with https://pkg.go.dev/bufio#Reader.ReadString). I made a local change with what I think a fix is, and I looked at 3 tests that set an exitcode and test this behavior (ensuring the last line returned with an EOF is handled) are passing. So I think that's great.

There's a bigger test TestKubeLoggingWithReconnect that is failing when I make this change to eofReadCloser, but I haven't dug into why.

dleehr: Tried again and it succeeded, and also succeeded for @matoval

dleehr: 106bd6625da9a971200233dd915458628dba8318

dleehr (approved):
LGTM 🎉

lranjbar (changes requested):
With these PRs for the escalation's we have now recreated the logic of [PodContainerHealthy().](https://github.com/ansible/receptor/blob/devel/pkg/workceptor/pod.go#L16-L64) I think it is time to look at calling this function instead of leaving the logic in RunWorkWithLogger()

dleehr:
> With these PRs for the escalation's we have now recreated the logic of [PodContainerHealthy().](https://github.com/ansible/receptor/blob/devel/pkg/workceptor/pod.go#L16-L64) I think it is time to look at calling this function instead of leaving the logic in RunWorkWithLogger()

Thank you for your review @lranjbar. I agree that it is a better practice to use a specific helper like `PodContainerHealthy()` and I see that it was added in #1371 over a month ago but was not discussed for this effort.

I've been engaged in this review and #1400, so I took a look at `PodContainerHealthy` to see how it could fit here. While it's a good abstraction of our operational logic prior to #1400 (with some additional checks), it's not a drop-in replacement for the logic that has been iterated on, tested, and reviewed in #1400 / #1411. At a glance, there are differences to handling exit codes and `Ready` that don't align. These are areas the team has focused a great deal on, and they are critical to our response here.

Our priority with this PR is to get our updated logic in a build out for testing and feedback (and be ready to incorporate that feedback). I'm totally onboard with refactoring after we confirm a fix, but before is premature in my opinion. So I don't think we should refactor this to use `PodContainerHealthy()` at this time.

The good news is that once we deliver a build that addresses issues, we can confidently abstract them into `pod.go`. What do you think?

Edit: Since you're more familiar with `PodContainerHealthy()`, do you think it would provide an improvement over the logic we have here? I meant to ask that up front.

lranjbar:
> > With these PRs for the escalation's we have now recreated the logic of [PodContainerHealthy().](https://github.com/ansible/receptor/blob/devel/pkg/workceptor/pod.go#L16-L64) I think it is time to look at calling this function instead of leaving the logic in RunWorkWithLogger()
> 
> Thank you for your review @lranjbar. I agree that it is a better practice to use a specific helper like `PodContainerHealthy()` and I see that it was added in #1371 over a month ago but was not discussed for this effort.
> 
> I've been engaged in this review and #1400, so I took a look at `PodContainerHealthy` to see how it could fit here. While it's a good abstraction of our operational logic prior to #1400 (with some additional checks), it's not a drop-in replacement for the logic that has been iterated on, tested, and reviewed in #1400 / #1411. At a glance, there are differences to handling exit codes and `Ready` that don't align. These are areas the team has focused a great deal on, and they are critical to our response here.
> 
> Our priority with this PR is to get our updated logic in a build out for testing and feedback (and be ready to incorporate that feedback). I'm totally onboard with refactoring after we confirm a fix, but before is premature in my opinion. So I don't think we should refactor this to use `PodContainerHealthy()` at this time.
> 
> The good news is that once we deliver a build that addresses issues, we can confidently abstract them into `pod.go`. What do you think?
> 
> Edit: Since you're more familiar with `PodContainerHealthy()`, do you think it would provide an improvement over the logic we have here? I meant to ask that up front.

It was brought up several times before any pull request was made by two team members.

The fact that you chose to dismiss that multiple times means that you made a third team member recreate the identical logic in the hero function. Instead of using the functions that were made a month ago, for this purpose.

That discussion was before #1399, #1400 and #1411 even existed.  Now is the time.

AaronH88:
The temperature is rather high on this one, 

I am not seeing the advantage of delaying merging this. 

We will revisit how we can refactor interacting with Pods and Containers into a more generalised solution, taking into account the work already done in that space. 
Given the Sev 1 is already a week old, and we have a fully tested and reviewed PR here, I am more comfortable accepting the risk of potential code duplication rather than accepting the risk of new tests and a refactor of [PodContainerHealthy()](https://github.com/ansible/receptor/blob/devel/pkg/workceptor/pod.go#L16-L64) 

Once we have the customers back to green we will revisit 😄

PR #1412, opened 2025-08-14
Add tests to AcceptLoop in conn.go

lranjbar:
Exports AcceptLoop() so that tests can be written against it directly. This one has been particularly hard to target through other means.

Finishes off AAP-47012


AaronH88:
The code looks good, 

Any ideas why the coverage went down by nearly 3% with this? 🤔

lranjbar:
> The code looks good,
> 
> Any ideas why the coverage went down by nearly 3% with this? 🤔

No idea initially it was reporting +0.30% or something like this. 🤔 I'm rerunning the jobs

AaronH88 (approved):
lgtm

matoval (approved):

PR #1417, opened 2025-08-20
Check container is running before attaching stdin & check context canceled first when reading line

matoval:
Check container is running before attaching stdin & check context canceled first when reading line

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -374,7 +374,21 @@ mainLoop:
 		for { // check between every line read to see if we need to stop reading
 			line, err := streamReader.ReadString('\n')
 			if err != nil {
-				// First check if the error is not EOF, if error is not EOF retry 5 times if error persists set error and mark the job as failed.
+				// Check if the context was canceled and the work state isn't "Succeeded".
+				// If so, set the error and mark the job as failed.
+				if kw.GetContext().Err() == context.Canceled {
+					if kw.Status().State != WorkStateSucceeded {
+						errMsg := fmt.Sprintf("Context was canceled while reading logs for pod %s/%s. This is unrecoverable. Marking the job as failed and exiting. Error: %s",
+							podNamespace,
+							podName,
+							err.Error(),
+						)
+						*stdoutErr = fmt.Errorf("%s", errMsg)
+						kw.GetWorkceptor().nc.GetLogger().Error(errMsg)
+					}
+				}
+
+				// Check if the error is not EOF, if error is not EOF retry 5 times if error persists set error and mark the job as failed.
 				if err != io.EOF {
 					retryGetLogStream--
 					if retryGetLogStream > 0 {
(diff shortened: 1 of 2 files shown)


pkg/workceptor/kubernetes.go
```diff
@@ -374,7 +374,21 @@ mainLoop:
...
+					if kw.Status().State != WorkStateSucceeded {
+						errMsg := fmt.Sprintf("Context was canceled while reading logs for pod %s/%s. This is unrecoverable. Marking the job as failed and exiting. Error: %s",
+							podNamespace,
+							podName,
+							err.Error(),
+						)
+						*stdoutErr = fmt.Errorf("%s", errMsg)
+						kw.GetWorkceptor().nc.GetLogger().Error(errMsg)
+					}
+				}
```
dleehr: Should we `return` here if the context is canceled? It sounds like if it's "unrecoverable", we should right?

For the other case (context is canceled and state is `WorkStateSucceeded`), it seems like we should we return there too. Do we need to keep processing logs or clean anything up?

pkg/workceptor/kubernetes.go
```diff
@@ -900,6 +914,51 @@ func (kw *KubeUnit) RunWorkUsingLogger() {
...
+
+						continue
+					}
+					kw.GetWorkceptor().nc.GetLogger().Error("Container in %s pod is not running, retries exhausted", podName)
+
+					return
+				}
+
+				break podLoop
+			}
```
dleehr: Can we switch to this pattern we have in `KubeLoggingWithReconnect`:

https://github.com/ansible/receptor/blob/4acbfbed9c2f983ffd11f9049e6d19de76250f2b/pkg/workceptor/kubernetes.go#L424-L441

I think it's a lot easier to read to first locate the container status, second confirm we found it, then third decide what to do with it.

pkg/workceptor/kubernetes.go
```diff
@@ -900,6 +914,51 @@ func (kw *KubeUnit) RunWorkUsingLogger() {
...
+
+				return
+			}
+			retryCount = 5
+
+			for _, containerStatus := range podDetails.Status.ContainerStatuses {
+				if containerStatus.Name == WorkerContainerName && containerStatus.State.Running == nil {
+					retryCount--
+					if retryCount > 0 {
+						kw.GetWorkceptor().nc.GetLogger().Info("Container in %s pod is not running, continuing try to wait for container to get into running state, will retry %v more times.", podName, retryCount)
```
dleehr: It looks like we just log that it's not running. Can we log what the actual state is? Or if we refactor this to a `switch`, we can log it there. But it would be good to be able to tell from the logs if it's `Waiting` ⏳  or `Terminated` 💀

pkg/workceptor/kubernetes.go
```diff
@@ -374,7 +374,21 @@ mainLoop:
...
+				// Check if the context was canceled and the work state isn't "Succeeded".
+				// If so, set the error and mark the job as failed.
+				if kw.GetContext().Err() == context.Canceled {
+					if kw.Status().State != WorkStateSucceeded {
+						errMsg := fmt.Sprintf("Context was canceled while reading logs for pod %s/%s. This is unrecoverable. Marking the job as failed and exiting. Error: %s",
+							podNamespace,
+							podName,
+							err.Error(),
+						)
+						*stdoutErr = fmt.Errorf("%s", errMsg)
```
dleehr: I remember in #1400 that we removed a check for canceled context, and that line was one of the first things @AaronH88 called out from our initial review of the log messages. I think the analysis was that we were checking the context and marking an error, but we weren't examining or logging the underlying error, so it might have masked a real error. I can't find a comment other than https://github.com/ansible/receptor/pull/1400#discussion_r2266349052.

Since this does check the context canceled and logs the error from reading the stream, I think it's covering that case. But I would like to get @AaronH88 to weigh in.

dleehr (commented):
Nice progress 👍. Had some questions about the exit cases and some feedback around readaiblity/enhancements.

pkg/workceptor/kubernetes.go
matoval: My thinking here is yes we should return if context canceled and status not succeeded because we can't get any logs from the kube api anymore.

I think we need to verify if the context can be canceled after getting the pod from the kube API and we get context canceled, status not succeeded and we have pod/container data to log.

AaronH88: Agreed, 
Ill move the return to the outside if statement, line 390 ish, and return anyway, for all cases where the context is canceled

pkg/workceptor/kubernetes.go
AaronH88: Yeah good points,

Their seems to be an action on us to map out all cases where the context gets canceled and improve that flow, but it would be outside the scope of this current work.

I agree with both of ye, this current check seems to be good enough for now 😄

pkg/workceptor/kubernetes.go
```diff
@@ -900,6 +916,86 @@ func (kw *KubeUnit) RunWorkUsingLogger() {
 		kw.UpdateBasicStatus(WorkStateRunning, "Pod Running", stdout.Size())
 		streamWait.Done()
 	} else {
+		retryCount := 5
+		prevPodDelay, curPodDelay := 0, 1
```
AaronH88: Lets start at 1 -> 2 not 1 -> 1

Ill make this change now

pkg/workceptor/kubernetes.go
```diff
@@ -900,6 +916,86 @@ func (kw *KubeUnit) RunWorkUsingLogger() {
...
+					kw.GetWorkceptor().nc.GetLogger().Info("Error getting pod while trying to attach stdin: '%s' , continuing try to get pod up to %v more times.", kubeErr, retryCount)
+
+					time.Sleep(time.Second * time.Duration(curPodDelay))
+					prevPodDelay, curPodDelay = curPodDelay, prevPodDelay+curPodDelay
+
+					continue
+				}
+				kw.GetWorkceptor().nc.GetLogger().Error("Error getting pod %s/%s, after retries exhausted. Error: %s", podNamespace, podName, kubeErr)
+
+				return
```
AaronH88, line 944: returning from RunWorkUsingLogger() ends our interaction with the job, so for each of these returns in this function we need to make sure the job status is in the correct state. 

Ill add some status updates calls for these returns.

pkg/workceptor/kubernetes.go
```diff
@@ -374,7 +374,23 @@ mainLoop:
 		for { // check between every line read to see if we need to stop reading
 			line, err := streamReader.ReadString('\n')
 			if err != nil {
-				// First check if the error is not EOF, if error is not EOF retry 5 times if error persists set error and mark the job as failed.
+				// Check if the context was canceled and the work state isn't "Succeeded".
+				// If so, set the error and mark the job as failed.
+				if kw.GetContext().Err() == context.Canceled {
+					if kw.Status().State != WorkStateSucceeded {
```
AaronH88: The context could also be canceled for both a failed and canceled job, 
I think we add an or for both these cases also

AaronH88 (approved):
lgtm

PR #1418, opened 2025-08-25
Add runWorkUsingTCP test

matoval:
Add runWorkUsingTCP test

### DIFF
--- a/pkg/workceptor/kubernetes_test.go
+++ b/pkg/workceptor/kubernetes_test.go
@@ -4767,7 +4767,8 @@ func TestKubeUnit_RunWorkUsingTCP(t *testing.T) {
 						workerContainer := &corev1.Container{}
 						for _, container := range pod.Spec.Containers {
 							if container.Name == workceptor.WorkerContainerName {
-								workerContainer = &container
+								containerCopy := container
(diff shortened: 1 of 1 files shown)


davemulford (approved):
Very nice tests! LGTM

pkg/workceptor/kubernetes_test.go
```diff
@@ -4655,3 +4655,430 @@ func TestKubeUnit_RunWorkUsingLogger_ContainerStateSwitch(t *testing.T) {
...
+		testNamespace = "default"
+		testUnitDir   = "/tmp/test/tcp/unit/"
+	)
+
+	// Create the unit directory and stdin file for the test
+	err := os.MkdirAll(testUnitDir, 0o755)
+	if err != nil {
+		t.Fatalf("Failed to create test unit directory: %v", err)
+	}
+	defer os.RemoveAll(testUnitDir)
```
lranjbar: Instead of using defer we can use `t.Cleanup()` this will take into account unexpected panics.

lranjbar (changes requested):
Very small change to use t.Cleanup() instead of defer

lranjbar (approved):

PR #1425, opened 2025-09-04
Add env vars for the start duration and amount of retries for kube api retries

matoval:
Add env vars for the start duration and amount of retries for kube api retries


docs/source/user_guide/configuration_options.rst
```diff
@@ -942,7 +942,33 @@ Work Kubernetes
...
+
+.. list-table:: Kubernetes Environment Variables
+    :header-rows: 1
+    :widths: auto
+
+    * - Variable
+      - Description
+      - Default value
+      - Type
+    * - ``RECEPTOR_KUBE_TIMEOUT_START``
```
AaronH88, line 963: Perhaps its worth noting the valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".

docs/source/user_guide/configuration_options.rst
```diff
@@ -942,7 +942,33 @@ Work Kubernetes
...
+      - Initial timeout duration between Kubernetes API retry attempts (e.g., "1s", "500ms")
+      - 1s
+      - string (duration)
+    * - ``RECEPTOR_KUBE_RETRY_COUNT``
```
AaronH88, line 968: This is still a fib back off right? 

If yes, then I think these docs need to be re-worded to explain the non linear effect adding a bigger number here will have

pkg/workceptor/kubernetes.go
```diff
@@ -226,23 +226,42 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+
+func (kw *KubeUnit) GetKubeRetryCount() int {
+	// RECEPTOR_KUBE_RETRY_COUNT
+	// default: 5
+	kubeRetryCount := 5
+	envRetryCount := os.Getenv("RECEPTOR_KUBE_RETRY_COUNT")
+	if envRetryCount != "" {
+		var err error
+		kubeRetryCount, err = strconv.Atoi(envRetryCount)
+		if err != nil || kubeRetryCount < 1 {
```
davemulford, line 262: Should there be an upper bound on the retry count? A high number plus a fib backoff could mean very long wait times if the retry count is set to something like 100.

pkg/workceptor/kubernetes.go
```diff
@@ -346,7 +365,7 @@ mainLoop:
 				retryGetPod,
 				err,
 			)
-			time.Sleep(time.Second * time.Duration(curPodDelay))
+			time.Sleep(time.Duration(int64(kw.GetKubeTimeoutStart()) * int64(curPodDelay)))
```
davemulford: May need some overflow protection here. With `RECEPTOR_KUBE_TIMEOUT_START` set to 1 (the default value), setting `RECEPTOR_KUBE_RETRY_COUNT=25` will result in a 20 hour sleep time. Setting `RECEPTOR_KUBE_RETRY_COUNT=50` will cause a duration overflow value of `-1627909h 44m 8.709551616s`

davemulford (changes requested):
Some bounds changes requested around the `RECEPTOR_KUBE_RETRY_COUNT` environment variable.

dleehr (commented):
Discussed on collab call, can we remove the additional casts around time.Duration and int64 when we already have the timeout value as a time.Duration? This *may* help with the int64 overflow, but still should be tested.

pkg/workceptor/kubernetes.go
dleehr: Good catch on the overflow @davemulford since this is an `int64`. We also discussed a bit on collab call to see if we can simplify `time.Duration(int64(kw.GetKubeTimeoutStart()) * int64(curPodDelay)` to `kw.GetKubeTimeoutStart() * curPodDelay`.

Since `kw.GetKubeTimeoutStart()`already returns a `time.Duration` there shouldn't be any need to convert to an `int64` and then back to a `time.Duration`. 

However, even if we keep the timeout as a `time.Duration` we still need to multiply it by the delay count (expressed as an integer), so we should check that to prevent overflows there.

pkg/workceptor/kubernetes.go
```diff
@@ -226,23 +226,64 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+			kubeRetryCount = 5
+		}
+	}
+	kw.GetWorkceptor().nc.GetLogger().Debug("RECEPTOR_KUBE_RETRY_COUNT: %d", kubeRetryCount)
+
+	return kubeRetryCount
+}
+
+func (kw *KubeUnit) GetSleepDuration(multipler int) time.Duration {
+	const maxInt64 = int64(^uint64(0) >> 1)
```
dleehr: Can we use the const `math.MaxInt64` from https://pkg.go.dev/math#pkg-constants instead of calculating it ourselves?

I confirmed in a go playground this is the same value https://go.dev/play/p/IXOi06uylHk

```
math.MaxInt64 9223372036854775807
calculated maxint64 9223372036854775807
```

dleehr (changes requested):
Nice improvements in the refactor. I think having the functions helps a lot.

One nitpick on using a const provided by `math` instead of calculating something ourselves.

pkg/workceptor/kubernetes.go
```diff
@@ -226,23 +227,61 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
-			kw.GetWorkceptor().nc.GetLogger().Warning("Invalid value for RECEPTOR_OPEN_LOGSTREAM_TIMEOUT: %s. Ignoring", envTimeout)
-			openLogStreamTimeout = 1
+			kw.GetWorkceptor().nc.GetLogger().Warning("Invalid value for RECEPTOR_KUBE_TIMEOUT_START: %s. Ignoring", envTimeout)
+			kubeTimeoutStart = 1 * time.Second
 		}
+		// ignore if exceeds limit, use default
+		if kubeTimeoutStart > time.Minute*5 {
+			kw.GetWorkceptor().nc.GetLogger().Warning("RECEPTOR_KUBE_TIMEOUT_START of: %d is larger than the max timeout of 5m. Default of 1s will be used", kubeTimeoutStart)
+			kubeTimeoutStart = 1 * time.Second
+		}
```
dleehr: If the user exceeds the max value, I think it makes more sense to fall back to the max value (5m) instead of the default value (1s)

pkg/workceptor/kubernetes.go
```diff
@@ -226,23 +227,61 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+		kubeRetryCount, err = strconv.Atoi(envRetryCount)
+		if err != nil || kubeRetryCount < 1 {
+			// ignore error, use default
+			kw.GetWorkceptor().nc.GetLogger().Warning("Invalid value for RECEPTOR_KUBE_RETRY_COUNT: %s. Default of 5 will be used", envRetryCount)
+			kubeRetryCount = 5
+		}
+		// ignore if exceeds limit, use default
+		if kubeRetryCount > 100 {
+			kw.GetWorkceptor().nc.GetLogger().Warning("RECEPTOR_KUBE_RETRY_COUNT of: %d is larger than the max retry count of 100. Default of 5 will be used", kubeRetryCount)
+			kubeRetryCount = 5
```
dleehr: Same feedback here - if we have a maximum and it's exceeded, let's fall back to that maximum instead of the default.

docs/source/user_guide/configuration_options.rst
```diff
@@ -942,7 +942,43 @@ Work Kubernetes
...
+      - Number of retry attempts for Kubernetes API operations. Uses exponential backoff with Fibonacci-like sequence.
+      - 5
+      - 1-100
+      - int
+
+**Important Notes:**
+
+- **Exponential Backoff**: Retry delays increase exponentially using a Fibonacci-like sequence. For example, with ``RECEPTOR_KUBE_TIMEOUT_START=1s``, retry delays will be: 1s, 2s, 3s, 5s, 8s, etc.
+- **Timeout Start Limit**: ``RECEPTOR_KUBE_TIMEOUT_START`` values exceeding 1 minute will be capped at the maximum of 1 minute.
```
dleehr, line 977: In the code we are removing support for `RECEPTOR_OPEN_LOGSTREAM_TIMEOUT` and reading `RECEPTOR_KUBE_TIMEOUT_START`. Do we need to document that in this guide? Do we have users relying on `RECEPTOR_OPEN_LOGSTREAM_TIMEOUT` that will be confused?

docs/source/user_guide/configuration_options.rst
```diff
@@ -942,7 +942,43 @@ Work Kubernetes
...
+      - string (duration)
+    * - ``RECEPTOR_KUBE_RETRY_COUNT``
+      - Number of retry attempts for Kubernetes API operations. Uses exponential backoff with Fibonacci-like sequence.
+      - 5
+      - 1-100
+      - int
+
+**Important Notes:**
+
+- **Exponential Backoff**: Retry delays increase exponentially using a Fibonacci-like sequence. For example, with ``RECEPTOR_KUBE_TIMEOUT_START=1s``, retry delays will be: 1s, 2s, 3s, 5s, 8s, etc.
```
dleehr: Genuine question (perhaps a nitpick): is it accurate to describe our backoff as `Exponential` / `exponentially`? Fibonacci is different, right?

dleehr (approved):
Code and tests LGTM. Couple minor questions about the user experience, but nothing to hold up.

docs/source/user_guide/configuration_options.rst
matoval: That's a great point! I will document replacing `RECEPTOR_OPEN_LOGSTREAM_TIMEOUT` with the more general `RECEPTOR_KUBE_TIMEOUT_START` which replaces it.

docs/source/user_guide/configuration_options.rst
tyraziel: Just call it "Fibonacci Increment Backoff", or "Fibonacci Backoff"?

tyraziel:
```suggestion
- **Fibonacci Backoff**: Retry delays increase by Fibonacci increments. For example, with ``RECEPTOR_KUBE_TIMEOUT_START=1s``, retry delays will be: 1s, 2s, 3s, 5s, 8s, etc.
```

pkg/workceptor/kubernetes.go
```diff
@@ -226,23 +227,68 @@ func podRunningAndReady(kw KubeUnit) func(event watch.Event) (bool, error) {
...
+		if err != nil || kubeRetryCount < 1 {
+			// ignore error, use default
+			kw.GetWorkceptor().nc.GetLogger().Warning("Invalid value for RECEPTOR_KUBE_RETRY_COUNT: %s. Default of 5 will be used", envRetryCount)
+			kubeRetryCount = 5
+		}
+		// ignore if exceeds limit, use default
+		if kubeRetryCount > 100 {
+			kw.GetWorkceptor().nc.GetLogger().Warning("RECEPTOR_KUBE_RETRY_COUNT of: %d is larger than the max retry count of 100. Default of 5 will be used", kubeRetryCount)
+			kubeRetryCount = 100
+		}
```
tyraziel:
```suggestion
		// ignore if exceeds limit, use default
		if kubeRetryCount > 100 {
			kw.GetWorkceptor().nc.GetLogger().Warning("RECEPTOR_KUBE_RETRY_COUNT of: %d is larger than the max retry count of 100. Retry count of 100 will be used", kubeRetryCount)
			kubeRetryCount = 100
		}
```

Default of 5 doesn't seem to apply here.

tyraziel (changes requested):
I can't find the code, but I think at some point we should cap the Fibonacci retry wait time if we aren't.

matoval:
> I can't find the code, but I think at some point we should cap the Fibonacci retry wait time if we aren't.

We added a cap of 1m start time and 5m max sleep time.

davemulford (approved):
LGTM! Approving with the assumption that we will resolve Dan's latest comment about documenting the removal of an environment variable.

tyraziel (approved):

PR #1435, opened 2025-09-29
Correct log messages to identify the pod name instead of container name

dleehr:
Some log messages were including the name of the container (which is a constant `worker`) rather than the name of the pod. This change updates those log statements to print the name of the pod since it is useful when reviewing logs.

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -521,7 +521,11 @@ mainLoop:
 						continue mainLoop
 					}
 					// Retrying hasn't worked we will error and mark the job as failed
-					kw.GetWorkceptor().nc.GetLogger().Error("Container in %s pod is running but is continuing to stream EOF after retries exhausted", podName)
+					kw.GetWorkceptor().nc.GetLogger().Error("%s/%s: %s is running but is continuing to stream EOF after retries exhausted",
+						podNamespace,
+						podName,
+						WorkerContainerName,
+					)
 					*stdoutErr = fmt.Errorf("detected Error: %s for pod %s/%s. Pod is running but is continuing to stream EOF after retries exhausted", err,
 						podNamespace,
 						podName,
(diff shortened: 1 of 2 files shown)


pkg/workceptor/kubernetes.go
```diff
@@ -521,7 +521,7 @@ mainLoop:
 						continue mainLoop
 					}
 					// Retrying hasn't worked we will error and mark the job as failed
-					kw.GetWorkceptor().nc.GetLogger().Error("Container in %s pod is running but is continuing to stream EOF after retries exhausted", WorkerContainerName)
+					kw.GetWorkceptor().nc.GetLogger().Error("Container in %s pod is running but is continuing to stream EOF after retries exhausted", podName)
```
PabloHiro: Can we log the whole namespace/pod and container info? 

Something like:

```suggestion
					kw.GetWorkceptor().nc.GetLogger().Error("%s/%s: %s is running but is continuing to stream EOF after retries exhausted", podNamespace, podName, WorkerContainerName)
```

We should homogenize and expand the information logged in this file as long as we don't need to pass additional context or change function interfaces.

PabloHiro (changes requested):

pkg/workceptor/kubernetes.go
fincamd: +1 to Pablo, I think we should be printing as much info as we can, not just picking one of them. We have all the info, let's put it to use

dleehr: > Can we log the whole namespace/pod and container info?
> 
> Something like:
> 
> We should homogenize and expand the information logged in this file as long as we don't need to pass additional context or change function interfaces.

I don't disagree. Consistency and additional context make debugging a lot easier. I'll have a look to see what can be done without exploding the scope.

On that point, the most important piece of information to log is the pod name, which includes the controller Job ID and obviously connects this to a unique Pod in the cluster. The pod name wasn't included in the `running but is continuing to stream EOF after retries exhausted` message, so we couldn't connect these `retries exhausted` messages to failed AAP jobs or their pods.

The other info is less critical - container name is a constant specified by receptor (always `"worker"`) and the namespace is generally already known (and not changing across pods). There's no problem with adding this informatino, but the purpose of this PR was to address a specific gap.

dleehr: @PabloHiro to clarify - are you requesting for additional information logged in this line or are you requesting changes across all of `kubernetes.go` to standardize the log formatting?

PabloHiro: > see what can be done without exploding the scope

Let's keep scope small and simply update the modified lines to align with how we want this too look in the future. Other parts of the code already have the `%s/%s, podNamespace, podName` structure, I'd say we try to at least align on that.

PabloHiro (approved):

matoval (approved):

pkg/workceptor/kubernetes.go
arrestle: We have a [AAP-41768](https://issues.redhat.com/browse/AAP-41768) ticket that would  add pod and container identifiers automatically to each log message that's sitting in the backlog until next release, so I'll add a comment there to make sure this is covered going forward.

arrestle:
Note we should update Kubernetes.go with [structured logs](https://github.com/ansible/receptor/pull/1300)

PR #1437, opened 2025-10-03
Fix: Prevent race condition in job status updates

matoval:
Fix: Prevent race condition in job status updates

  Problem:
  A race condition in RunWorkUsingLogger allowed a job's failed state to be overwritten with "Finished" status, resulting in jobs that appeared both errored and finished simultaneously.

  Root Cause:
  The final status transition used a non-atomic check-then-update pattern:
```go
  // CHECK (acquires and releases lock)
  if kw.Status().State == WorkStateRunning {
      // UPDATE (acquires and releases lock separately)
      kw.UpdateBasicStatus(WorkStateSucceeded, "Finished", stdout.Size())
  }
```

  This Time-Of-Check to Time-Of-Use (TOCTOU) race allowed:
  1. A concurrent goroutine to set status to WorkStateFailed
  2. The main thread to read stale Status().State (still sees Running)
  3. The main thread to overwrite with WorkStateSucceeded/"Finished"

  Fix:
  Changed to use UpdateFullStatus with a callback for atomic check-and-update within a single lock acquisition:

```go
  kw.UpdateFullStatus(func(status *StatusFileData) {
      // Atomically check and update within single lock
      if status.State == WorkStateRunning {
          status.State = WorkStateSucceeded
          status.Detail = "Finished"
          status.StdoutSize = stdout.Size()
      }
  })
```

### DIFF
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -1155,8 +1155,15 @@ func (kw *KubeUnit) RunWorkUsingLogger() {
 	}
 
 	// only transition from WorkStateRunning to WorkStateSucceeded if WorkStateFailed is set we do not override
-	if kw.GetContext().Err() != context.Canceled && kw.Status().State == WorkStateRunning {
-		kw.UpdateBasicStatus(WorkStateSucceeded, "Finished", stdout.Size())
+	if kw.GetContext().Err() != context.Canceled {
+		kw.UpdateFullStatus(func(status *StatusFileData) {
+			// Atomically check and update within single lock to prevent race condition
+			if status.State == WorkStateRunning {
+				status.State = WorkStateSucceeded
+				status.Detail = "Finished"
+				status.StdoutSize = stdout.Size()
+			}
+		})
 	}
 }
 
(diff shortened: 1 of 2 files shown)


dleehr (approved):
Change looks good and should close a gap in concurrent behavior that might cause a problem 👍 

Looking through the tests though, you found a good way to test the condition in a mock, but it looks like the test is only testing the mock. Am I reading that right? It looks like this anonymous callback function added is copied-and-pasted into the test. Is there something we can do to make sure the test runs the code that was changed?

Playing devil's advocate - If I changed the new logic in `RunWorkUsingLogger()` (let's say I change `status.State = WorkStatePending`), would this test fail? It should - so if it does, nevermind 😁. 

If not, maybe as a start we could move this callback to be a named function

```golang
func(status *StatusFileData) {
			// Atomically check and update within single lock to prevent race condition
			if status.State == WorkStateRunning {
				status.State = WorkStateSucceeded
				status.Detail = "Finished"
				status.StdoutSize = stdout.Size()
			}

```

That way we could at least cover it in this test. 

I assume the simple logic (Running->Succeeded without testing for race conditions) is already covered in tests for `RunWorkUsingLogger()`. It would be great to test this race condition logic in those too, but I think this is a good step.

matoval:
> Add test reproducing error state and finished status race condition

Good catch! I merged two tests into one and messed it up 😅 I'm fix it now

PabloHiro (approved):
Incredible catch

dleehr (approved):
LGTM 🚀

PR #1439, opened 2025-10-06
[AAP-49041] Upgrade Quic library to 0.54.1*

arrestle:
# AAP-49041: Upgrade QUIC Library to v0.54.1

## Summary

Upgrades Receptor's QUIC library from v0.50.1 to v0.54.1 to address breaking changes and enable Epic AAP-51480 certificate work. *DO NOT PANIC REGARDING LARGE NUMBER OF CHANGES* 

Note follow on [AAP-49041:Fix context leak in AcceptLoop when RemoteAddr is not Receptor Addr#1442](https://github.com/ansible/receptor/pull/1442) fixes linting error.

Review only: 

- go.mod - Version bump
- interfaces.go - Add explicit QUIC interface definitions for mocking
- conn.go - Export adapter types (`QuicListenerAdapter`, `QuicConnAdapter`)
- adapter_test.go - New comprehensive adapter tests
- conn_test.go - Updated for new interfaces
- tests/functional/mesh/conn_test.go - Updated for new interfaces

What else changed:
- Fixed pre-existing goroutine test bugs (t.Fatalf → t.Errorf)

All meaningful code is tested

## Why v0.54.1and not v0.55.0?

QUIC v0.55.1requires Go 1.24+. Current environment uses Go 1.23.9, so v0.54.1 is the appropriate target.

## Impact

Unblocks AAP-51479 (certificate deduplication) which depends on stable QUIC interfaces.

## References

- [QUIC v0.54.0 Release](https://github.com/quic-go/quic-go/releases/tag/v0.54.0)
- [Epic AAP-51480](https://issues.redhat.com/browse/AAP-51480) Receptor CA Bundle Optimizations

PR #1442, opened 2025-10-07
AAP-49041: Fix context leak in AcceptLoop when RemoteAddr is not Receptor Addr

arrestle:
# AAP-49041: Fix context leak in AcceptLoop when RemoteAddr is not Receptor Addr

## Summary

Fixes a context leak detected by golangci-lint when RemoteAddr type assertion fails in AcceptLoop.

## Problem

When `RemoteAddr().(Addr)` type assertion fails, `MonitorUnreachable` is not started, leaving the `ccancel` function uncalled. This causes the context and cleanup goroutine to leak.

## Solution

Add `defer ccancel()` in the else branch to ensure cleanup on all code paths.

## Changes

- See latest changes in latest commit  [Fix context leak in AcceptLoop when RemoteAddr is not Receptor Addr](https://github.com/ansible/receptor/pull/1442/commits/85870fb1695adb7b6388d13c3665484e293799a8) commit. For merge reasons this is built on top of [Previous PR](https://github.com/ansible/receptor/pull/1439).

## Testing

**Primary verification:** Static linter (golangci-lint) reports:
```
the ccancel function is not used on all paths (possible context leak)
```

**Test verification:** `TestAcceptLoopContextCancelCleanup` uses `TestOnContextCancelDeferCalled` atomic bool to prove the defer executes.
## Caveat: Test Hook in Production Code

This PR includes `TestOnContextCancelDeferCalled` (atomic.Bool) in production code for test verification. 

**Rationale:**
- Context leaks are difficult to detect in unit tests
- Teammate requested proof that ccancel() wasn't initially called (removing fix fails test.)
- Follows existing pattern (KeepAliveForQuicConnections)
- Thread-safe using sync/atomic
- Only used in tests (nil/false in production)

**Alternative:** Remove the test hook and rely solely on the linter. Team to decide.


pkg/netceptor/conn.go
```diff
@@ -312,6 +317,12 @@ func (li *Listener) AcceptLoop(ctx context.Context) {
 			rAddr, ok := conn.RemoteAddr().(Addr)
 			if ok {
 				go MonitorUnreachable(li.pc, doneChan, rAddr, ccancel)
+			} else {
+				li.s.Logger.Debug("Remote address is not a Receptor address, skipping unreachable monitoring")
+				defer func() {
+					TestOnContextCancelDeferCalled.Store(true)
+					ccancel()
+				}()
```
PabloHiro: According to this:
https://github.com/golang/go/blob/27937289dc9fccf1f5513475145799087f39b964/src/context/context.go#L230

It is safe to call the cancel function simultaneously or multiple times. I believe we don't need the lock.

We can simply do:
```suggestion
			} else {
				li.s.Logger.Debug("Remote address is not a Receptor address, skipping unreachable monitoring")
			}
```

And some lines above:
```
cctx, ccancel := context.WithCancel(li.s.context)
defer ccancel()
```

What do you think? I find it simpler and easier to understand.

PabloHiro (changes requested):

pkg/netceptor/conn_test.go
```diff
@@ -82,6 +82,82 @@ func TestCancelRead(t *testing.T) {
 	conn.CancelRead()
 }
 
+// TestAcceptLoopNonReceptorAddr tests that AcceptLoop handles non-Receptor RemoteAddr correctly
+// and properly cancels the connection context via the lifecycle goroutine.
+func TestAcceptLoopNonReceptorAddr(t *testing.T) {
```
matoval, line 87: This test still passes if we remove `defer connCancel()`

matoval (changes requested):
TestAcceptLoopNonReceptorAddr isn't testing if connCancel is actually being called.

PabloHiro (approved):
Not sure how easy it is to proof this via unit tests and this has been opened for a while. So LGTM.

Expanding a bit:
* If this is something that can be caught statically by a linter, I see limited value in trying to proof that it the leak is gone.
* Worst case scenario we have reduced the circumstances in which a leak can occur. Best case we have reduce all circumstances. I would merge as is.

pkg/netceptor/conn.go
```diff
@@ -299,25 +299,28 @@ func (li *Listener) AcceptLoop(ctx context.Context) {
...
 				doneOnce: &sync.Once{},
-				ctx:      cctx,
+				ctx:      connCtx,
 			}
 			rAddr, ok := conn.RemoteAddr().(Addr)
 			if ok {
-				go MonitorUnreachable(li.pc, doneChan, rAddr, ccancel)
+				go MonitorUnreachable(li.pc, doneChan, rAddr, connCancel)
+			} else {
+				li.s.Logger.Debug("Remote address is not a Receptor address, skipping unreachable monitoring")
```
davemulford, line 320:
```suggestion
				li.s.Logger.Debug("Remote address is not a Receptor address, skipping unreachable monitoring")
				connCancel()
```

pkg/netceptor/conn.go
```diff
@@ -299,25 +299,28 @@ func (li *Listener) AcceptLoop(ctx context.Context) {
...
 			}
 			go func() {
+				defer connCancel()
```
davemulford, line 323:
```suggestion
```

pkg/netceptor/conn_test.go
```diff
@@ -82,6 +82,82 @@ func TestCancelRead(t *testing.T) {
...
+	mockQL.EXPECT().Accept(gomock.Any()).Return(mockQC, nil).Times(1)
+	mockQC.EXPECT().RemoteAddr().Return(&net.TCPAddr{IP: net.ParseIP("127.0.0.1"), Port: 8080}).Times(1)
+	mockQC.EXPECT().AcceptStream(gomock.Any()).Return(mockQS, nil).Times(1)
+	mockQS.EXPECT().Read(gomock.Any()).DoAndReturn(func(b []byte) (int, error) {
+		b[0] = 0
+
+		return 1, nil
+	}).Times(1)
+	mockQS.EXPECT().Close().Return(nil).AnyTimes()
+	mockQC.EXPECT().Context().Return(context.Background()).AnyTimes()
```
davemulford, line 109: As a suggestion to confirm that `connCancel()` is called:

```go
cancelCalled := false
mockQC.EXPECT().Context().DoAndReturn(func() (context.Context, context.CancelFunc) {
    ctx, cancel := context.WithCancel(context.Background())
    return ctx, func() {
        cancelCalled = true
        cancel()
    }
})
```

davemulford (changes requested):
From PR review, the following suggested changes were identified.

1. Move `connCancel()` into the else-block and remove the `defer`.
2. There is a suggestion on how to test that `connCancel()` is called.

Note that there are other places where context cancelation is not explicitly tested. We are adding that here as the fix is for a context leak.

pkg/netceptor/conn_test.go
dleehr: Good call. I also spotted [`AfterFunc` in the context package](https://pkg.go.dev/context#AfterFunc) which looks like another way to do this - since it's designed to call a provided function when the context is canceled.

From https://pkg.go.dev/context#AfterFunc:
> ```
>func AfterFunc(ctx Context, f func()) (stop func() bool)`
> ```
> AfterFunc arranges to call f in its own goroutine after ctx is canceled. If ctx is already canceled, AfterFunc calls f immediately in its own goroutine.
>
> Multiple calls to AfterFunc on a context operate independently; one does not replace another.
>
> Calling the returned stop function stops the association of ctx with f. It returns true if the call stopped f from being run. If stop returns false, either the context is canceled and f has been started in its own goroutine; or f was already stopped. The stop function does not wait for f to complete before returning. If the caller needs to know whether f is completed, it must coordinate with f explicitly.

So if we want to know within a test that a context was cancelled, we can call `AfterFunc` on the context and provide a function that sets `cancelCalled = true`.

fincamd (approved):

pkg/netceptor/conn_test.go
arrestle: **Re: Testing that `connCancel()` is called**

Thanks for the feedback on test verification. I wanted to clarify the context hierarchy and discuss testing approaches.

### Context hierarchy in AcceptLoop

- `ctx` (parameter) - Controls the accept loop itself
- `connCtx` (line 302) - Connection's lifecycle context, created from `li.s.context`
- `connCancel` - Cancels `connCtx`, preventing the leak

### Current implementation

```go
connCtx, connCancel := context.WithCancel(li.s.context)
// ...
go func() {
    defer connCancel()  // Cleanup when lifecycle goroutine exits
    select {
        case <-li.DoneChan:
        case <-connCtx.Done():
        case <-doneChan:
    }
}()
```

### Testing challenge

`connCtx` is created as a local variable inside the goroutine, making it difficult to observe from unit tests without refactoring. I explored using `context.AfterFunc` (Dan's suggestion), but it only detects parent context cancellation, not child contexts.

### Current test

Verifies the non-Receptor address code path executes correctly without panics/deadlocks. The linter (golangci-lint) statically verifies `connCancel()` is called on all paths.

### Options

1. Accept linter verification as primary validation
2. Refactor to expose `connCtx` for testing (seems heavyweight for this fix)
3. Different approach you'd recommend?

Thoughts?

arrestle:
> * Move `connCancel()` into the else-block and remove the `defer`.
> 
> *

Was not able to move the defer into an else block because the linter complains.

arrestle:
> 2\. There is a suggestion on how to test that `connCancel()` is called.

The suggestion was how to verify the main context was cancelled, not the sub (connection) context.

pkg/netceptor/conn_test.go
dleehr: Thanks for checking `AfterFunc`. It seemed like a good fit that would be nice and clean but if it doesn't work for this case, try @matoval's suggestion (that @davemulford captured for us at the start of this thread).

The expectation is that we have a test that covers the cancellation case. I'm glad that the linter is satisfied but the linter is not running tests.

pkg/netceptor/conn_test.go
```diff
@@ -82,6 +82,82 @@ func TestCancelRead(t *testing.T) {
...
+
+		// Close the connection, which should trigger the lifecycle goroutine's defer conn_cancel()
+		err := conn.Close()
+		if err != nil {
+			t.Errorf("Close returned error: %v", err)
+		}
+
+		// Give the lifecycle goroutine time to clean up
+		time.Sleep(10 * time.Millisecond)
+
```
davemulford, line 152: This will test whether conn's context was canceled.

```suggestion

		// If the context error is anything other than "context canceled" then
		// the connCancel() cancel function was not called.
		connErr := conn.Context().Err().Error()
		if connErr != "context canceled" {
			t.Fatalf("Connection context should be canceled. Was: %s", connErr)
		}
```

To make the context accessible from within tests, the following function should be added to `pkg/netceptor/conn.go` around line 580 after the `SetWriteDeadline` function.

```go
func (c *Conn) Context() context.Context {
	return c.ctx
}
```

davemulford (changes requested):
I've made a suggestion on how the context cancellation can be asserted. Please check it out and let me know what you think.

pkg/netceptor/conn_test.go
arrestle: This was added manually to satisfy the linter.

pkg/netceptor/conn.go
arrestle: This needs to be added outside the else block.

davemulford (approved):
LGTM, so approving. I'd still like @dleehr's and @matoval's thoughts on how we're ensuring the context is canceled is a valid way to do so.

pkg/netceptor/conn.go
```diff
@@ -299,25 +299,28 @@ func (li *Listener) AcceptLoop(ctx context.Context) {
...
 				doneChan: doneChan,
 				doneOnce: &sync.Once{},
-				ctx:      cctx,
+				ctx:      connCtx,
 			}
 			rAddr, ok := conn.RemoteAddr().(Addr)
 			if ok {
-				go MonitorUnreachable(li.pc, doneChan, rAddr, ccancel)
+				go MonitorUnreachable(li.pc, doneChan, rAddr, connCancel)
+			} else {
```
matoval, line 319:
```suggestion
			} else {
			  connCancel()
```

pkg/netceptor/conn.go
matoval, line 323:
```suggestion
```

matoval (changes requested):
Great work, I like the way you check if context was canceled.

I do have a suggestion and I know it was mentioned before but I'm not a fan of calling defer connCancel() inside the goroutine because the connCancel is only called passively when the goroutine is done. I prefer calling connCancel explicitly in the else if conn.RemoteAddr isn't ok.

pkg/netceptor/conn.go
arrestle: Hi @matoval,

Thanks for the review! Regarding calling `connCancel()` directly in the else block:

I'm concerned about the timing - if we call `connCancel()` in the else block, the context would be canceled before `li.SendResult()` sends the connection to the caller. The caller would receive a connection with an already-canceled context.

`ok == false` is a type assertion check (checking if RemoteAddr is a Receptor Addr type), not indicating the connection is broken. Even in the else case, the connection works normally - it can't use MonitorUnreachable but still needs proper lifecycle cleanup.

The defer in the goroutine lets the connection be used first, then cleans up when it ends. Would this approach work for your concerns?

See new code comments at lines 246-248 and 332 for a more detailed explanation.

Thanks!

matoval: Thanks for the explanation! This makes sense

matoval (approved):
lgtm

PR #1444, opened 2025-10-13
Write kube error to stdoutErr when terminated container reason isn't completed or error

matoval:
Write kube error to stdoutErr when terminated container reason isn't completed or error.

When kubenetes container state is terminated we need to know if the job completed execution or was interrupted. If the terminated reason is completed the job finished and if the reason is error the job returned an error but completed execution. If there reason for termination is anything else this means the execution was interrupted and we want to write the error to stdoutErr.

### DIFF
--- a/pkg/workceptor/command_test.go
+++ b/pkg/workceptor/command_test.go
@@ -115,11 +115,14 @@ func TestStart(t *testing.T) {
 
 	mockBaseWorkUnit.EXPECT().GetWorkceptor().Return(w).Times(2)
 	mockNetceptor.EXPECT().GetLogger().Times(2)
-	mockBaseWorkUnit.EXPECT().UpdateBasicStatus(gomock.Any(), gomock.Any(), gomock.Any())
+	// Expect two calls to UpdateBasicStatus:
+	// 1. First call from Start() with "Launching command runner"
+	// 2. Second call from runCommand() when cmd.Start() fails (receptor not in PATH)
+	mockBaseWorkUnit.EXPECT().UpdateBasicStatus(gomock.Any(), gomock.Any(), gomock.Any()).Times(2)
 	statusExpectCalls(mockBaseWorkUnit)
 
 	mockBaseWorkUnit.EXPECT().UnitDir()
-	mockBaseWorkUnit.EXPECT().UpdateFullStatus(gomock.Any())
+	// When cmd.Start() fails, UpdateFullStatus and MonitorLocalStatus are not called
 	mockBaseWorkUnit.EXPECT().MonitorLocalStatus().AnyTimes()
 	mockBaseWorkUnit.EXPECT().UpdateFullStatus(gomock.Any()).AnyTimes()
 	wu.Start()
--- a/pkg/workceptor/command_test.go
+++ b/pkg/workceptor/command_test.go
@@ -115,14 +115,10 @@ func TestStart(t *testing.T) {
 
 	mockBaseWorkUnit.EXPECT().GetWorkceptor().Return(w).Times(2)
 	mockNetceptor.EXPECT().GetLogger().Times(2)
-	// Expect two calls to UpdateBasicStatus:
-	// 1. First call from Start() with "Launching command runner"
-	// 2. Second call from runCommand() when cmd.Start() fails (receptor not in PATH)
-	mockBaseWorkUnit.EXPECT().UpdateBasicStatus(gomock.Any(), gomock.Any(), gomock.Any()).Times(2)
+	mockBaseWorkUnit.EXPECT().UpdateBasicStatus(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
 	statusExpectCalls(mockBaseWorkUnit)
 
 	mockBaseWorkUnit.EXPECT().UnitDir()
-	// When cmd.Start() fails, UpdateFullStatus and MonitorLocalStatus are not called
 	mockBaseWorkUnit.EXPECT().MonitorLocalStatus().AnyTimes()
 	mockBaseWorkUnit.EXPECT().UpdateFullStatus(gomock.Any()).AnyTimes()
 	wu.Start()
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -540,6 +540,11 @@ mainLoop:
 							podName,
 							WorkerContainerName,
 							containerState.Terminated.ExitCode, containerState.Terminated.Reason, containerState.Terminated.Message)
+						*stdoutErr = fmt.Errorf("pod %s/%s terminated with exit code %d: %s",
+							podNamespace,
+							podName,
+							containerState.Terminated.ExitCode,
+							containerState.Terminated.Message)
 					}
 
 					// We need to check if last line has data
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -544,21 +544,20 @@ mainLoop:
 							WorkerContainerName,
 							containerState.Terminated.ExitCode, containerState.Terminated.Reason, containerState.Terminated.Message)
 						*stdoutErr = fmt.Errorf("pod %s/%s execution interrupted: reason=%s, exit=%d, msg=%s",
-              podNamespace,
-              podName,
-              reason,
-              containerState.Terminated.ExitCode,
-              containerState.Terminated.Message)
+							podNamespace,
+							podName,
+							reason,
+							containerState.Terminated.ExitCode,
+							containerState.Terminated.Message)
 					}
 
 					// Log completion (whether success or error)
 					kw.GetWorkceptor().nc.GetLogger().Info("%s/%s: %s terminated with reason=%s, exit=%d",
-							podNamespace,
-							podName,
-							WorkerContainerName,
-							reason,
-							containerState.Terminated.ExitCode)
-
+						podNamespace,
+						podName,
+						WorkerContainerName,
+						reason,
+						containerState.Terminated.ExitCode)
 
 					// We need to check if last line has data
 					if line != "" {
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -13,6 +13,7 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -538,7 +539,8 @@ mainLoop:
 						reason := containerState.Terminated.Reason
 						// Whitelist: "Completed" and "Error" mean the program ran to completion
 						// Everything else (OOMKilled, Evicted, etc.) means execution was interrupted
-						if reason != "Completed" && reason != "Error" {
+						allowedReasons := []string{"Completed", "Error"}
+						if !slices.Contains(allowedReasons, reason) {
 							kw.GetWorkceptor().nc.GetLogger().Warning("%s/%s: %s execution was interrupted, exit code: %d, terminated reason: %s and terminated message: %s",
 								podNamespace,
 								podName,
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -533,7 +533,7 @@ mainLoop:
 
 					return
 				case containerState.Terminated != nil:
-					
+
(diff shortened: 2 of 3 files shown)


pkg/workceptor/kubernetes.go
```diff
@@ -534,12 +534,23 @@ mainLoop:
...
+						*stdoutErr = fmt.Errorf("pod %s/%s terminated with exit code %d: %s",
+							podNamespace,
+							podName,
+							containerState.Terminated.ExitCode,
+							containerState.Terminated.Message)
+					} else if containerState.Terminated.ExitCode == 1 {
+						kw.GetWorkceptor().nc.GetLogger().Info("%s/%s: %s has terminated with exit code 1 (treated as success)",
+							podNamespace,
+							podName,
+							WorkerContainerName)
```
dleehr: Style feedback on this section:

I think this does a good job of calling out **why** the special case is handled differently, but this still seems like a tricky piece of inverted logic. I find it helps to have these special cases make things as obvious/explicit as possible, something like:

```golang
if exitcode != 0 {
  kw.GetWorkceptor().nc.GetLogger().Info(...)

  // Typically nonzero means an error and we should set stdoutError to indicate failure, 
  // but we treat 1 as a special case because our most common workload signals 1 to indicate overall success but task failures.
  if exitcode == 1 {
    // Log that this is treated as success and do not set stdoutErr. WorkUnit will be marked as succeeded
    kw.GetWorkceptor().nc.GetLogger().Info("handling exit code 1 as success")
  } else {
    // clear failure. Set stdoutErr so that it WorkUnit is marked as failure
    *stdoutErr = fmt.Errorf(...)
  } 
}
```

Nothing wrong with the proposed logic, just trying to leave us clues for future refactorings. I don't think receptor should be responsible for handling this special case, so I think it's worth a few extra comments justifying it 😄

dleehr (approved):
Meta-feedback: On the title and description - how do these changes change the behavior of error handling when Kubernetes evicts a Pod? ~~it looks to me like the changes are focused on handling exitcode 1, which is documented to be an expected exitcode from `ansible-runner` when it completes (i.e. was not evicted) but one or more internal operations (tasks) failed. Am I missing something?~~ 

Edit: I see now. Previously we were only setting `stdoutErr` if we failed to get that last line (regardless of exit code) With this change we will set `stdoutErr` if the exitcode is anything other than `0` or `1`. So my feedback around this - should we `return` after setting `stdoutErr`? If we don't return, might we overwrite `stdoutErr` later (when processing the line?)

Substance: On the exitcode 1 case - you've identified that to execute `ansible-runner` and process its results appropriately, we need to handle ExitCode `1` differently than other nonzero codes (treat it as success), which is a great catch!

Overall this is a great step that will improve outcomes. My thoughts are that it's more than just a special case. It binds receptor's behavior to an implementation detail from a specific workload type (`ansible-runner` exits 1 to mean partial success) instead of following the [established design](https://en.wikipedia.org/wiki/Exit_status#POSIX). So I'd like to understand that a bit more.

The questions I have:
- Would other (non-`ansible-runner`)  use cases be negatively impacted here?
- Is this a regression we introduced since 1.5.7 and we're now correcting? 

It's probably the right move to make here, just want to understand the contours there.

Approving since my only real feedback here is style and I want to know more 😄

pkg/workceptor/kubernetes.go
```diff
@@ -534,12 +534,23 @@ mainLoop:
...
 							podName,
 							WorkerContainerName,
 							containerState.Terminated.ExitCode, containerState.Terminated.Reason, containerState.Terminated.Message)
+						*stdoutErr = fmt.Errorf("pod %s/%s terminated with exit code %d: %s",
+							podNamespace,
+							podName,
+							containerState.Terminated.ExitCode,
+							containerState.Terminated.Message)
+					} else if containerState.Terminated.ExitCode == 1 {
+						kw.GetWorkceptor().nc.GetLogger().Info("%s/%s: %s has terminated with exit code 1 (treated as success)",
```
tyraziel: Do we evaluate what receptor is executing along with the "non-zero" exit code?

In other words, exit code 1 is ansible-runner specific for the playbook execution when a task fails - so should we treat all exit code 1s like this or not?

tyraziel: I would also remove (treated as success)

dleehr (changes requested):
@PabloHiro raised a good point about the other controlled exit codes, some noted in https://github.com/ansible/ansible/issues/19720

I'm going to mark this changes requested until we have a decision on what to do here.

PabloHiro (approved):
LGTM, looking better now :)

pkg/workceptor/command_test.go
dleehr, line 1: These changes in `command_test.go` - are they related to the changes in `kubernetes.go`?

pkg/workceptor/kubernetes.go
```diff
@@ -533,15 +533,32 @@ mainLoop:
...
 					return
 				case containerState.Terminated != nil:
-					// We got EOF and the pod terminated, we will log the terminated information
-					if containerState.Terminated.ExitCode != 0 {
+					reason := containerState.Terminated.Reason
+
+					// Whitelist: "Completed" and "Error" mean the program ran to completion
+					// Everything else (OOMKilled, Evicted, etc.) means execution was interrupted
+					if reason != "Completed" && reason != "Error" {
 						kw.GetWorkceptor().nc.GetLogger().Info("%s/%s: %s has terminated, with nonzero exit code: %v, terminated reason: %v and terminated message: %v",
```
dleehr: This line will log that the container terminated with a `nonzero exit code`. But we're no longer checking the exit code value, so I don't think it makes sense to assume it's nonzero and hard-code that in the log message. (In practice it's _probably_ going to be nonzero if the reason wasn't `Completed` but it's also an assumption)

I think this log message would work better to log that the container terminated with an `unexpected reason` since that's what we're checking. That said, we have another `GetLogger.Info()` just after this `if {}` block (`// Log completion`) with most of the same information (it's just omitting the `.Message`). 

Can we consolidate these into one `GetLogger.Info()` on completion?

pkg/workceptor/kubernetes.go
```diff
@@ -533,15 +533,32 @@ mainLoop:
 
 					return
 				case containerState.Terminated != nil:
-					// We got EOF and the pod terminated, we will log the terminated information
-					if containerState.Terminated.ExitCode != 0 {
+					reason := containerState.Terminated.Reason
+
+					// Whitelist: "Completed" and "Error" mean the program ran to completion
+					// Everything else (OOMKilled, Evicted, etc.) means execution was interrupted
+					if reason != "Completed" && reason != "Error" {
```
dleehr: I think Pablo already gave feedback and helped fix this from a `||` to an `&&`, so I'll offer a readability suggestion for next time. I wonder if Sonar is suggesting these 🤔 

When we have a set of allowed values (or a set of forbidden ones) I think it's **far** easier to read (and maintain) if we check for membership in a collection (e.g. an array or slice) rather than writing an if statement that checks for a mismatch against multiple constants and `&&`ing all of those conditions together:

```go
func checkReason(reason string) {
	allowedReasons := []string{"Completed", "Error"}
	if !slices.Contains(allowedReasons, reason) {
		fmt.Printf("%q is not an allowed reason\n", reason)
	} else {
		fmt.Printf("%q - now that's a good reason\n", reason)
	}
}

func main() {
	checkReason("OOMKilled") // "OOMKilled" is not an allowed reason
	checkReason("Completed") // "Completed" - now that's a good reason
}
```

dleehr (changes requested):
Thanks for going back and digging in to the exit code behaviors. It sounds like the behavior of our most common workload (`ansible-runner`) makes it unreasonable to assume that nonzero exit codes indicate a total failure. Therefore in the receptor logic, a nonzero exit code should not be enough to set `stdoutErr` and mark the work failed. I appreciate exploring the `containerState.Terminated.Reason` as an alternative to make that decision. I think this is likely the right move but I would like to get a bit more clarity and defensiveness.

The proposed change is to ignore the exit code and check `.Reason` against one of two hard-coded strings (`Completed` or `Error`). Basing the behavior entirely on these hard-coded strings is a fragile assumption and I'd like some more context there.

This would bind decisions (success or failure) to the terminated Reason reported by Kubernetes. If I understand correctly, the assumption is `Completed` means total success (likely an exit code `0` but we don't check), and that `Error` means Kubernetes ran the process ran to completion but the process exited with some error (likely nonzero exit). These aren't different code paths, so please correct me if that's not accurate. Finally, we're assuming any reason other than these two means the process did not complete (e.g. due to an eviction or OOM Kill).

How did we arrive at this list of two strings? Is there documentation/data/evidence that supports these values? Is there documentation somewhere that tells us `Completed` is the exact string we should look for? Never `completed` or `success` or `Finished`? Are these consistent across all versions of Kubernetes? Are there plans to change or expand? I reviewed the reference for the ContainerStateTerminated spec - https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#containerstateterminated-v1-core and it doesn't provide a list of expected values, just that it's a `string` indicating a `(brief) reason from the last termination of the container`. 

Some other questions:

- Under what conditions would we get false positives/false negatives? - how will we get failures reported as success or success reported as failures?
- I understand why we can't exclusively use exit codes, but I still consider an exit code of zero to be a golden signal that clearly indicates success. What do you think about treating exit code zero as successful termination and only checking the reason (and possibly setting `stdoutErr` if the exit code was nonzero?)

Overall I like the direction. Exit codes are clear but unfortunately don't completely cover our bases. I think the direction to use the `Reason` makes sense, but that brings in different assumptions and I think we need to be as defensive as possible.

pkg/workceptor/kubernetes.go
PabloHiro: Oh nono, I messed up and thought `||` was correct, but it is not, so I deleted my comment. But yeah, it is a symptom that it is not easy to read.

dleehr: Thanks for clarifying - I misunderstood the misunderstanding 🤣! But that explains why I couldn't find the comment anymore. Pointing out hard-to-read logic is valuable feedback! Let's use that as a signal to give our brains a break and let the computers do the hard work 😁.

pkg/workceptor/command_test.go
matoval: Yes, because of the changes in kubernetes.go the EXPECTs are different

matoval: Yes because of the change in kubernetes.go command_test.go has to account for different, that be said adding a check for exit code zero might change this again.

pkg/workceptor/kubernetes.go
matoval: I'm going to add a check for exit code zero before checking the terminated reason so this message might still make sense. I'll clean it up.

pkg/workceptor/kubernetes.go
```diff
@@ -533,13 +534,43 @@ mainLoop:
...
+								containerState.Terminated.Message)
+						}
+					} else {
+						// Log successful completion
+						kw.GetWorkceptor().nc.GetLogger().Info("%s/%s: %s completed successfully",
 							podNamespace,
 							podName,
-							WorkerContainerName,
-							containerState.Terminated.ExitCode, containerState.Terminated.Reason, containerState.Terminated.Message)
+							WorkerContainerName)
```
dleehr: I like this behavior 👍 

nit: The only note I have on readability to [check positive cases first](https://testing.googleblog.com/2023/10/improve-readability-with-positive.html). We basically have

```golang
if ! exitcode == 0 {
  if ! slices.Contains(allowedReasons, reason) {
    // failure, set stdoutErr
  } else {
    // acceptable failure, don't set stdoutErr
} else {
  // success
}
```

But I think it's easier to read if we can get rid of the `!` when we don't need them:

```golang
if exitcode == 0 {
  // success
} else {
  if slices.Contains(allowedReasons, reason) {
    // acceptable failure, don't set stdoutErr
  } else {
    // failure, set stdoutErr
}
```

I've heard arguments for both ways - what do you think?

dleehr (approved):
LGTM. Awesome work! Leaving one comment on readability - but just a matter of opinion.

pkg/workceptor/kubernetes.go
matoval: That makes sense, I agree it's better readability

PR #1445, opened 2025-10-14
Sonar Cloud workflow fixes

davemulford:
These changes fix two main issues:

1. A typo in the sonar cloud workflow (was looking for "Code Coverage" when it should have looked for "Codecov").
2. The actions in `sonar_checks.yml` was triggered by two workflows: one that reported coverage for `receptor` and another for `receptorctl`.

Actions that are triggered by other actions completing only wait for _a single_ parent action to complete. This caused a problem where the sonar cloud workflow would run as soon as one of the coverage reporting workflows completed. If the other coverage reporting workflow hadn't finished, the expected code coverage artifact would not exist, causing the sonar cloud workflow to fail.

This also deletes the Codecov reporting tool which we are migrating away from.


tyraziel (approved):
LGTM

matoval (approved):

PR #1448, opened 2025-10-20
Add receptor and receptorctl to coverage upload

davemulford:
Make receptor and receptorctl test coverage uploading happen in a single step. SonarQube expects it to happen this way, instead of running multiple scans.


dleehr (approved):

tyraziel (approved):
CI :)

PR #1452, opened 2025-10-21
Fix "Error locating unit" in execution node

PabloHiro:
AAP-51624

## The bug
Every time a work unit gets released in a controller node of receptor, the following lines are logged in the corresponding execution node:
```
ERROR 2025/08/06 16:04:59 Error locating unit: aapp01ZOF8tB3W {"node_id":"aap-exec01.example.com","remote_id":"aap-ctrl01.example.com"}
ERROR 2025/08/06 16:04:59 : unknown work unit aapp01ZOF8tB3W  {"remote_id":"aap-ctrl01.example.com","node_id":"aap-exec01.example.com"} 
```

## Root cause
The lines are logged in the execution node because every time `Release` is called in the controller node, the following callstack is triggered in the controller node:

https://github.com/ansible/receptor/blob/28517aed1a91e9dab1561e0014db6eada5faee6d/pkg/workceptor/remote_work.go#L705-L707

`cancelOrRelease` calls `runAndMonitor`:

https://github.com/ansible/receptor/blob/28517aed1a91e9dab1561e0014db6eada5faee6d/pkg/workceptor/remote_work.go#L661

`runAndMonitor` calls `monitorRemoteUnit`:
https://github.com/ansible/receptor/blob/28517aed1a91e9dab1561e0014db6eada5faee6d/pkg/workceptor/remote_work.go#L694-L696

`monitorRemoteUnit` calls `monitorRemoteStatus`:
https://github.com/ansible/receptor/blob/28517aed1a91e9dab1561e0014db6eada5faee6d/pkg/workceptor/remote_work.go#L581-L588

`monitorRemoteStatus`:
https://github.com/ansible/receptor/blob/28517aed1a91e9dab1561e0014db6eada5faee6d/pkg/workceptor/remote_work.go#L513-L517

And the problem with `monitorRemoteStatus` is that it submits a `work status` command right after the work unit was released in `cancelOrRelease`:

https://github.com/ansible/receptor/blob/28517aed1a91e9dab1561e0014db6eada5faee6d/pkg/workceptor/remote_work.go#L307

So when `work status` command reaches the execution node, it triggers the bug, because the work unit has been released already.

## How to fix this?

* There are related issues such as https://github.com/ansible/receptor/pull/1275 (AAP-22149) that fix this by adding a new state, `Released`. However this would require additional changes in Controller.
* Discussion in AAP-22149 suggests modifying status so an additional flag can be passed to ignore errors and supress logs.

However, what I implemented in this PR is removing the submission of the `work status` command entirely. My reasoning is as follows:
* The status retrieved from the execution node does not seem to be used for anything. The local status in the controller gets updated to `LocalCancelled` and `LocalReleased` before the status command gets submitted.
* What is the point of retrieving the status of a work unit we want to release anyway?
* This fix will get rid of the consistent and reproducible bug.
* This fix will NOT get rid of real race-conditions where a status call is submitted at the same time a release call is. BUT we don't have proof of seeing this in the wild.

### DIFF
--- a/pkg/workceptor/remote_work_test.go
+++ b/pkg/workceptor/remote_work_test.go
@@ -434,9 +434,8 @@ func TestRemoteWorkLifecycleOperations(t *testing.T) {
 					"execution\n", // Hello message with remote node ID
 					"{\"State\": 4, \"Detail\": \"Cancelled\", \"StdoutSize\": 0}\n", // Acknowledgment to release command
 				}
-				anyTimes := true // Needed because this will monitor the remote work unit in a loop
-				createRemoteWorkNetworkSetup(t, ctrl, contextWithCancel, messages, mockNetceptor, mockBaseWorkUnit, tmpDir, remoteExtraData, anyTimes)
-				mockBaseWorkUnit.EXPECT().Release(false).Return(nil).AnyTimes() // Called after monitoring completes (may not complete before context cancels)
+				createRemoteWorkNetworkSetup(t, ctrl, contextWithCancel, messages, mockNetceptor, mockBaseWorkUnit, tmpDir, remoteExtraData, false)
+				mockBaseWorkUnit.EXPECT().Release(false).Return(nil)
 				err = wu.Release(false)
 			case "force_release_already_started":
 				messages := []string{
--- a/pkg/workceptor/remote_work.go
+++ b/pkg/workceptor/remote_work.go
@@ -328,13 +328,11 @@ func (rw *remoteUnit) monitorRemoteStatus(mw *utils.JobContext, forRelease bool)
 		}
 		if status[:5] == "ERROR" {
 			if strings.Contains(status, "unknown work unit") {
-				if !forRelease {
-					rw.GetWorkceptor().nc.GetLogger().Debug("Work unit %s on node %s is gone.\n", remoteUnitID, remoteNode)
-					rw.UpdateFullStatus(func(status *StatusFileData) {
-						status.State = WorkStateFailed
-						status.Detail = "Remote work unit is gone"
-					})
-				}
+				rw.GetWorkceptor().nc.GetLogger().Debug("Work unit %s on node %s is gone.\n", remoteUnitID, remoteNode)
+				rw.UpdateFullStatus(func(status *StatusFileData) {
+					status.State = WorkStateFailed
+					status.Detail = "Remote work unit is gone"
+				})
 
 				return
 			}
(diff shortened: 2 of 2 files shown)


arrestle:
Nice work Pablo! What do you think of a mutex on the `remoteUnit` struct ? This mutex could be locked in `cancelOrRelease` when release is true to avoid releasing the same unit twice.

Also, in `monitorRemoteStatus`, can we `red, ok := currentStatus.ExtraData.(*RemoteExtraData)` and see if the remote unit is released, and if it is, suppress the error?

PabloHiro:
> Nice work Pablo! What do you think of a mutex on the `remoteUnit` struct ? This mutex could be locked in `cancelOrRelease` when release is true to avoid releasing the same unit twice.
> 
> Also, in `monitorRemoteStatus`, can we `red, ok := currentStatus.ExtraData.(*RemoteExtraData)` and see if the remote unit is released, and if it is, suppress the error?

If two simultaneous releases happen, with this fix the controller node would log:
```
ERROR 2025/10/29 12:19:44 Error locating unit: controlXbTsvr8L {"remote_id":"hop","node_id":"control"}
ERROR 2025/10/29 12:19:44 : unknown work unit controlXbTsvr8L {"node_id":"control","remote_id":"hop"}
```
Which is triggered by the second release command submission. But no errors would be logged in the execution node.

However, without the fix, in addition to the above, the execution node would throw:
```
ERROR 2025/10/29 12:22:34 Error locating unit: controlM37ZmF7B {"node_id":"execution","remote_id":"hop"}
ERROR 2025/10/29 12:22:34 : unknown work unit controlM37ZmF7B {"node_id":"execution","remote_id":"hop"}
```

The bug report refers to this error in the **execution** node, so this addresses it.

arrestle (approved):
Good work Pablo!

arrestle (approved):

matoval (approved):
Great work, tested it locally didn't get "Error locating unit" or "unknown work unit" with this fix like we do with devel.

pkg/workceptor/remote_work_test.go
```diff
@@ -434,9 +434,8 @@ func TestRemoteWorkLifecycleOperations(t *testing.T) {
 					"execution\n", // Hello message with remote node ID
 					"{\"State\": 4, \"Detail\": \"Cancelled\", \"StdoutSize\": 0}\n", // Acknowledgment to release command
 				}
-				anyTimes := true // Needed because this will monitor the remote work unit in a loop
-				createRemoteWorkNetworkSetup(t, ctrl, contextWithCancel, messages, mockNetceptor, mockBaseWorkUnit, tmpDir, remoteExtraData, anyTimes)
-				mockBaseWorkUnit.EXPECT().Release(false).Return(nil).AnyTimes() // Called after monitoring completes (may not complete before context cancels)
+				createRemoteWorkNetworkSetup(t, ctrl, contextWithCancel, messages, mockNetceptor, mockBaseWorkUnit, tmpDir, remoteExtraData, false)
+				mockBaseWorkUnit.EXPECT().Release(false).Return(nil)
```
PabloHiro, line 438: @matoval here I am removing the expectation that monitoring is performed

matoval (approved):
LGTM

PR #1456, opened 2025-10-22
[AAP-51479] Move ReceptorVerifyFunc to tlsconfig.go and write unit tests

arrestle:
# Refactor: Move ReceptorVerifyFunc to tlsconfig.go

## Summary

Moves certificate verification code from `netceptor.go` (2,089 lines) to `tlsconfig.go` (487 lines) and refactors into 7 testable helper functions (4 primary + 3 utilities). All TLS-related code is now grouped together and comprehensively tested with 18 test cases across 7 test functions achieving 92.6% code coverage.

**Key Improvements:**
- ✅ **No breaking changes** - 100% backward compatible
- ✅ **Performance optimization** - Hash caching reduces redundant computations (up to 80% reduction with multiple same-algorithm fingerprints)
- ✅ **Enhanced algorithm support** - Added SHA-224 and SHA-384 to existing SHA-256/SHA-512
- ✅ **Better testability** - From 0% to 92.6% code coverage with comprehensive unit tests

## Motivation

- Improves code organization for AAP-51479 investigation
- Groups all TLS configuration and verification logic in one file
- Reduces size of oversized `netceptor.go` (2,089 lines)
- Extracts helper functions for better testability
- Adds comprehensive unit test coverage (92.6%) for certificate verification logic
- **Note:** Certificate deduplication was investigated but rejected to avoid risk of invalidating user-provided certificate bundles

### Before: Monolithic Function
```mermaid
graph TD
    A[ReceptorVerifyFunc<br/>170 lines, 0% tested] --> A1[Parse certs]
    A1 --> A2[Verify fingerprints<br/>57 lines inline]
    A2 --> A3[Build options<br/>26 lines inline]  
    A3 --> A4[Add intermediates]
    A4 --> A5[Verify chain]
    A5 --> A6[Verify node ID<br/>11 lines inline]
    
    style A fill:#ffffff,stroke:#666666,stroke-width:2px,color:#000000
    style A1 fill:#ffffff,stroke:#666666,stroke-width:1px,color:#000000
    style A2 fill:#ffffff,stroke:#666666,stroke-width:1px,color:#000000
    style A3 fill:#ffffff,stroke:#666666,stroke-width:1px,color:#000000
    style A4 fill:#ffffff,stroke:#666666,stroke-width:1px,color:#000000
    style A5 fill:#ffffff,stroke:#666666,stroke-width:1px,color:#000000
    style A6 fill:#ffffff,stroke:#666666,stroke-width:1px,color:#000000
```

[...]

PR #1457, opened 2025-10-23
[AAP-51479] Add error handling when certs overflow the QUIC RFC buffer size

arrestle:
# [AAP-51479] Add error handling when certs overflow the QUIC RFC buffer size

## Summary

When QUIC CRYPTO_BUFFER_EXCEEDED error occurs provide feedback and information on how to self-resolve.

## Problem

QUIC handshakes exceed 16,384-byte crypto buffer when peer intermediates duplicate certificates already in the chain. Enterprise scenario: 154 CAs × 98 bytes = 18,342 bytes (fails at 17,035 byte threshold).

## Solution

Provide direction for receptor users and KCS number for Red Hat customers

## Impact

Clarify customer responsibilities

## Notes for Reviewers

- Depends on PR #1 (helper function extraction)

---

AAP-51479 | Depends on PR #1456

Co-authored-by: Claude <claude@anthropic.com>


pkg/workceptor/remote_work.go
```diff
@@ -90,7 +90,19 @@ func (rw *remoteUnit) GetConnection(ctx context.Context) (net.Conn, *bufio.Reade
 		rw.GetWorkceptor().nc.GetLogger().Info("Connection to %s failed with error: %s",
 			rw.Status().ExtraData.(*RemoteExtraData).RemoteNode, err)
 		errStr := err.Error()
-		if strings.Contains(errStr, "CRYPTO_ERROR") {
+		if strings.Contains(errStr, "CRYPTO_BUFFER_EXCEEDED") {
+			shouldExit := false
+			rw.UpdateFullStatus(func(status *StatusFileData) {
+				status.Detail = fmt.Sprintf("QUIC crypto buffer exceeded. CA bundle may be too large (limit: 16KB). See KCS 7129200: %s", errStr)
+				if !status.ExtraData.(*RemoteExtraData).RemoteStarted {
```
davemulford: Do we know that `status.ExtraData` is never `nil`? We may want to check that before attempting the type conversion. This will prevent a possible panic condition.

pkg/workceptor/remote_work_test.go
```diff
@@ -589,3 +589,86 @@ func TestRemoteWorkConnectToRemoteEnhanced(t *testing.T) {
...
+		statusLock := &sync.RWMutex{}
+		var capturedDetail string
+
+		// Set up mocks
+		mockBaseWorkUnit.EXPECT().GetStatusLock().Return(statusLock).AnyTimes()
+		mockBaseWorkUnit.EXPECT().GetStatusWithoutExtraData().DoAndReturn(func() *workceptor.StatusFileData {
+			return &workceptor.StatusFileData{}
+		}).AnyTimes()
+		mockBaseWorkUnit.EXPECT().GetStatusCopy().Return(workceptor.StatusFileData{
+			ExtraData: remoteExtraData,
```
davemulford: Should test for the case where `ExtraData` is `nil`.

pkg/workceptor/remote_work_test.go
```diff
@@ -589,3 +589,86 @@ func TestRemoteWorkConnectToRemoteEnhanced(t *testing.T) {
...
+
+		// Verify it's a helpful message
+		expectedPhrases := []string{"QUIC crypto buffer exceeded", "CA bundle", "too large"}
+		for _, phrase := range expectedPhrases {
+			if !strings.Contains(capturedDetail, phrase) {
+				t.Errorf("Expected status detail to contain '%s', got: %s", phrase, capturedDetail)
+			}
+		}
+
+		ctrl.Finish()
```
davemulford: Should this be called with a `defer` statement right after being created? See the [go mock documentation](https://pkg.go.dev/github.com/golang/mock/gomock).

davemulford (changes requested):
Looks good overall! A couple of minor things that I've called out in comments (handling nil case and using defer in tests).

pkg/workceptor/remote_work.go
```diff
@@ -90,7 +90,19 @@ func (rw *remoteUnit) GetConnection(ctx context.Context) (net.Conn, *bufio.Reade
 		rw.GetWorkceptor().nc.GetLogger().Info("Connection to %s failed with error: %s",
 			rw.Status().ExtraData.(*RemoteExtraData).RemoteNode, err)
 		errStr := err.Error()
-		if strings.Contains(errStr, "CRYPTO_ERROR") {
+		if strings.Contains(errStr, "CRYPTO_BUFFER_EXCEEDED") {
+			shouldExit := false
```
PabloHiro: Instead of repeating the same pattern as in the condition below, what do you think of simplifying? Example:

```go
		var detail string
		if strings.Contains(errStr, "CRYPTO_BUFFER_EXCEEDED") {
			detail = fmt.Sprintf("QUIC crypto buffer exceeded. CA bundle may be too large (limit: 16KB). See KCS 7129200: %s", errStr)
		} else if strings.Contains(errStr, "CRYPTO_ERROR") {
			detail = fmt.Sprintf("TLS error connecting to remote service: %s", errStr)
		}

		if detail != "" {
			shouldExit := false
				status.Detail = detail
				if !status.ExtraData.(*RemoteExtraData).RemoteStarted {
					shouldExit = true
					status.State = WorkStateFailed
				}
			})
			if shouldExit {
				return nil, nil
			}
		}
```

pkg/workceptor/remote_work.go
arrestle: After some research, adding defensive code to every access of status.ExtraData in 18 different places feels like bloat since we always call `rw.SetStatusExtraData(red)` in the [the constructor](https://github.com/ansible/receptor/blob/6b20e21601c159b2ac16fcdf53d1790b42adc56e/pkg/workceptor/remote_work.go#L718). 

Let me know if you agree.

pkg/workceptor/remote_work_test.go
arrestle: See above

pkg/workceptor/remote_work_test.go
arrestle: Good catch, however, I've determined that the test Cleanup command will automatically Finish() any controllers with *testing.T passed in, so it turns out we don't need to defer or call it at all.  See **Note:** below

```Go
// Finish checks to see if all the methods that were expected to be called were called.
// It is not idempotent and therefore can only be invoked once. See the **Note** below.
//
// Note: If you pass a *testing.T into [NewController], you no longer
// need to call ctrl.Finish() in your test methods.
func (ctrl *Controller) Finish() {
	// If we're currently panicking, probably because this is a deferred call.
	// This must be recovered in the deferred function.
	err := recover()
	ctrl.finish(false, err)
}
```

pkg/workceptor/remote_work.go
davemulford: You're right, that would be a different (and much larger) task. I want to be really clear: I'm definitely not asking for that.

My suggestion is only for the new instance you've added in this PR.

My reasoning is that the underlying field being asserted could potentially become nil between when the struct is initialized and when this assertion happens. What are your thoughts on adding that check just for the new code in this PR?

pkg/workceptor/remote_work_test.go
davemulford: Very nice!

davemulford (approved):
LGTM :shipit:

pkg/workceptor/remote_work.go
```diff
@@ -90,15 +90,27 @@ func (rw *remoteUnit) GetConnection(ctx context.Context) (net.Conn, *bufio.Reade
...
 			rw.Status().ExtraData.(*RemoteExtraData).RemoteNode, err)
 		errStr := err.Error()
-		if strings.Contains(errStr, "CRYPTO_ERROR") {
+
+		var detail string
+		if strings.Contains(errStr, "CRYPTO_BUFFER_EXCEEDED") {
+			detail = fmt.Sprintf("QUIC crypto buffer exceeded. CA bundle may be too large (limit: 16KB). See KCS 7129200: %s", errStr)
+		} else if strings.Contains(errStr, "CRYPTO_ERROR") {
+			detail = fmt.Sprintf("TLS error connecting to remote service: %s", errStr)
+		}
```
matoval, line 100: What if the error doesn't contain `CRYPTO_ERROR` or `CRYPTO_BUFFER_EXCEEDED` ?

matoval (changes requested):
What other possible error might we get if the connect to remote fails?

PabloHiro (approved):
LGTM

pkg/workceptor/remote_work.go
PabloHiro: It retries in a loop until context is cancelled. I think that if we try to log the rest of the errors it could find here we have the risk of flooding `receptor.log`

matoval (approved):

PR #1458, opened 2025-10-24
Test coverage for remote_work.go

PabloHiro:
AAP-51624

This is needed to have coverage for: https://github.com/ansible/receptor/pull/1452

Changes:
* Deleted unit tests that did not provide any additional coverage, and did not validate anything: `TestRemoteWorkConnectToRemoteMissingExtraData`, `TestRemoteWorkStartRemoteUnitInvalidResponse`, `TestRemoteWorkGetConnection`, `TestRemoteWorkConnectAndRun`, `TestRemoteWorkStartRemoteUnit`, `TestRemoteWorkCancelOrReleaseRemoteUnit`, `TestRemoteWorkMonitoring`, `TestRemoteWorkExpiration`.
* Despite their names, the tests above simply validated that the constructor was working as expected.
* Found a way to mock properly the replies received by an execution node when launching commands from a controller node.
* Added tests for the different logic paths in `remote_work.go`. This PR increases the coverage of this file from 19.6% to 59.9%.

### DIFF
--- a/pkg/workceptor/remote_work_test.go
+++ b/pkg/workceptor/remote_work_test.go
@@ -6,24 +6,113 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net"
+	"os"
+	"path/filepath"
 	"strings"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/ansible/receptor/pkg/logger"
+	"github.com/ansible/receptor/pkg/netceptor"
+	"github.com/ansible/receptor/pkg/netceptor/mock_netceptor"
 	"github.com/ansible/receptor/pkg/workceptor"
 	"github.com/ansible/receptor/pkg/workceptor/mock_workceptor"
 	"go.uber.org/mock/gomock"
 )
 
-func createRemoteWorkTestSetup(t *testing.T) (workceptor.WorkUnit, *mock_workceptor.MockBaseWorkUnitForWorkUnit, *mock_workceptor.MockNetceptorForWorkceptor, *workceptor.Workceptor) {
+// createRemoteWorkNetworkSetup creates a mock network Conn for testing remote work operations.
+// It takes a list of messages to be sent to the mock Conn and sets up the mock netceptor and base work unit expectations.
+func createRemoteWorkNetworkSetup(t *testing.T, ctrl *gomock.Controller, ctx context.Context, messages []string, mockNetceptor *mock_workceptor.MockNetceptorForWorkceptor, mockBaseWorkUnit *mock_workceptor.MockBaseWorkUnitForWorkUnit, tmpDir string, remoteExtraData *workceptor.RemoteExtraData, anytimes bool) {
+	t.Helper()
+
+	// Create a mock Conn using the mock interfaces
+	mockPacketConner := mock_netceptor.NewMockPacketConner(ctrl)
+	mockQuicConnection := mock_netceptor.NewMockQuicConnectionForConn(ctrl)
+	mockQuicStream := mock_netceptor.NewMockQuicStreamForConn(ctrl)
+
+	messageIndex := 0
+	// Set up reads to return different messages based on the provided list
+	readExpectation := mockQuicStream.EXPECT().Read(gomock.Any()).DoAndReturn(func(b []byte) (int, error) {
+		if messageIndex < len(messages) {
+			msg := messages[messageIndex]
+			messageIndex++
+			copy(b, msg)
+
+			return len(msg), nil
+		} else {
+			return 0, ctx.Err()
+		}
+	})
+	if anytimes {
+		readExpectation.AnyTimes()
+	} else {
+		readExpectation.Times(len(messages))
+	}
+	mockQuicStream.EXPECT().Write(gomock.Any()).Return(0, nil).AnyTimes()
+	mockQuicStream.EXPECT().Close().Return(nil).AnyTimes()
+	mockQuicStream.EXPECT().SetReadDeadline(gomock.Any()).Return(nil).AnyTimes()
+	mockQuicStream.EXPECT().SetWriteDeadline(gomock.Any()).Return(nil).AnyTimes()
+	mockQuicStream.EXPECT().SetDeadline(gomock.Any()).Return(nil).AnyTimes()
+
+	var cancelFunc context.CancelFunc = func() {}
+	mockPacketConner.EXPECT().Cancel().Return(&cancelFunc).AnyTimes()
+	mockPacketConner.EXPECT().Close().Return(nil).AnyTimes()
+	mockPacketConner.EXPECT().LocalService().Return("test-service").AnyTimes()
+	mockPacketConner.EXPECT().GetLogger().Return(logger.NewReceptorLogger("")).AnyTimes()
+
+	mockQuicConnection.EXPECT().CloseWithError(gomock.Any(), gomock.Any()).Return(nil).AnyTimes()
+	mockQuicConnection.EXPECT().RemoteAddr().Return(&net.TCPAddr{IP: net.ParseIP("127.0.0.1"), Port: 8080}).AnyTimes()
+	mockQuicConnection.EXPECT().LocalAddr().Return(&net.TCPAddr{IP: net.ParseIP("127.0.0.1"), Port: 8080}).AnyTimes()
+
+	mockConn := netceptor.NewConn(
+		netceptor.New(ctx, "test-node"), // s: *Netceptor
+		mockPacketConner,                // pc: PacketConner
+		mockQuicConnection,              // qc: QuicConnectionForConn
+		mockQuicStream,                  // qs: QuicStreamForConn
+		make(chan struct{}),             // doneChan
+		&sync.Once{},                    // doneOnce
+		ctx,                             // ctx
+	)
+
+	// Set up mock netceptor expectations
+	mockNetceptor.EXPECT().GetClientTLSConfig(gomock.Any(), gomock.Any(), gomock.Any()).Return(&tls.Config{}, nil).AnyTimes()
+	mockNetceptor.EXPECT().DialContext(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(mockConn, nil).AnyTimes()
+
+	// Set up common mock base work unit expectations
+	mockBaseWorkUnit.EXPECT().Load().Return(nil).AnyTimes()
+	mockBaseWorkUnit.EXPECT().StdoutFileName().Return(filepath.Join(tmpDir, "stdout")).AnyTimes()
+	mockBaseWorkUnit.EXPECT().UnitDir().Return(tmpDir).AnyTimes()
+	mockBaseWorkUnit.EXPECT().UpdateFullStatus(gomock.Any()).Do(func(updateFunc interface{}) {
+		updateFunc.(func(*workceptor.StatusFileData))(&workceptor.StatusFileData{
+			ExtraData: remoteExtraData,
+		})
+	}).AnyTimes()
+	mockBaseWorkUnit.EXPECT().UpdateBasicStatus(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
+	mockBaseWorkUnit.EXPECT().LastUpdateError().Return(nil).AnyTimes()
+
+	// Create temporary directory and stdin file
+	_ = os.MkdirAll(tmpDir, 0o755)
+	stdinFile, err := os.Create(filepath.Join(tmpDir, "stdin"))
+	if err != nil {
+		t.Errorf("Error creating temporary file: %v", err)
+	} else {
+		stdinFile.Close()
+		t.Cleanup(func() {
+			os.Remove(filepath.Join(tmpDir, "stdin"))
+		})
+	}
+}
+
+// createRemoteWorkTestSetup creates the basic setup for remote work unit tests.
+func createRemoteWorkTestSetup(t *testing.T, ctx context.Context) (workceptor.WorkUnit, *mock_workceptor.MockBaseWorkUnitForWorkUnit, *mock_workceptor.MockNetceptorForWorkceptor, *workceptor.Workceptor, *gomock.Controller) {
+	t.Helper()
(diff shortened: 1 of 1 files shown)


pkg/workceptor/remote_work_test.go
```diff
@@ -6,24 +6,84 @@ import (
...
+	"github.com/ansible/receptor/pkg/netceptor/mock_netceptor"
 	"github.com/ansible/receptor/pkg/workceptor"
 	"github.com/ansible/receptor/pkg/workceptor/mock_workceptor"
 	"go.uber.org/mock/gomock"
 )
 
-func createRemoteWorkTestSetup(t *testing.T) (workceptor.WorkUnit, *mock_workceptor.MockBaseWorkUnitForWorkUnit, *mock_workceptor.MockNetceptorForWorkceptor, *workceptor.Workceptor) {
+// createRemoteWorkNetworkSetup creates a mock network Conn for testing remote work operations.
+// It takes a list of messages to be sent to the mock Conn.
+func createRemoteWorkNetworkSetup(t *testing.T, ctrl *gomock.Controller, ctx context.Context, messages []string) (*netceptor.Conn, *mock_netceptor.MockQuicStreamForConn) {
```
arrestle: Excellent use of a helper function with variable message!

pkg/workceptor/remote_work_test.go
```diff
@@ -220,65 +258,208 @@ func TestRemoteWorkLifecycleOperations(t *testing.T) {
...
-					}
-				}
+				})
 			}
 
 			var err error
-			switch tt.operation {
-			case "start":
+			switch tt.name {
+			case "start not started":
```
arrestle: This switch statement  is difficult to read and likely hard to maintain. Consider separate test functions for this scenario with additional helper functions to help set up mocks to cut down on duplicated code, or, if you prefer to keep a single TestRemoteWorkLifecycleOperations function, consider adding a setupFunc to the tests array, and add the additional mocks to the setupFunc. 

see https://go.dev/blog/subtests for more information.

pkg/workceptor/remote_work_test.go
```diff
@@ -220,65 +258,208 @@ func TestRemoteWorkLifecycleOperations(t *testing.T) {
...
+					updateFunc.(func(*workceptor.StatusFileData))(&workceptor.StatusFileData{
+						ExtraData: remoteExtraData,
+					})
+				}).AnyTimes()
+				mockBaseWorkUnit.EXPECT().UnitDir().Return("/tmp/start_not_started").AnyTimes()
+				// Add expectation for UpdateBasicStatus and LastUpdateError which will be called by the monitoring goroutine
+				mockBaseWorkUnit.EXPECT().UpdateBasicStatus(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
+				mockBaseWorkUnit.EXPECT().LastUpdateError().Return(nil).AnyTimes()
+				// Create temporary file for stdin
+				_ = os.MkdirAll("/tmp/start_not_started", 0o755)
```
arrestle: Pablo, the use of /tmp files is common across our codebase and causes flakiness, better to use t.TempDir() 

```suggestion
// Correct approach:
tempDir := t.TempDir() // Automatically cleaned up after test
stdinPath := filepath.Join(tempDir, "stdin")
if err := os.WriteFile(stdinPath, []byte{}, 0755); err != nil {
    t.Fatalf("Failed to create stdin: %v", err)
}
mockBaseWorkUnit.EXPECT().UnitDir().Return(tempDir).AnyTimes()
```

pkg/workceptor/remote_work_test.go
```diff
@@ -6,24 +6,84 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net"
+	"os"
+	"path"
```
arrestle: https://pkg.go.dev/path/filepath for filenames. path is for urls. helps with linting.
```suggestion
	"filepath"
```

pkg/workceptor/remote_work_test.go
```diff
@@ -6,24 +6,84 @@ import (
...
+
+	// Create a mock Conn using the mock interfaces
+	mockPacketConner := mock_netceptor.NewMockPacketConner(ctrl)
+	mockQuicConnection := mock_netceptor.NewMockQuicConnectionForConn(ctrl)
+	mockQuicStream := mock_netceptor.NewMockQuicStreamForConn(ctrl)
+
+	messageIndex := 0
+	// Set up reads to return different messages based on the provided list
+	mockQuicStream.EXPECT().Read(gomock.Any()).DoAndReturn(func(b []byte) (int, error) {
+		if messageIndex < len(messages) {
```
arrestle, line 37: If you have multiple tests running in parallel this could be a concurrency issue and cause some flakiness with shared mocks.

pkg/workceptor/remote_work_test.go
```diff
@@ -6,24 +6,84 @@ import (
...
+			msg := messages[messageIndex]
+			messageIndex++
+			copy(b, msg)
+
+			return len(msg), nil
+		} else {
+			ctx.Done()
```
arrestle: If you're attempting to wait here use `<- ctx.Done()` to wait for the channel to complete.
If you don't need to wait you can just `return 0, ctx.Err()`

pkg/workceptor/remote_work_test.go
```diff
@@ -220,65 +258,208 @@ func TestRemoteWorkLifecycleOperations(t *testing.T) {
...
+				_ = os.MkdirAll("/tmp/start_not_started", 0o755)
+				_, tmpErr := os.Create(path.Join("/tmp/start_not_started", "stdin")) //nolint:gosec // G304: test file using hardcoded temp path
+				if tmpErr != nil {
+					t.Errorf("Error creating temporary file: %v", tmpErr)
+				}
+				defer os.Remove(path.Join("/tmp/start_not_started", "stdin"))
 				err = wu.Start()
-			case "restart":
+			case "start already started":
+				err = wu.Start()
```
arrestle, line 396: These could cause a memory leak since we aren't waiting to see if they stop. Maybe not SO important in a pipeline?

pkg/workceptor/remote_work_test.go
```diff
@@ -63,7 +125,7 @@ func TestRemoteWorkUnredactedStatus(t *testing.T) {
 
 func TestRemoteWorkSetFromParams(t *testing.T) {
```
arrestle, line 156: The name of this test is confusing, it's really testing only GetStatusCopy. perhaps we need a new Jira to test Get and SetFromParams.

arrestle (changes requested):
Pablo, excellent work on this ambitious and well-thought out PR. Smart to remove the low-value constructor tests and replace with fewer tests and less code but with 3x the coverage.

pkg/workceptor/remote_work_test.go
PabloHiro: I cannot use `TempDir()` here because `stdin` is hardcoded in the code. If I use `TempDir()` it will generate something like `stdin0u401ummd` and the code will complain that it cannot find a `stdin` file.

pkg/workceptor/remote_work_test.go
PabloHiro: I am gonna reply here to explain what I did.

For each sub-test I am creating a new context here:
https://github.com/ansible/receptor/blob/e60692c71235a222fda5c53ec965dae38db3267b/pkg/workceptor/remote_work_test.go#L304-L311

And I am calling `t.Cleanup` to cancel the context. This is the same context that is passed to `createRemoteWorkNetworkSetup`. So every test instance should have its own separate context.

Also since `createRemoteWorkTestSetup` creates a new `gomock` instance for each subtest, every test instance should have its own separate gomock instance.

I faced a lot of leaks and concurrency issues and I only managed to solved them with this solution. 


So, for:

> If you have multiple tests running in parallel this could be a concurrency issue and cause some flakiness with shared mocks.

I think the code addresses this as-is.

> These could cause a memory leak since we aren't waiting to see if they stop. Maybe not SO important in a pipeline?

Calling the `cancel` function for the context upon test cleanup should handle this. Without it, I would get:

```
INFO 2025/10/28 11:02:30 Connection to execution failed with error: multiple Read calls return no data or error
```

With it, I get:
```
INFO 2025/10/28 11:03:18 Connection to execution failed with error: context canceled
```

pkg/workceptor/remote_work_test.go
PabloHiro: Updated the tests so all the repetitive stuff is under `createRemoteWorkNetworkSetup` now. Thank you for the call out.

PabloHiro: Please let me know if it looks more maintainable now.

pkg/workceptor/remote_work_test.go
PabloHiro: Done, thank you! It was available under `path/filepath`.

pkg/workceptor/remote_work_test.go
PabloHiro: Done, thank you!

pkg/workceptor/remote_work_test.go
PabloHiro: Let's keep the convo in:
https://github.com/ansible/receptor/pull/1458#discussion_r2468542930

pkg/workceptor/remote_work_test.go
PabloHiro: I only updated this test to match function signatures and to avoid race conditions as it was using a shared object. Let's keep it for another one :pray:

pkg/workceptor/remote_work_test.go
PabloHiro: After moving this to a helper function the linter is not complaining anymore either hehe

pkg/workceptor/remote_work_test.go
arrestle: I added AAP-56901 as a 1-point jira to linked to this one. We should talk about creating new tickets when we uncover issues of this type.

pkg/workceptor/remote_work_test.go
arrestle: The switch statement still has significant duplication. Cases like "start not started", "restart already started", "restart unknown work unit", etc. repeat 10-15 lines of identical mock setup and file operations.
This makes the test harder to maintain, any change to the mock setup pattern requires updating multiple places. Consider extracting a helper:

```go
func setupRemoteOperation(t *testing.T, ctrl *gomock.Controller, ctx context.Context,
    mockNetceptor *mock_workceptor.MockNetceptorForWorkceptor,
    mockBaseWorkUnit *mock_workceptor.MockBaseWorkUnitForWorkUnit,
    messages []string, testName string) {
    
    mockConn, _ := createRemoteWorkNetworkSetup(t, ctrl, ctx, messages)
    mockNetceptor.EXPECT().GetClientTLSConfig(gomock.Any(), gomock.Any(), gomock.Any()).Return(&tls.Config{}, nil).AnyTimes()
    mockNetceptor.EXPECT().DialContext(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(mockConn, nil).AnyTimes()
    
    tempDir := t.TempDir()
    mockBaseWorkUnit.EXPECT().Load().Return(nil).AnyTimes()
    mockBaseWorkUnit.EXPECT().StdoutFileName().Return(filepath.Join(tempDir, "stdout")).AnyTimes()
    mockBaseWorkUnit.EXPECT().UnitDir().Return(tempDir).AnyTimes()
    mockBaseWorkUnit.EXPECT().UpdateBasicStatus(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
    mockBaseWorkUnit.EXPECT().LastUpdateError().Return(nil).AnyTimes()
    
    if err := os.WriteFile(filepath.Join(tempDir, "stdin"), []byte{}, 0644); err != nil {
        t.Fatalf("Failed to create stdin: %v", err)
    }
}
```

This reduces each case to ~3 lines instead of ~25, and fixes the temp file issues we discussed.

pkg/workceptor/remote_work_test.go
arrestle: Oh I see, that's clever! Maybe add a comment to  make this clear:

```go
// Each test calls createRemoteWorkNetworkSetup, creating a unique messageIndex,
// that is captured by this closure, so parallel tests don't interfere with each other
```

pkg/workceptor/remote_work_test.go
PabloHiro: Are we looking at the same code :thinking: ? I moved everything to `createRemoteWorkNetworkSetup`. There are no duplicated lines for any cases

matoval (approved):
Great work, should we add a test to ensure monitorRemoteUnit isn't called during release?

pkg/workceptor/remote_work_test.go
PabloHiro: Soo it seems like what I did was not enough. I added a new case that triggered a leak. I had to define the context out of `createRemoteWorkTestSetup`, and define it on a per test basis. You were right. I added this comment:

https://github.com/ansible/receptor/blob/d325be39a16034126c976547f6e8642b2d8ba64c/pkg/workceptor/remote_work_test.go#L342-L343

PabloHiro:
> Great work, should we add a test to ensure monitorRemoteUnit isn't called during release?

@matoval I have added a new test case `release_already_started`. This one will pass now. But will fail with https://github.com/ansible/receptor/pull/1452, as there is no need to perform any monitoring :)

pkg/workceptor/remote_work_test.go
PabloHiro: @arrestle please review again whenever you can.

matoval:
> > Great work, should we add a test to ensure monitorRemoteUnit isn't called during release?
> 
> @matoval I have added a new test case `release_already_started`. This one will pass now. But will fail with #1452, as there is no need to perform any monitoring :)

Maybe when we merge it to #1452 we can flip the logic so if the code ever changes in the future calling monitorRemoteUnit during release this test will fail?

arrestle (approved):

PR #1460, opened 2025-10-28
[AAP-55666] Fix set minimum TLS version to 1.3 as Default

PabloHiro:
# [AAP-55666] Add Tests for MinTLS13 Default Change

**AI Attribution:** Tests developed with AI assistance (Claude/Cursor).

## What to Review

- Unit tests verify `MinVersion` is set correctly in config
- Functional tests verify both TLS 1.3 and 1.2 connections work end-to-end

## What's Not Tested

**Negotiated TLS version verification** - Functional tests don't explicitly assert which TLS version was negotiated during handshake.

**Why:** The `QuicConnectionForConn` interface doesn't expose `ConnectionState()`. Adding it would require interface changes and updating all mocks.

**Recommendation:** Create a new Jira PR to add ConnectionState() Interface and add additional diagnostics. We could capture additional data points on the TLS connections: [ConnectionState](https://pkg.go.dev/crypto/tls#ConnectionState)


matoval (approved):
Great work! Thank you for adding tests to cover this change 🙌🏽

davemulford (approved):

fincamd (approved):

PabloHiro:
@arrestle thank you for the great work, I honestly was not expecting you would reuse my branch, I don't wanna take away the credit.

PR #1461, opened 2025-10-31
Add doc explaining kubernetes.go current workflow and functionality

matoval:
The `kubernetes_workflow.md` document explains:

This document describes the Kubernetes worker in Receptor that enables **Ansible Automation Platform (AAP)** to run Ansible jobs in Kubernetes pods. It includes Mermaid diagrams of the workflow (authentication, pod creation, log streaming), documents error scenarios (API failures, pod lifecycle issues, long-running jobs), and notes how each is handled (or where behavior is unclear). The worker creates pods, streams stdin to run playbooks, captures stdout/stderr via Kubernetes log streaming or TCP, and monitors pod lifecycle to handle success, failure, and cancellation. It covers both the recommended logger streaming method (for K8s >= 1.23.14) and the legacy TCP method, including automatic reconnection logic for long-running jobs.


docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1189 @@
...
+
+- Container starts, crashes, restarts repeatedly
+
+**Current handling:**
+
+- ⚠️ **Limited detection**: Current `podRunningAndReady()` doesn't explicitly check for CrashLoopBackOff
+- ⚠️ **Timeout behavior**: Pod stuck in CrashLoopBackOff may timeout if `podPendingTimeout` is set
+- ⚠️ **Container status checks**: May be detected via container status checks during pod watch, but not explicitly handled
+
+**Impact:** May timeout waiting for ready state, or may be detected via container status checks.
```
PabloHiro: `CrashLoopBackOff` this cannot happen because the pod spec we use does not restart the pod.

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1189 @@
...
+
+**Impact:** Job fails immediately on write error. Partial logs may be present in stdout file.
+
+### Known Unknowns
+
+These scenarios either have unclear handling or require further investigation:
+
+1. **API Server Network Partitions**: What happens if network partitions between Receptor and API server during job execution? (Partial answer: Retries may help, but eventual consistency issues unclear)
+
+2. **Etcd Backend Failures**: How does API server backend failure affect in-flight jobs? (Partial answer: API calls fail, retries occur, but no explicit handling)
```
PabloHiro: I think this is covered already in the scenarios above about the Kubernetes API. If ETCD fails, the API usually throttles/timeouts, which is handled via retries. I would remove this to reduce noise.

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1189 @@
...
+
+3. **Multi-Region API Server Failover**: Behavior during API server failover events? (Unknown: Depends on `client-go` retry behavior)
```
PabloHiro: If failover happens behind DNS, and works, it should be transparent for receptor.

If it does not work, it would fall under `Kube API Domain Name Cannot Be Resolved`. I would remove this one too to reduce noise.

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1189 @@
...
+
+4. **Resource Quota Exhaustion**: What happens when namespace quota is exhausted mid-job? (Partial answer: Pod creation fails, but handling of quota exhaustion during execution unclear)
```
PabloHiro: Quotas are used at scheduling time for whatever the pod-spec has in `requests`.
Quotas are used at runtime for whatever the pod-spec has in `limits`. 

The first case falls under `Pod Cannot Be Scheduled`. The second under `Pod Is Killed`.  would remove this one too to reduce noise.

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1189 @@
...
+
+5. **Node Draining/Eviction**: What if node is drained while pod is running? (Partial answer: Pod eviction detected via terminated state, but timing of detection unclear)
```
PabloHiro: Well, this is a real problem, but we know what happens, that the pod is killed. This should fall under `Pod Is Killed`.

About `timing of detection unclear`, I don't think we have seen this anywhere, and would be a separate problem, that we do not query Kubernetes API frequently enough to detect pod state transitions.

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1189 @@
...
+
+6. **Log Stream Line Limits**: Are there limits on individual log line length? (Unknown: Depends on Kubernetes and `client-go`)
```
PabloHiro: Isn't this the same as `Log Returned by Kube API Is Too Long`?

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1189 @@
...
+
+7. **Concurrent Job Limits**: How many concurrent jobs can run? (Unknown: Depends on Kubernetes cluster capacity and Receptor configuration)
```
PabloHiro: I think this would be a separate problem too. The refactor is about making logic more robust, not making performance better too.

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1189 @@
...
+
+8. **Pod Template Mutations**: What if pod spec is mutated after creation by admission controllers or webhooks? (Unknown: Original spec used, mutations not tracked)
+
+9. **Image Pull Secrets Expiration**: What if image pull secret expires during job execution? (Unknown: Only affects initial image pull, but not investigated)
```
PabloHiro: This is not a problem. If image cannot be pulled initially, it falls under `Pod Cannot Be Scheduled`. If pod is already running, secret expiration does not play a role.

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1189 @@
...
+
+10. **Network Policy Blocking**: What if network policies block log stream access? (Unknown: Would manifest as connection errors, but not explicitly handled)
```
PabloHiro: Same as `Kube API Connection Refused`. Would remove.

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1189 @@
...
+
+## Related Documentation
+
+- [Receptor Work Submit Flow](receptor_work_submit_flow.md) - General work submission flow
+- [Add Listener Backend](AddListenerBackend.md) - Backend connection handling
+- [AAP Handbook: How a Job Runs](https://handbook.eng.ansible.com/docs/AAP/Services/Controller/how-a-job-runs) - AAP's use of Receptor for job execution
```
PabloHiro: Handbook is an internal URL, will not work for external contributors.

PabloHiro (changes requested):

PabloHiro (approved):
LGTM, but I think the diagrams 1,3,4,6 are overkill. I think most people would read the code instead.

matoval:
> LGTM, but I think the diagrams 1,3,4,6 are overkill. I think most people would read the code instead.

I agree that these diagrams are complex and hard to read but I think that demonstrates how complex and hard to read kuberenetes.go is. I think these diagrams are a good visual representation of kuberentes.go.

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1155 @@
...
+
+1. **WorkStatePending (0)**: Initial state, connecting to Kubernetes or creating pod
+2. **WorkStateRunning (1)**: Pod running, streaming data
+3. **WorkStateSucceeded (2)**: Work completed successfully. Determined by:
+   - Exit code 0, OR
+   - Exit code != 0 but termination reason is "Completed" or "Error" (indicates normal program completion with error)
+4. **WorkStateFailed (3)**: Work failed. Determined by:
+   - Exit code != 0 AND termination reason indicates interruption (OOMKilled, Evicted, etc.), OR
+   - Other errors occurred during execution (stream errors, pod failures, etc.)
+5. **WorkStateCanceled (4)**: Work was canceled, pod deleted
```
dleehr, line 808: Are work states specific to kubernetes execution or are these global work state definitions that also apply to command?

Put another way, if we wanted to add more work states to represent different stages of `Pending` in a Kubernetes execution, would we break things?

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1155 @@
...
+
+- Cannot connect to Kubernetes API server (server down, network issues, firewall)
+
+**Current handling:**
+
+- ❌ **No explicit retry**: Connection errors during `connectToKube()` are returned immediately (line 1558-1560)
+- ❌ **No retry in CreatePod()**: TODO comment on line 849 mentions adding retry logic but not implemented
+- ✅ **Retry in log stream**: `kubeLoggingConnectionHandler()` retries up to `GetKubeRetryCount()` times with backoff (line 311-324)
+- ✅ **Retry in Get pod**: When resuming, retries 5 times with 200ms delay (line 878-901)
+- ✅ **Retry in log reconnection**: Main loop retries getting pod with exponential backoff (line 384-398)
```
dleehr: If we're referencing individual line numbers, let's provide a permanent GitHub link, like 

https://github.com/ansible/receptor/blob/8b24174d290b3224ab51845830ed15728dd73635/pkg/workceptor/kubernetes.go#L311-L324

My workflow here was to visit https://github.com/ansible/receptor/blob/devel/pkg/workceptor/kubernetes.go, select the lines, then hit the <kbd>Y</kbd> key to turn that into a permalink.

Otherwise, these line numbers will be inaccurate if we change things.

davemulford (changes requested):
Changes to make:

- Light mode text.
- Change "alt" to "if" in the diagrams.

@dleehr to review the LoggingWithReconnect diagram (diagram 4)

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1155 @@
...
+
+#### Pod Is Killed
+
+**What happens:**
+
+- Pod evicted, OOMKilled, node shutdown, manual deletion
+
+**Current handling:**
+
+- ✅ **Watch detects deletion**: `podRunningAndReady()` returns `NotFound` if pod deleted (line 188-190)
```
dleehr: I think deletion is a different case. When a pod is evicted or terminated for some reason, the `Pod` still exists in the Kubernetes API (even if it not active on any node) and we can check the Kube API and read that state.

If the pod is **deleted** entirely, it no longer exists in Kubernetes. It sounds like the `podRunningAndReady()` can handle deletions, but how do we handle a pod deletion in the middle of a job?

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1155 @@
...
+
+**Impact:** Properly detected and classified. Work state depends on both exit code AND termination reason. Jobs marked as succeeded for normal completions (even with non-zero exit) but failed for interrupted executions.
+
+### Container Execution Errors
+
+#### Container Executing Work Is Killed
+
+**What happens:**
+
+- Container running the work is killed (OOMKilled, SIGKILL, container runtime issues)
```
dleehr: How is this different from `OOMKilled` in the section above? I believe OOMKilled applies at the container level and not the pod (see https://stackoverflow.com/a/74209616), so maybe it applies here (container) but not above (pod).

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1155 @@
...
+- Sidecar containers or init containers fail or are killed
+
+**Current handling:**
+
+- ⚠️ **Limited detection**: Only checks `WorkerContainerName` container (line 491-496)
+- ⚠️ **No sidecar monitoring**: Doesn't check status of other containers
+- ⚠️ **Pod phase impact**: If sidecar failures cause pod to fail, pod phase change is detected
+- ⚠️ **Init container failures**: May prevent pod from starting, detected via pod phase
+
+**Impact:** If worker container unaffected, job continues. If pod fails due to sidecar, detected via pod phase.
```
dleehr: My understanding is that receptor is only concerned with a single container named `worker`. I believe it's possible that users providing a custom pod template can specify additional containers (e.g. a sidecar). I don't remember if we include the spec for `initContainers` or not.

More broadly I don't think this is common pattern or a major concern. It's fine to document the what-if, but I suggest that we note that this handling isn't as critical as some others.

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1155 @@
...
+
+**Current handling:**
+
+- ✅ **Context check in log reading**: Detects `context.Canceled` and marks job as failed if not already succeeded (line 426-439)
+- ✅ **Context propagation**: Uses `kw.GetContext()` throughout for cancellation propagation
+- ✅ **Cancel() method**: Deletes pod when `Cancel()` is called (line 1838-1852)
+- ⚠️ **No graceful shutdown**: No attempt to wait for current operation to complete
+- ⚠️ **Pod cleanup**: Pod is deleted immediately on cancel, may interrupt running job
+
+**Impact:** Job marked as failed on context cancellation. Pod deleted, potentially losing partial output.
```
dleehr: What would cause the context to be canceled?

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1155 @@
...
+
+**Current handling:**
+
+- ✅ **Streaming approach**: Uses streaming reads (`bufio.NewReader`) not loading all logs into memory (line 420)
+- ✅ **Line-by-line processing**: Processes one line at a time (line 422)
+- ⚠️ **No size limits**: No explicit maximum log size limits
+- ⚠️ **Disk space**: Depends on available disk space for stdout file
+- ⚠️ **Memory**: Should be safe due to streaming, but very long individual lines may cause issues
+
+**Impact:** Large logs should work due to streaming, but disk space may become an issue.
```
dleehr: Do we have any specific cases where this happened?

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1155 @@
...
+    LogSuccess --> CheckLastLine{Last line<br/>has data?}
+    LogErrorComplete --> CheckLastLine
+    LogInterrupted --> Exit9([Exit - interrupted])
+    
+    CheckLastLine -->|Yes| ProcessLast[ProcessLogLine for last line]
+    ProcessLast --> WriteLast[Write last line to stdout]
+    WriteLast --> Exit10([Exit - normal completion])
+    CheckLastLine -->|No| Exit10
+    
+    CheckState2 -->|Unknown| LogUnknown[Log: Unexpected state<br/>Continue]
```
dleehr: What's the meaning of `Continue` here? When I saw this in the diagram it confused me because it reads like we're continuing execution but we immediately transition to an exit.

```suggestion
    CheckState2 -->|Unknown| LogUnknown[Log: Unexpected state]
```

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1155 @@
...
+    KubeUnit->>KubeUnit: RunWorkUsingLogger()
+
+    alt Resume existing pod
+        KubeUnit->>KubeAPI: Get pod by name
+        KubeAPI-->>KubeUnit: Existing pod
+        Note over KubeUnit: skipStdin = true
+    else Create new pod
+        KubeUnit->>KubeAPI: CreatePod()
+        KubeAPI-->>KubeUnit: Pod created
+        Note over KubeUnit: skipStdin = false
```
dleehr: Can we add a note/sentence to explain why `skipStdin` would be `true`? I think it's when we connect to an existing pod, but it didn't jump out at me in the diagram.

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1155 @@
...
+            KubeAPI->>Pod: Send EOF
+            Pod-->>KubeAPI: Stdin closed
+            KubeAPI-->>KubeUnit: Stream complete
+            
+            alt Stdin stream success
+                KubeUnit->>KubeUnit: UpdateBasicStatus(WorkStateRunning)
+            else Stdin stream error
+                KubeUnit->>KubeUnit: UpdateBasicStatus(WorkStateFailed)
+            end
+        and Stream stdout from pod
```
dleehr: Are `stdin` and `stdout` streaming at the same time? Hard to tell from the diagram if these happen sequentially or in parallel. I think they technically happen in parallel (concurrent goroutines) but I could see a reader interpreting we send stdin to completion first. \

Does `par` mean parallel?

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1155 @@
...
+flowchart TD
+    Start([KubeLoggingWithReconnect]) --> MainLoop[Main reconnection loop]
+    
+    MainLoop --> CheckStdin{stdinErr<br/>!= nil?}
+    CheckStdin -->|Yes| Exit1([Exit - stdin failed])
+    CheckStdin -->|No| GetPod[Get pod with retry<br/>backoff]
+    
+    GetPod --> PodError{Pod get<br/>error?}
+    PodError -->|Yes| RetryPod{Retries<br/>remaining?}
+    RetryPod -->|Yes| SleepPod[Sleep with<br/>exponential backoff]
```
dleehr: Some places we refer to an "exponential backoff" and some places a "fibonacci backoff". Can we be consistent here so that readers know we're talking about the same practice each time? I don't think we have two different backoffs implemented in our code, do we?

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1155 @@
...
+        else Pod Completed
+            KubeAPI-->>KubeUnit: Pod Succeeded
+            KubeUnit->>KubeUnit: Handle completion
+        end
+        
+        Note over KubeUnit: skipStdin = false
+    end
+
+    alt !skipStdin (new pod)
+        KubeUnit->>KubeAPI: Create SPDY executor (attach)
```
dleehr, line 182: Just commenting on the first mention of a SPDY executor - I think this only appears in diagrams, but it would be good to define what it is in the text.

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1155 @@
...
+
+This diagram details the sophisticated reconnection logic used in `KubeLoggingWithReconnect()` to handle log stream disconnections.
+
+```mermaid
+%%{init: {'theme':'base', 'themeVariables': { 'primaryColor':'#4a9eff', 'primaryTextColor':'#ffffff', 'primaryBorderColor':'#4a9eff', 'lineColor':'#6bb3ff', 'secondaryColor':'#1a1a1a', 'tertiaryColor':'#2a2a2a', 'background':'#000000', 'mainBkgColor':'#000000', 'secondBkgColor':'#1a1a1a', 'nodeBkg':'#1a1a1a', 'nodeBorder':'#4a9eff', 'clusterBkg':'#0a0a0a', 'clusterBorder':'#4a9eff', 'defaultLinkColor':'#6bb3ff', 'titleColor':'#ffffff', 'edgeLabelBackground':'#1a1a1a', 'actorBkg':'#000000', 'actorBorder':'#4a9eff', 'actorTextColor':'#ffffff', 'actorLineColor':'#4a9eff', 'signalColor':'#ffffff', 'signalTextColor':'#ffffff', 'labelBoxBkgColor':'#000000', 'labelBoxBorderColor':'#4a9eff', 'labelTextColor':'#ffffff', 'loopTextColor':'#ffffff', 'noteBkgColor':'#000000', 'noteTextColor':'#ffffff', 'noteBorderColor':'#ff6b35', 'activationBkgColor':'#1a1a1a', 'activationBorderColor':'#4a9eff', 'sequenceNumberColor':'#000000'}}}%%
+flowchart TD
+    Start([KubeLoggingWithReconnect]) --> MainLoop[Main reconnection loop]
```
dleehr, line 518: Is it this loop? https://github.com/ansible/receptor/blob/8b24174d290b3224ab51845830ed15728dd73635/pkg/workceptor/kubernetes.go#L376

dleehr (changes requested):
Very useful to see this. I reviewed the diagrams 4 and 5 since these document our most critical logic flow. Everything in there aligns with my understanding. I did raise some questions to clarify though in line-level comments, so please have a look at those.

Additionally I have two other questions:

1. We've done a lot of work recently to improve execution outcomes on Kubernetes. Does that work only apply to `KubeLoggingWithReconnect` or does it also apply to other code execution modes?
2. Can we document what determines which mode (tcp/non-reconnect/reconnect) will be used? I believe there are some configuration options, but if Receptor connects to a stable, supported version of OCP, what will be the default behavior?

Other feedback:

- I think it would help to organize these diagrams into individual files. Have one document with the high-level overview of Kubernetes, then link to files with the individual modes that describe the flows. That way, if I only care about `KubeLoggingWithReconnect` the other stuff is off the screen.
- Please make sure the text is visible in both light and dark modes ☯️

matoval:
> Changes to make:
> 
> * Light mode text.
> * Change "alt" to "if" in the diagrams.
> 
> @dleehr to review the LoggingWithReconnect diagram (diagram 4)

Unfortunately mermaid diagrams don't support replacing "alt" with "if".

docs/diagrams/kubernetes_workflow.md
matoval: Work states are global work state definitions that are set in workunitbase but this doc is explaining each state specifically referencing what that state means when running kube work. 

If we wanted to add more work states we could but we would need to make sure we update all the places work states are checked and also if we add a kube specific state then that work state would only be set and checked in kube work.

docs/diagrams/kubernetes_workflow.md
matoval: I agree, I'll remove the reference to line numbers unless absolutely necessary and add permalinks if needed.

docs/diagrams/kubernetes_workflow.md
matoval: I updated this doc to mention that podRunningAndReady() only checks if deleted during startup but doesn't  watch for a deletion event after startup. When a deletion happens after startup the log stream closes and retries Get() which fails then the job is marked as failed.

davemulford (approved):
I reviewed the Kube API Errors and Pod Lifecycle Sections. Those LGTM.

docs/diagrams/kubernetes_workflow.md
matoval: I updated the doc to have two distinct sections, one for pod killed and one for container killed.

docs/diagrams/kubernetes_workflow.md
matoval: Updated to include what would cause the context to be canceled.

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1135 @@
...
+    KubeUnit->>KubeUnit: Wait for both streams to complete
+    
+    alt Both streams successful
+        KubeUnit->>KubeUnit: UpdateFullStatus(WorkStateSucceeded)
+    else Error occurred
+        KubeUnit->>KubeUnit: UpdateBasicStatus(WorkStateFailed)
+    end
+```
+
+### Diagram 5: Logger Reconnection Logic
```
PabloHiro, line 512: I re-reviewed this diagram carefully against the code and the logic flow seems to represent it accurately. The logic is nightmarish though

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1135 @@
...
+    StreamErrors --> StdoutError[Stdout Stream Error:<br/>Log stream EOF/timeout]
+    StreamErrors --> NonEOFError[Non-EOF Error:<br/>Unexpected stream error]
+
+    AuthErrors --> ConfigError[Config Parse Error:<br/>Invalid kubeconfig]
+    AuthErrors --> ClusterError[InCluster Error:<br/>Not running in cluster]
+
+    TimeoutErrors --> PodPendingTimeout[Pod Pending Timeout:<br/>Pod didn't become ready]
+    TimeoutErrors --> LogStreamTimeout[4-hour Log Stream Timeout:<br/>Kubernetes API closes stream]
+
+    PodFailed --> HandlePodFailed[Handle:<br/>1. Try to get pod logs<br/>2. Update status to Failed<br/>3. Return error with details]
```
PabloHiro: Checked and the code simply does:
```suggestion
    PodFailed --> HandlePodFailed[Handle: Return error with details]
```

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1135 @@
...
+
+    PodCompleted --> HandlePodCompleted[Handle:<br/>1. Check container exit code<br/>2. If exit != 0, return error<br/>3. If exit == 0, return ErrPodCompleted]
+
+    ImagePullBack --> HandleImagePull[Handle:<br/>1. Retry check 3 times<br/>2. If still failing, return ErrImagePullBackOff]
+
+    NotFound --> HandleNotFound[Handle:<br/>1. During startup: Return NotFound error<br/>2. During execution: Retry Get with backoff<br/>3. After retries exhausted: Mark as Failed]
```
PabloHiro: Checked and the code simply does:
```suggestion
    NotFound --> HandleNotFound[Handle: Return error with details]
```

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1135 @@
...
+    
+    StdinError --> RetryStdin{Retries<br/>remaining?}
+    RetryStdin -->|Yes| RetryStdinAction[Retry with 200ms delay<br/>Max: GetKubeRetryCount times]
+    RetryStdinAction --> RetryStdin
+    RetryStdin -->|No| FailStdin[Mark work as Failed<br/>Signal stdout to stop]
```
PabloHiro: I think this block is not accurate. There is no logic for this. If the SPDY executor cannot be created we error directly:

https://github.com/ansible/receptor/blob/a62135cd92e685eab020f18a767c53bfba31a340/pkg/workceptor/kubernetes.go#L911-L916

And if Stdin reader fails, we do not retry:
https://github.com/ansible/receptor/blob/a62135cd92e685eab020f18a767c53bfba31a340/pkg/workceptor/kubernetes.go#L935-L944

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1135 @@
...
+    
+    NonEOFError --> RetryNonEOF
+    
+    ConfigError --> FailAuth[Mark as Failed:<br/>Cannot authenticate]
+    ClusterError --> FailAuth
+    
+    PodPendingTimeout --> FailTimeout[Mark as Failed:<br/>Pod didn't become ready]
+    LogStreamTimeout --> ReconnectLogs
+    
+    ReconnectLogs --> MainLoop[Continue main loop]
```
PabloHiro: I would remove this line. This diagram describes how we handle every error, it does not describe flows in the code.

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1135 @@
...
+    RetryLogRead --> ReconnectLogs
+    RetryNonEOF -->|No| FailNonEOF[Mark as Failed]
+    
+    NonEOFError --> RetryNonEOF
+    
+    ConfigError --> FailAuth[Mark as Failed:<br/>Cannot authenticate]
+    ClusterError --> FailAuth
+    
+    PodPendingTimeout --> FailTimeout[Mark as Failed:<br/>Pod didn't become ready]
+    LogStreamTimeout --> ReconnectLogs
```
PabloHiro: Would remove this line, LogStreamTimeout is not a "Timeout" error, its symptom is an extra EOF when querying the API

docs/diagrams/kubernetes_workflow.md
```diff
@@ -0,0 +1,1135 @@
...
+
+    StreamErrors --> StdinError[Stdin Stream Error:<br/>SPDY executor failure]
+    StreamErrors --> StdoutError[Stdout Stream Error:<br/>Log stream EOF/timeout]
+    StreamErrors --> NonEOFError[Non-EOF Error:<br/>Unexpected stream error]
+
+    AuthErrors --> ConfigError[Config Parse Error:<br/>Invalid kubeconfig]
+    AuthErrors --> ClusterError[InCluster Error:<br/>Not running in cluster]
+
+    TimeoutErrors --> PodPendingTimeout[Pod Pending Timeout:<br/>Pod didn't become ready]
+    TimeoutErrors --> LogStreamTimeout[4-hour Log Stream Timeout:<br/>Kubernetes API closes stream]
```
PabloHiro: I would remove this line, this is already handled under the `Error=EOF?` branch

PabloHiro (commented):
After re-reviewing the Diagram 7 carefully, I think we need to tweak them as they are misleading.

docs/diagrams/kubernetes_workflow.md
matoval: No specific cases for this issue but we have seen customers get the "no space left on device" error when enable increase logging. That being said I don't think Receptor should be checking for disk space or available memory. AAP does check memory during the capacity algorithm and can check cpu capacity if enabled.

I updated this section to add more context on this "issue".

docs/diagrams/kubernetes_workflow.md
matoval: Looking at the code we have a debug message saying we will continue but then we fail right after. I will update this diagram to represent the code and we should fix the debug message.

docs/diagrams/kubernetes_workflow.md
matoval: Yes `par` means parallel, I updated the diagram to show the different goroutines launch behavior.

docs/diagrams/kubernetes_workflow.md
matoval: I updated the doc to only refer to the backoff as `fibonacci backoff`, you're correct we only have one backoff implementation and referring to it with exponential and fibonacci is confusing.

docs/diagrams/kubernetes_workflow.md
matoval: I updated the docs to explain what SPDY executor is and does.

docs/diagrams/kubernetes_workflow.md
matoval: Correct

docs/diagrams/kubernetes_workflow.md
matoval: Update diagram 7

docs/diagrams/kubernetes_workflow.md
matoval: Removed

docs/diagrams/kubernetes_workflow.md
matoval: removed

docs/diagrams/kubernetes_workflow.md
matoval: removed

docs/diagrams/kubernetes_workflow.md
PabloHiro: @matoval, diagram 5 is not rendering now:

```
Unable to render rich display

Parse error on line 65:
...Unknown[Debug log: "Will continue"<br/>M
-----------------------^
Expecting 'SQE', 'DOUBLECIRCLEEND', 'PE', '-)', 'STADIUMEND', 'SUBROUTINEEND', 'PIPE', 'CYLINDEREND', 'DIAMOND_STOP', 'TAGEND', 'TRAPEND', 'INVTRAPEND', 'UNICODE_TEXT', 'TEXT', 'TAGSTART', got 'STR'

For more information, see https://docs.github.com/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams#creating-mermaid-diagrams
```

dleehr (changes requested):
Diagram 5 is not rendering

docs/diagrams/kubernetes_workflow.md
matoval: This issue is now fixed

matoval:
> Diagram 5 is not rendering

This issue is now fixed

dleehr (approved):
LGTM. Major feedback has been addressed. I still think it will be easier to understand if we separate the different flows (tcp, logger, logger with reconnect) into their own files, but that can be done later.

Great to get this recorded so we understand the scope and complexity of this part of the codebase.

PR #1462, opened 2025-10-31
Remove Kube retry counter decrement on EOF with job in Running status

davemulford:
This changes the behavior when a Kube worker encounters an `EOF` while a job pod is reported in the `Running` status by Kubernetes. Previously, this would decrement the retry counter which would eventually exhaust, and fail the job.

Now, an `EOF` from a `Running` job will no longer decrement the retry counter. The affect is that jobs on Kubernetes will not be failed by receptor when producing no output for long periods of time. I.e., we now treat an `EOF` as a reconnect signal instead of an error condition.

Fixes AAP-54803


pkg/workceptor/kubernetes.go
```diff
@@ -506,33 +506,17 @@ mainLoop:
...
+					kw.GetWorkceptor().nc.GetLogger().Info(
+						"Detected EOF Error: %s for pod %s/%s in with container state: Running. Job may not be complete. Will continue attempting to run job.",
+						err,
 						podNamespace,
 						podName,
 					)
 
-					return
+					time.Sleep(kw.GetSleepDuration(curContainerDelay))
+					prevContainerDelay, curContainerDelay = curContainerDelay, prevContainerDelay+curContainerDelay
```
matoval: The delay could grow to a very long time, should we add a max for the Fibonacci backoff?

matoval (commented):
Great work, I have one question about the Fibonacci backoff.

pkg/workceptor/kubernetes.go
```diff
@@ -506,33 +506,17 @@ mainLoop:
 				case containerState.Running != nil:
 					// We got EOF but pod is running, is this because we checked too fast? Will it turn into a terminated state soon or are we hitting the 4 hour log stream kube error. We will attempt to reconnect a max of 5 times in order to cover both cases
```
PabloHiro: We need to update the comment now :sweat_smile:

pkg/workceptor/kubernetes.go
davemulford: We want to constrain the `prevContainerDelay` and `curContainerDelay` so those don't overflow.

pkg/workceptor/kubernetes.go
davemulford: Will do!

dleehr (changes requested):
Nice progress. On review I have one request for the tests.

Previously we were testing that each `EOF` resulted in a retry operation and checked that the sequence of log messages showed a decrement in the number of retries left: `Will retry 2 more times`, `Will retry 1 more times`, `continuing to stream...`. This was a good test because it confirmed the behavior we intended. We should see the number go down and eventually give up.

Now, the tests just check for a single occurrence of the `will continue` log message. Does that demonstrate that we retry indefinitely and never exhaust? Can we update the tests to show that we keep retrying past the retry limit?

pkg/workceptor/kubernetes.go
davemulford: @matoval I've added a function to calculate the next Fibonacci values. Returned values are constrained between 0 and 1000.

If you like this, I will add unit tests for the `GetNextFibonacciValues()` function next.

matoval: This will work but should the max be 300 to match the 5 minute max? Does creating a var to have a single place the max is set make sense here?

davemulford: The lowest you can go and still allow for the 5-minute max sleep time is 400. In that case, the previous Fibonacci values are `prevDelay=144` and `curDelay=233`.

I'll add a constant to hold the maximum Fibonacci value.

matoval: That's a good point, fib isn't going to be exactly 300

PabloHiro (approved):
LGTM

pkg/workceptor/kubernetes_test.go
```diff
@@ -920,12 +931,21 @@ func TestKubeLoggingWithReconnect(t *testing.T) {
...
+				"Error opening log stream for pod Test_Namespace/Test_Name. Will retry 3 more times. Error: context canceled",
+				"Error opening log stream for pod Test_Namespace/Test_Name. Will retry 2 more times. Error: context canceled",
+				"Error opening log stream for pod Test_Namespace/Test_Name. Will retry 1 more times. Error: context canceled",
+				"Error opening log stream for pod Test_Namespace/Test_Name. Error: context canceled",
+			},
+			validatedCountedLogs: true,
+			countedLogMsgs: []countedLogMsg{
+				{
+					text:          "Detected EOF Error: EOF for pod Test_Namespace/Test_Name in with container state: Running. Job may not be complete. Will continue attempting to run job.",
+					expectedCount: 11, // 1 + (2sec timeout * 200ms sleep)
```
dleehr, line 947: 👍

dleehr (approved):
LGTM 🎉

PR #1484, opened 2025-11-21
Reduce cognitive complexity of ReceptorNames function

PabloHiro:
AAP-60060

## Summary
Reduces cognitive complexity of the `ReceptorNames` function in `pkg/utils/other_name.go` by extracting two helper functions:
- `decodeReceptorName`: Handles decoding of a single ASN.1 value
- `extractNamesFromExtension`: Processes a single extension

This refactoring reduces nesting levels and improves code readability while maintaining identical functionality.

## Test plan
- [x] All existing tests pass
- [x] No changes to public API
- [x] Behavior remains identical

### DIFF
--- a/pkg/utils/other_name.go
+++ b/pkg/utils/other_name.go
@@ -20,28 +20,28 @@ type OtherNameDecode struct {
 }
 
 // decodeReceptorName decodes a single ASN.1 value and extracts the Receptor name if present.
-func decodeReceptorName(value asn1.RawValue) (string, bool, error) {
+func decodeReceptorName(value asn1.RawValue) (string, error) {
 	if value.Tag != 0 {
-		return "", false, nil
+		return "", nil
 	}
 
 	on := OtherNameDecode{}
 	_, err := asn1.UnmarshalWithParams(value.FullBytes, &on, "tag:0")
 	if err != nil {
-		return "", false, err
+		return "", err
 	}
 
 	if !on.ID.Equal(OIDReceptorName) {
-		return "", false, nil
+		return "", nil
 	}
 
 	var name string
 	_, err = asn1.Unmarshal(on.Value.Bytes, &name)
 	if err != nil {
-		return "", false, err
+		return "", err
 	}
 
-	return name, true, nil
+	return name, nil
 }
 
 // extractNamesFromExtension extracts Receptor names from a single extension.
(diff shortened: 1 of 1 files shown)


pkg/utils/other_name.go
```diff
@@ -19,34 +19,66 @@ type OtherNameDecode struct {
 	Value asn1.RawValue
 }
 
+// decodeReceptorName decodes a single ASN.1 value and extracts the Receptor name if present.
+func decodeReceptorName(value asn1.RawValue) (string, bool, error) {
```
dleehr: 🤔 Only thought here is that we can probably revise this to return `(string, error)` instead of `(string, bool, error)`. It's a style choice and not really worth debate. The only reason I mention it is that the purpose of the change is already to reduce cognitive complexity, but SonarQube isn't complaining (nor is our linter) so I suppose I have my decision 😁.

When returning multiple values I prefer to stick with as few as possible (two) unless we need more. Three is still reasonable so I don't object. Plus there's value in the `bool` giving us an explicit value for `found` later, so I agree with this.

dleehr (approved):
LGTM

pkg/utils/other_name.go
fincamd: I think we can turn this into a two-value return function. From what I'm reading, in the refactored function `decodeReceptorName`, we have four return statements that return a 3-tuple. The first 3 return an empty string, false and error object or nil. In these cases, we could calculate the found bool value by asking if the length of the name string is > 0 so the logic stays intact.

Now the fourth and last one is the one that will determine if we can change this into a 2-value return function. Is there any way @PabloHiro that this code returns an empty string (without an error)?
```go
	var name string
	_, err = asn1.Unmarshal(on.Value.Bytes, &name)
```
If this code will never return an empty string, then we can safely remove the found bool value and calculate the length of the returned string to tell whether a name was found in the `extractNamesFromExtension` function.

Plus, this is the pattern most functions follow: returning (1) a value of X type and (2) an error object or nil.

Hence, we could change this code to something like this:

```go
	names := make([]string, 0)
	for _, value := range values {
		name, err := decodeReceptorName(value)
		if err != nil {
			return nil, err
		}
		if len(name) > 0 {
			names = append(names, name)
		} else {
            // Should we worry about this scenario? I don't have context on what this function does
        }
	}
```

fincamd (changes requested):

pkg/utils/other_name.go
PabloHiro: Implemented your change Dani :+1:

fincamd (approved):
marvelous, LGTM

PR #1485, opened 2025-11-27
Remove usage of os.TempDir

PabloHiro:
AAP-59224

Fix for the following Sonarqube issue:
https://sonarcloud.io/project/issues?open=AZWFYq7JnpgT4YCYDYyc&id=ansible_receptor

### DIFF
--- a/pkg/types/main_test.go
+++ b/pkg/types/main_test.go
@@ -29,7 +29,8 @@ func TestMainInitNodeID(t *testing.T) {
 	for _, testCase := range mainInitNodeIDTestCases {
 		t.Run(testCase.name, func(t *testing.T) {
 			cfg := NodeCfg{
-				ID: testCase.nodeID,
+				ID:      testCase.nodeID,
+				DataDir: "/tmp/test-receptor",
 			}
 			err := cfg.Init()
 			if err == nil && testCase.expectedErr != "" {
@@ -58,7 +59,8 @@ func TestNodeCfgInitEmptyID(t *testing.T) {
 
 func TestNodeCfgInitReservedID(t *testing.T) {
 	cfg := NodeCfg{
-		ID: "localhost",
+		ID:      "localhost",
+		DataDir: "/tmp/test-receptor",
 	}
 
 	err := cfg.Init()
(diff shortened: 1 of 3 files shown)


fincamd (approved):
LGTM

davemulford (approved):
LGTM

davemulford (changes requested):
Adding request changes so we have time to make sure the two places where `workceptor.New()` is called are giving the expected value for the `baseDir` parameter.

PabloHiro:
> Adding request changes so we have time to make sure the two places where `workceptor.New()` is called are giving the expected value for the `baseDir` parameter.

In the code (not tests), this is called only in:

https://github.com/ansible/receptor/blob/561d9e579cecfebf9931616d6c0b833230d5eca9/pkg/types/main.go#L81

It uses `receptorDataDir`, which comes from `cfg.DataDir`:

https://github.com/ansible/receptor/blob/561d9e579cecfebf9931616d6c0b833230d5eca9/pkg/types/main.go#L56

It has a default value of `/tmp/receptor` hinted in:

https://github.com/ansible/receptor/blob/561d9e579cecfebf9931616d6c0b833230d5eca9/pkg/types/main.go#L22

The actual default is loaded here:

https://github.com/ansible/receptor/blob/561d9e579cecfebf9931616d6c0b833230d5eca9/cmd/defaults.go#L105

Not sure where the second call to `New()` is?

davemulford:
> Not sure where the second call to `New()` is?

The second place we were looking at is here:

https://github.com/ansible/receptor/blob/561d9e579cecfebf9931616d6c0b833230d5eca9/tests/functional/mesh/lib.go#L608

I'm just now noticing that's in a functional test.

davemulford:
We want to make sure that `/tmp/receptor` is the default data directory value in both config v1 and v2. @matoval is checking into this.

matoval (approved):
This default is set but only when using config V2: https://github.com/ansible/receptor/blob/561d9e579cecfebf9931616d6c0b833230d5eca9/cmd/defaults.go#L105

The default is set for config v1 here: https://github.com/ansible/receptor/blob/561d9e579cecfebf9931616d6c0b833230d5eca9/pkg/types/main.go#L22

The comment `default:"/tmp/receptor"` is used to set the default.


LGTM

davemulford (approved):

PR #1487, opened 2025-11-27
Fix SonarQube issues: Remove unnecessary else clauses

PabloHiro:
AAP-60060

## Summary
- Removes redundant else clauses when the if block contains a return statement
- Improves code readability and addresses SonarQube code quality issues
- Affects 6 files across the codebase with consistent pattern fixes

## Changes
- `pkg/netceptor/conn.go`: Remove else clause in tracer function
- `pkg/types/main.go`: Remove else clause in ReceptorPyroscopeCfg.Init function  
- `internal/version/version.go`: Remove else clause in validateVersion function
- `pkg/workceptor/kubernetes.go`: Remove else clause in RunWorkUsingLogger function
- `pkg/workceptor/kubernetes_test.go`: Remove else clauses in test functions (2 instances)
- `pkg/workceptor/remote_work_test.go`: Remove else clause in test helper function

### DIFF
--- a/pkg/types/main.go
+++ b/pkg/types/main.go
@@ -153,12 +153,11 @@ func (pyroscopeCfg ReceptorPyroscopeCfg) Init() error {
 		DisableGCRuns:     pyroscopeCfg.DisableGCRuns,
 		HTTPHeaders:       pyroscopeCfg.HTTPHeaders,
 	})
-
 	if err != nil {
 		return err
-	} else {
-		return nil
 	}
+
+	return nil
 }
 
 func getUploadRate(cfg ReceptorPyroscopeCfg) time.Duration {
--- a/pkg/types/main.go
+++ b/pkg/types/main.go
@@ -153,11 +153,8 @@ func (pyroscopeCfg ReceptorPyroscopeCfg) Init() error {
 		DisableGCRuns:     pyroscopeCfg.DisableGCRuns,
 		HTTPHeaders:       pyroscopeCfg.HTTPHeaders,
 	})
-	if err != nil {
-		return err
-	}
 
-	return nil
+	return err
 }
 
 func getUploadRate(cfg ReceptorPyroscopeCfg) time.Duration {
--- a/pkg/workceptor/kubernetes.go
+++ b/pkg/workceptor/kubernetes.go
@@ -1072,13 +1072,12 @@ func (kw *KubeUnit) RunWorkUsingLogger() {
 					prevContainerDelay, curContainerDelay = GetNextFibonacciValues(prevContainerDelay, curContainerDelay)
 
 					continue podLoop
-				} else {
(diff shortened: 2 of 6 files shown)


pkg/types/main.go
```diff
@@ -153,12 +153,11 @@ func (pyroscopeCfg ReceptorPyroscopeCfg) Init() error {
...
 		HTTPHeaders:       pyroscopeCfg.HTTPHeaders,
 	})
-
 	if err != nil {
 		return err
-	} else {
-		return nil
 	}
+
+	return nil
```
fincamd: Hmm, why not just return the `err` object? We don't need this if else or the if, then return nil

fincamd (changes requested):
Minor nitpick

pkg/types/main.go
davemulford: As @fincamd suggests let's simply return `err` as that catches both cases, where an error exists or doesn't.

```suggestion

    return err
```

pkg/workceptor/kubernetes.go
```diff
@@ -1072,13 +1072,12 @@ func (kw *KubeUnit) RunWorkUsingLogger() {
...
-					kw.GetWorkceptor().nc.GetLogger().Error("%s", errMsg)
-					kw.UpdateBasicStatus(WorkStateFailed, errMsg, 0)
-
-					return
 				}
+				errMsg := fmt.Sprintf("Container in %s pod is not running container state unknown, retries exhausted", podName)
+				kw.GetWorkceptor().nc.GetLogger().Error("%s", errMsg)
+				kw.UpdateBasicStatus(WorkStateFailed, errMsg, 0)
+
+				return
```
davemulford: During PR review, the team decided that keeping this `else` statement makes the code more readable. Instead of a code change, we should be able to tell Sonar that we're okay with this as-is and to not warn about it in the future.

davemulford (changes requested):
Overall looks good. However, a couple of requests: 1 for the suggested change to simply `return err` instead of an if-statement; the other to NOT remove the `else` block from kubernetes.go due to code complexity.

pkg/workceptor/kubernetes.go
PabloHiro: Marked the Sonar issue as "Accepted" here:
https://sonarcloud.io/project/issues?impactSoftwareQualities=RELIABILITY&issueStatuses=OPEN%2CCONFIRMED&id=ansible_receptor&open=AZq8AxD4ytKuOluVLf5X

pkg/types/main.go
PabloHiro: Done!

fincamd (approved):

davemulford (approved):
LGTM

dleehr (approved):

PR #1489, opened 2025-11-28
Fix SonarQube issue godre:S8166: Replace t.Fatal with t.Error in goroutines

PabloHiro:
AAP-60060

## Summary
Fixes 4 SonarQube issues with rule `godre:S8166` by replacing `t.Fatal`/`t.Fatalf` with `t.Error`/`t.Errorf` when called from goroutines in test files.

## Issue
SonarQube rule `godre:S8166` flags when `t.Fatal` or `t.Fatalf` is called from a goroutine, as this causes `runtime.Goexit()` to be called which doesn't properly terminate the test. Instead, `t.Error` or `t.Errorf` should be used in goroutines.

## Changes
- **pkg/workceptor/lock_test.go**: Fixed 3 instances (lines 65, 69, 73)
  - Line 65: `t.Fatalf` → `t.Errorf` 
  - Line 69: `t.Fatalf` → `t.Errorf`
  - Line 73: `t.Fatal` → `t.Error`
- **pkg/backends/websocket_interop_test.go**: Fixed 1 instance (line 93)
  - Line 93: `t.Fatalf` → `t.Errorf`

## SonarQube Issues Resolved
- [AZq8AxGXytKuOluVLf6N](https://sonarcloud.io/project/issues?open=AZq8AxGXytKuOluVLf6N&id=ansible_receptor) - pkg/workceptor/lock_test.go:65
- [AZq8AxGXytKuOluVLf6O](https://sonarcloud.io/project/issues?open=AZq8AxGXytKuOluVLf6O&id=ansible_receptor) - pkg/workceptor/lock_test.go:69
- [AZq8AxGXytKuOluVLf6P](https://sonarcloud.io/project/issues?open=AZq8AxGXytKuOluVLf6P&id=ansible_receptor) - pkg/workceptor/lock_test.go:73
- [AZq8AxLxytKuOluVLf7B](https://sonarcloud.io/project/issues?open=AZq8AxLxytKuOluVLf7B&id=ansible_receptor) - pkg/backends/websocket_interop_test.go:93

### DIFF
--- a/pkg/workceptor/lock_test.go
+++ b/pkg/workceptor/lock_test.go
@@ -5,6 +5,7 @@ package workceptor
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path"
@@ -33,6 +34,7 @@ func TestStatusFileLock(t *testing.T) {
 		waitTime := time.Duration(i) * baseWaitTime
 		totalWaitTime += waitTime
 		go func(iter int, waitTime time.Duration) {
+			defer wg.Done()
 			sfd := StatusFileData{}
 			sfd.UpdateFullStatus(statusFilename, func(status *StatusFileData) {
 				time.Sleep(waitTime)
@@ -40,20 +42,19 @@ func TestStatusFileLock(t *testing.T) {
 				status.StdoutSize = int64(iter)
 				status.Detail = fmt.Sprintf("%d", iter)
 			})
-			wg.Done()
 		}(i, waitTime)
 	}
 	ctx, cancel := context.WithCancel(context.Background())
 	wg2 := sync.WaitGroup{}
 	wg2.Add(numReaderThreads)
+	errChan := make(chan error, numReaderThreads)
 	for i := 0; i < numReaderThreads; i++ {
 		go func() {
+			defer wg2.Done()
 			sfd := StatusFileData{}
 			fileHasExisted := false
 			for {
 				if ctx.Err() != nil {
-					wg2.Done()
-
 					return
 				}
 				err := sfd.Load(statusFilename)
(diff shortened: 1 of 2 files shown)


fincamd (dismissed):
LGTM

pkg/workceptor/lock_test.go
```diff
@@ -62,15 +62,15 @@ func TestStatusFileLock(t *testing.T) {
...
 				}
 				detailIter, err := strconv.Atoi(sfd.Detail)
 				if err != nil {
-					t.Fatalf("Error converting status detail to int: %s", err)
+					t.Errorf("Error converting status detail to int: %s", err)
 				}
 				if detailIter >= 0 {
 					if int64(sfd.State) != sfd.StdoutSize || sfd.State != detailIter {
-						t.Fatal("Mismatched data in struct")
+						t.Error("Mismatched data in struct")
```
davemulford: Since `t.Fatal()` exits only the go routine, and not the test, instead of calling either `t.Fatal()` or `t.Error()` can the go func return an `error` value. Then, the calling function check for errors from any of the go routines, and if any exist then call `t.Fatal()`. Also, in any error condition do not wait on any WaitGroup, simply fail the test.

This prevents the `wg2.Wait()` from potentially waiting forever.

davemulford (changes requested):
Looks good! Some thoughts about handling multiple go routines, waiting, and error conditions.

pkg/workceptor/lock_test.go
PabloHiro: Updated! After iterating with Claude for a while, I made the following changes:

- `wg2.Done()` is deferred at the beginning of the go routines.
- Errors are sent to a channel. Since we are dealing with several go routines at the same time, a channel is needed to potentially store an error for every go routine.
- `cancel()` is called to cancel the context early if an error is found in any go routine.
- I have also learnt that it is best practice to use `%w` instead of `%s` when returning errors. See: https://stackoverflow.com/questions/61283248/format-errors-in-go-s-v-or-w

fincamd (approved):

davemulford (approved):
LGTM

PR #1493, opened 2025-12-01
Fix unhandled errors in test files

PabloHiro:
AAP-60060

Fixes unhandled error returns in test files across the codebase.

**Changes:**
- Handle PEM decode rest bytes and hash computation validity flags in TLS config tests
- Fix big.Int parsing to use integer string instead of decimal
- Fix variable shadowing in adapter test goroutine
- Handle context cancellation and other error returns in various test files


davemulford (approved):
LGTM

matoval (approved):

PR #1495, opened 2025-12-03
[AAP-60058] Tests for tcp.go

PabloHiro:
AAP-60058

* Refactored existing tests `TestTCPListenerStart` and `TestTCPDialerStart` so they use their corresponding constructor, `NewTCPListener` and `NewTCPDialer` .
* For `TCPDialer` added tests for `GetAddr` and `GetTLS`.
* For TCPListener added tests for `GetAddr`, `GetTLS`.
  * Removed `GetCost`, copy pasted unused code.
* For `TCPSession` added tests for `newTCPSession`, `Send`, `Recv`, `Close`.
* For `TCPListenerCfg` added tests for `GetCost`, `GetNodeCost`, `GetAddr`, `GetTLS`, `Prepare`.
* For `TCPDialerCfg` added tests for `Prepare`.

Tried using LLMs for this one, but ended up having to rewrite most of it manually, so not generated or co-authored by LLMs :grin:

### DIFF
--- a/pkg/backends/tcp.go
+++ b/pkg/backends/tcp.go
@@ -96,10 +96,6 @@ func (b *TCPListener) GetAddr() string {
 	return b.li.Addr().String()
 }
 
-func (b *TCPListener) GetCost() string {
-	return b.li.Addr().String()
-}
-
 func (b *TCPListener) GetTLS() *tls.Config {
 	return b.TLS
 }
--- a/pkg/backends/tcp_test.go
+++ b/pkg/backends/tcp_test.go
@@ -7,10 +7,121 @@ import (
 	"reflect"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/ansible/receptor/pkg/logger"
-	"github.com/ansible/receptor/pkg/netceptor"
 )
+
+func TestTCPDialer_GetAddr(t *testing.T) {
+	tests := []struct {
+		name    string
+		address string
+	}{
+		{
+			name:    "IPv4 address",
+			address: "127.0.0.1:8080",
+		},
+		{
+			name:    "IPv6 address",
+			address: "[::1]:8080",
+		},
+		{
+			name:    "Hostname",
+			address: "example.com:443",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			b, err := NewTCPDialer(tt.address, true, nil, nil)
+			if err != nil {
+				t.Errorf("NewTCPDialer() error = %v", err)
+
+				return
+			}
+			if got := b.GetAddr(); got != tt.address {
+				t.Errorf("TCPDialer.GetAddr() = %v, want %v", got, tt.address)
+			}
+		})
+	}
+}
+
+func TestTCPDialer_GetTLS(t *testing.T) {
+	tlsConfig := &tls.Config{
+		MinVersion: tls.VersionTLS12,
+	}
+
+	tests := []struct {
+		name string
+		tls  *tls.Config
+	}{
+		{
+			name: "No TLS",
+			tls:  nil,
+		},
+		{
+			name: "With TLS config",
+			tls:  tlsConfig,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			b, err := NewTCPDialer("127.0.0.1:8080", true, tt.tls, nil)
+			if err != nil {
+				t.Errorf("NewTCPDialer() error = %v", err)
+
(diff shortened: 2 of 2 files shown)


pkg/backends/tcp_test.go
```diff
@@ -7,11 +7,122 @@ import (
...
+	tests := []struct {
+		name    string
+		address string
+		redial  bool
+		tls     *tls.Config
+		wantErr bool
+	}{
+		{
+			name:    "Successful start",
+			address: "127.0.0.1:9998",
```
arrestle:
```suggestion
			address: "127.0.0.1:0",
```
this could fail intermittently, if you specify 0 you let the OS choose a port, which is more resilient.

pkg/backends/tcp_test.go
```diff
@@ -56,112 +167,658 @@ func TestNewTCPListener(t *testing.T) {
...
-				li:      nil,
-				innerLi: nil,
-				logger:  logger.NewReceptorLogger("TCPtest"),
-			},
-			args: args{
-				ctx: context.Background(),
-				wg:  &sync.WaitGroup{},
-			},
+			name:    "Successful start",
+			address: "127.0.0.1:9998",
```
arrestle:
```suggestion
			address: "127.0.0.1:0",
```
Same thing as before.

pkg/backends/tcp_test.go
```diff
@@ -7,11 +7,122 @@ import (
 	"reflect"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/ansible/receptor/pkg/logger"
-	"github.com/ansible/receptor/pkg/netceptor"
 )
 
+func TestTCPDialer_GetAddr(t *testing.T) {
```
arrestle, line 15: Consider an easy win to validate Preload and increase code coverage.

```suggestion
func TestTCPDialerCfg_PreReload(t *testing.T) {
    // Valid config - should pass
    cfg := TCPDialerCfg{Cost: 1.0}
    if err := cfg.PreReload(); err != nil {
        t.Errorf("PreReload() error = %v", err)
    }
    
    // Invalid config - should fail (proves it calls Prepare)
    cfg.Cost = -1.0
    if err := cfg.PreReload(); err == nil {
        t.Error("PreReload() expected error for negative cost")
    }
}

func TestTCPListenerCfg_PreReload(t *testing.T) {
    cfg := TCPListenerCfg{Cost: 1.0}
    if err := cfg.PreReload(); err != nil {
        t.Errorf("PreReload() error = %v", err)
    }
    
    cfg.Cost = 0.0
    if err := cfg.PreReload(); err == nil {
        t.Error("PreReload() expected error for zero cost")
    }
}

func TestTCPDialer_GetAddr(t *testing.T) {
```

arrestle (changes requested):
Hi Pablo,
Awesome work getting up to 61.8% - I added a couple of suggestions to cover Preload , but we can save the rest for another PR since we're already way over the 200 line guideline.

pkg/backends/tcp_test.go
PabloHiro: `PreReload` is wrapping `Prepare`, and I wrote tests for `Prepare`. That is why I skipped it :thinking:

pkg/backends/tcp.go
```diff
@@ -96,10 +96,6 @@ func (b *TCPListener) GetAddr() string {
 	return b.li.Addr().String()
 }
 
-func (b *TCPListener) GetCost() string {
-	return b.li.Addr().String()
-}
-
```
matoval, line 102: I don't think we should remove GetCost but fix it to get the cost

something like:
```
func (b *TCPListener) GetCost() float64 {
	return cfg.Cost
}
```

matoval (changes requested):
I don't think we should be deleting GetCost

pkg/backends/tcp.go
PabloHiro: It is a function that is not used anywhere in the code. I'd say we should not have code that is not used.

arrestle: tough one, but I would agree with Pablo on this, since this is a testing ticket. Maybe put a ticket in the backlog to determine best practices about determining and returning connection cost.

arrestle (approved):
Resolve the cost situation and otherwise LGTM.

pkg/backends/tcp.go
matoval: @PabloHiro You're correct, the GetCost that is actually used is TCPListenerCfg GetCost. We can delete this.

matoval (approved):

PR #1496, opened 2025-12-03
[AAP-60033]  Modify receptor container build to support rootless pods

arrestle:
## Summary

Fix container build failures for users running rootless Podman with high UIDs (e.g., corporate LDAP/AD users), and document container build requirements.

## Problem

When building the container with `make container` using rootless Podman, the build fails at `ADD source.tar.gz /source` with:

```
error setting ownership of "/" to <high-uid>:<high-uid>: lchown /: invalid argument
./.git/objects/pack/pack-*.pack: write bulk-writer: broken pipe
```

**Cause:** The tarball stores the host user's UID/GID and includes the `.git` directory, both of which cause issues in rootless Podman's user namespace.

## Solution

Replace `tar` with `git archive` in the Makefile. This fixes both issues because `git archive`:
- Creates archives with normalized ownership (no UID/GID issues)
- Automatically excludes the `.git` directory
- note: Remind developer that only committed files are included by git archive

## Documentation

Updated `CONTRIBUTING.md` to document container build requirements:
- Added Python 3 and Podman/Docker to development requirements
- Added instructions for building container images (`pip install build` + `make container`)

### Build completes successfully

```
[2/2] COMMIT quay.io/ansible/receptor:v1.6.2-gitf179c54
Successfully tagged quay.io/ansible/receptor:v1.6.2-gitf179c54
```

### Version check

```
$ podman run --rm quay.io/ansible/receptor:v1.6.2-gitf179c54 receptor --version
1.6.2+gitf179c54
```

### Service starts correctly

```
$ podman run --rm -p 7323:7323 quay.io/ansible/receptor:v1.6.2-gitf179c54
INFO 2025/12/03 17:52:26 Running control service control {"node_id":"4880ef8c6f19"}
INFO 2025/12/03 17:52:26 Initialization complete
```

### Binary and config present

```
$ podman run --rm quay.io/ansible/receptor:v1.6.2-gitf179c54 ls -la /usr/bin/receptor
-rwxr-xr-x. 1 root root 73164649 Dec  3 17:49 /usr/bin/receptor

$ podman run --rm quay.io/ansible/receptor:v1.6.2-gitf179c54 cat /etc/receptor/receptor.conf
---
- control-service:
    service: control
    filename: /tmp/receptor.sock

- tcp-listener:
    port: 7323
```

### CLI tools functional

```
$ receptor --help
Usage: receptor [--<action> [<param>=<value> ...] ...]
   --help: Show this help
   --config <filename>: Load additional config options from a YAML file
   ...

$ receptorctl --help
Usage: receptorctl [OPTIONS] COMMAND [ARGS]...
Commands:
  connect     Connect the local terminal to a Receptor service on a...
  ping        Ping a Receptor node.
  reload      Reload receptor configuration.
  status      Show the status of the Receptor network.
  traceroute  Do a traceroute to a Receptor node.
  version     Show version information for receptorctl and the receptor node
  work        Commands related to unit-of-work processing
```

---

*This PR was created with assistance from Claude (Anthropic).*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -232,7 +232,8 @@ space := $(subst ,, )
 CONTAINER_FLAG_FILE = .container-flag-$(VERSION)$(subst $(space),,$(subst /,,$(EXTRA_OPTS)))
 container: $(CONTAINER_FLAG_FILE)
 $(CONTAINER_FLAG_FILE): $(RECEPTORCTL_WHEEL) $(RECEPTOR_PYTHON_WORKER_WHEEL)
-	@git archive --format tar.gz HEAD > packaging/container/source.tar.gz .
+    # Developer Note: only committed files are included with git archive.
+	@git archive --format tar.gz HEAD > packaging/container/source.tar.gz
 	@cp $(RECEPTORCTL_WHEEL) packaging/container
 	@cp $(RECEPTOR_PYTHON_WORKER_WHEEL) packaging/container
 	$(CONTAINERCMD) build $(EXTRA_OPTS) packaging/container --build-arg VERSION=$(VERSION:v%=%) -t $(REPO):$(TAG) $(if $(LATEST),-t $(REPO):latest,)
--- a/Makefile
+++ b/Makefile
@@ -232,7 +232,7 @@ space := $(subst ,, )
 CONTAINER_FLAG_FILE = .container-flag-$(VERSION)$(subst $(space),,$(subst /,,$(EXTRA_OPTS)))
 container: $(CONTAINER_FLAG_FILE)
 $(CONTAINER_FLAG_FILE): $(RECEPTORCTL_WHEEL) $(RECEPTOR_PYTHON_WORKER_WHEEL)
-	@tar --exclude-vcs-ignores -czf packaging/container/source.tar.gz .
+	@tar --exclude-vcs --exclude-vcs-ignores --owner=0 --group=0 -czf packaging/container/source.tar.gz .
 	@cp $(RECEPTORCTL_WHEEL) packaging/container
 	@cp $(RECEPTOR_PYTHON_WORKER_WHEEL) packaging/container
 	$(CONTAINERCMD) build $(EXTRA_OPTS) packaging/container --build-arg VERSION=$(VERSION:v%=%) -t $(REPO):$(TAG) $(if $(LATEST),-t $(REPO):latest,)
--- a/Makefile
+++ b/Makefile
@@ -232,7 +232,7 @@ space := $(subst ,, )
 CONTAINER_FLAG_FILE = .container-flag-$(VERSION)$(subst $(space),,$(subst /,,$(EXTRA_OPTS)))
 container: $(CONTAINER_FLAG_FILE)
 $(CONTAINER_FLAG_FILE): $(RECEPTORCTL_WHEEL) $(RECEPTOR_PYTHON_WORKER_WHEEL)
-	@tar --exclude-vcs --exclude-vcs-ignores --owner=0 --group=0 -czf packaging/container/source.tar.gz .
+	@git archive --format tar.gz HEAD > packaging/container/source.tar.gz .
 	@cp $(RECEPTORCTL_WHEEL) packaging/container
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -232,7 +232,7 @@ space := $(subst ,, )
 CONTAINER_FLAG_FILE = .container-flag-$(VERSION)$(subst $(space),,$(subst /,,$(EXTRA_OPTS)))
 container: $(CONTAINER_FLAG_FILE)
 $(CONTAINER_FLAG_FILE): $(RECEPTORCTL_WHEEL) $(RECEPTOR_PYTHON_WORKER_WHEEL)
-	@tar --exclude-vcs-ignores -czf packaging/container/source.tar.gz .
+	@tar --exclude-vcs --exclude-vcs-ignores --owner=0 --group=0 -czf packaging/container/source.tar.gz .
```
dleehr: `make container` fails for me on macOS at the `tar` step because my copy of `tar` doesn't support `--exclude-vcs-ignores`. To be clear, this also fails on `devel`. It's not a regression introduced by this change, but I still can't (easily) test it. So I think if we're going to update the Makefile here, it would be good to make it work on multiple platforms.

```
$ make container
tar: Option --exclude-vcs-ignores is not supported
Usage:
  List:    tar -tf <archive-filename>
  Extract: tar -xf <archive-filename>
  Create:  tar -cf <archive-filename> [filenames...]
  Help:    tar --help
make: *** [.container-flag-v1.6.2+gitf179c54] Error 1
```

The reason is that my `tar` binary on macOS is actually [a version of BSD tar](https://man.freebsd.org/cgi/man.cgi?bsdtar(1)) and not [GNU tar](https://www.gnu.org/software/tar/) that you get on Linux. For the basic stuff (archiving/unarchiving) they're usually interchangable, but in this case, the version of `tar` on macOS doesn't support this arg and fails. As an OS-provided tool it's not something I can easily upgrade or replace.

I also have GNU tar, which works and is the `gtar` command. So one option we could pursue is to add logic to the Makefile to use `gtar` on macOS and fall back to `tar` otherwise. But `gtar` isn't installed by default so we should document that for macOS users.

Instead, I propose to use `git archive` to produce the archive instead of `tar`. Git has built-in support for making tar (or zip) archives from code and I think this aligns well with our use case. `git` already knows to exclude the `.git` directory and the `.gitignore`d files because it's git! That's basically the whole point of `git archive` - just produce an archive of the code and strip out everything else. It's a one line change (see below) and we already require `git` in the Makefile, so I don't see much downside. Can you see if this works on your podman environment?

```suggestion
	@git archive --format tar.gz HEAD > packaging/container/source.tar.gz .
```

I did check the ownership/permissions on the TAR produced by `git archive` and it looks OK to me:

```
$ git archive --format tar.gz HEAD | tar tvf - | grep netceptor\.go 
-rw-rw-r--  0 root   root    54747 Dec  2 14:28 pkg/netceptor/netceptor.go
```


I'm also not sure why we create an archive of `.` and then extract that to the container build context, but that's even further out of scope here 😁

arrestle: Let me try that commit suggestion on my system, that looks good.

To be clear I added` --exclude-vcs` and not `--exclude-vcs-ignores` however, it's a good bet that neither is supported by macos so not sure it matters. 

That said, can we identify that this PR addresses only specific environments and (if it's needed) address the mac os issue in another ticket?

dleehr: > To be clear I added` --exclude-vcs` and not `--exclude-vcs-ignores` however, it's a good bet that neither is supported by macos so not sure it matters.

Yes that was the [first thing I noted](https://github.com/ansible/receptor/pull/1496#discussion_r2586466768):

> To be clear, this also fails on devel. It's not a regression introduced by this change, but I still can't (easily) test it. 

> That said, can we identify that this PR addresses only specific environments and (if it's needed) address the mac os issue in another ticket?

It looks like you accepted the suggestion to use `git archive`, which makes it work on macOS. So if it's also working for you on Linux, I don't think there's anything left to do.

dleehr (approved):

PabloHiro (approved):

PR #1500, opened 2025-12-05
Test Coverage for pkg/netceptor/firewall_rules.go

melissalkelly:
AAP-57038

# Summary
Add comprehensive unit tests for firewall rule parsing functions to improve code coverage and ensure robust error handling.

# Changes
### 1. Added TestParseFirewallRules Function
- Edge case testing for ParseFirewallRules()
- Coverage: 100%

### 2. Added TestRegexCompare Function
- Edge case testing for regex pattern matching
- Coverage: 93.8%

### 3. Added TestStringCompare Function
- Edge case testing for exact string matching
- Coverage: 100%

### 4. Fix typo in pkg/netceptor/firewall_rules.go

# Files Modified 
- pkg/netceptor/firewall_rules_test.go
- pkg/netceptor/firewall_rules.go

### DIFF
--- a/pkg/netceptor/firewall_rules_test.go
+++ b/pkg/netceptor/firewall_rules_test.go
@@ -429,11 +429,11 @@ func TestRegexCompare(t *testing.T) {
 
 func TestStringCompare(t *testing.T){
 	tests := []struct {
-		name string
-		field string
-		value string
-		wantErr bool
-		testCases []struct {
+		name 		string
+		field 		string
+		value 		string
+		wantErr 	bool
+		testCases 	[]struct {
 			desc string
 			data *MessageData
 			wantMatch bool
--- a/pkg/netceptor/firewall_rules_test.go
+++ b/pkg/netceptor/firewall_rules_test.go
@@ -225,19 +225,19 @@ func TestParseFirewallRules(t *testing.T) {
 			if (err != nil) != tt.wantErr {
 				t.Fatalf("ParseFirewallRules() error = %v, wantErr %v", err, tt.wantErr)
 			}
-			
+
 			// Check error message
 			if tt.wantErrContains != "" && err != nil {
 				if !strings.Contains(err.Error(), tt.wantErrContains) {
 					t.Errorf("ParseFirewallRules() error = %v, want error containing %q", err, tt.wantErrContains)
 				}
 			}
-			
+
 			// Check number of rules
 			if len(rules) != tt.wantCount {
 				t.Errorf("ParseFirewallRules() returned %d rules, want %d", len(rules), tt.wantCount)
 			}
-			
+
 			// Run test cases
 			for _, tc := range tt.testCases {
 				got := rules[tc.ruleIndex](tc.data)
@@ -251,11 +251,11 @@ func TestParseFirewallRules(t *testing.T) {
 
 func TestRegexCompare(t *testing.T) {
 	tests := []struct {
-		name            string
-		field           string
-		value           string
-		wantErr         bool
-		testCases       []struct {
+		name      string
+		field     string
+		value     string
+		wantErr   bool
+		testCases []struct {
 			desc      string
 			data      *MessageData
 			wantMatch bool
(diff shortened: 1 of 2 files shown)


cmd/receptor-cl/receptor-cl
dleehr, line 1: I think maybe this file got added by mistake - it's a 75MB binary file. Can you remove it from your branch?

pkg/netceptor/firewall_rules_test.go
```diff
@@ -83,3 +84,495 @@ func TestFirewallRules(t *testing.T) {
...
+		rules           []FirewallRuleData
+		wantErr         bool
+		wantErrContains string
+		wantCount       int
+		testCases       []struct {
+			desc      string
+			ruleIndex int
+			data      *MessageData
+			want      FirewallResult
+		}
```
davemulford: You might want to create this struct as a stand-alone type so it can be referenced in the few places it's used. This will make it easier to change in the future, if needed.

davemulford (dismissed):
GREAT JOB! The regex testing is especially awesome to see; thank you for adding this.

I have one minor suggestion to create a new type for the test case struct in `TestParseFirewallRules`, but totally up to you if you think that would be helpful.

pkg/netceptor/firewall_rules_test.go
melissalkelly: Created firewallRuleTestCase and compareTestCase as stand-alone types

davemulford (dismissed):
New changes to using local test case structs looks great!

pkg/netceptor/firewall_rules_test.go
```diff
@@ -83,3 +97,436 @@ func TestFirewallRules(t *testing.T) {
...
+				},
+				{
+					desc:      "rule 1 continues for non-matching node1",
+					ruleIndex: 1,
+					data:      &MessageData{FromNode: "node1"},
+					want:      FirewallResultContinue,
+				},
+			},
+		},
+		// Test wiith invalid rule data
```
PabloHiro:
```suggestion
		// Test with invalid rule data
```

pkg/netceptor/firewall_rules_test.go
```diff
@@ -83,3 +97,436 @@ func TestFirewallRules(t *testing.T) {
...
+		{
+			name: "invalid rule",
+			rules: []FirewallRuleData{
+				{
+					"action": "",
+				},
+			},
+			wantErr:   true,
+			wantCount: 0,
```
PabloHiro: Would this make sense?
```suggestion
			wantCount: 0,
			wantErrContains: "unknown action:",
```

pkg/netceptor/firewall_rules_test.go
```diff
@@ -83,3 +97,436 @@ func TestFirewallRules(t *testing.T) {
...
+				},
+				{
+					"action":   "drop",
+					"fromnode": "node3",
+				},
+			},
+			wantErr:         true,
+			wantErrContains: "error in rule 1",
+			wantCount:       0,
+		},
```
PabloHiro, line 206: Would it make sense to add more test cases so we cover the different errors in `ParseFirewallRule`?

Something like:

```go
		rules: []FirewallRuleData{
				{
					"invalidkey": "value",
				},
		},
		wantErr: true,
		wantErrContains: "invalid firewall rule. unknown key: invalidkey"
```

```go
		rules: []FirewallRuleData{
				[],
		},
		wantErr: true,
		wantErrContains: "invalid firewall rule. see documentation for syntax"
```

```go
		rules: []FirewallRuleData{
				{
					true: "value",
				},
		},
		wantErr: true,
		wantErrContains: "invalid firewall rule. true must be a string"
```
^ I actually wonder if this last case would print an error correctly.

You built a nice engine to check all of this, if by adding a couple of test cases we get free coverage, I'd say we go for it.

PabloHiro (commented):
Good job @melissalkelly ! Mind reviewing my comments and checking if we can explicitly test the error handling? These are the lines marked as uncovered in Sonarqube, so I think it would be nice to explicitly validate that the errors are printed as expected.

PabloHiro (approved):
LGTM :)

davemulford (approved):

PR #1505, opened 2025-12-18
Improve Test Coverage for udp.go - UDPDialerSession

melissalkelly:
AAP 60059

# Summary
- Add unit tests for UDPDialerSession methods (Send, Recv, Close) and UDPDialer getters in pkg/backends/udp.go
- udp.go now reaches 57.9% total coverage

# Changes
### 1. TestUDPDialer_GetAddr()
- 100% coverage for GetAddr()
### 2. TestUDPDialer_GetTLS()
- 100% coverage for GetTLS()
### 3. TestNewUDPDialer()
- 100% coverage for NewUDPDialer()
### 4. TestUDPDialerSession_Send()
- 87.5% coverage for Send() *not tested is the "partial data sent" error
### 5. TestUDPDialerSession_Recv()
- 100% coverage for Recv() 
### 6. TestUDPDialerSession_Close()
- 100% coverage for Close()
### 7. Helper functions
- setupUDPServer() 
- setupUDPClient() 
- newTestUDPDialerSession
- setupUDPDialerSessionTest
- CheckTestError

### DIFF
--- a/pkg/backends/udp_test.go
+++ b/pkg/backends/udp_test.go
@@ -27,7 +27,7 @@ func TestNewUDPListener(t *testing.T) {
 		{
 			name: "Positive",
 			args: args{
-				address: "127.0.0.1:9997",
+				address: "127.0.0.1:0",
 				logger:  logger.NewReceptorLogger("UDPtest"),
 			},
 			wantErr: false,
@@ -73,7 +73,7 @@ func TestUDPListenerStart(t *testing.T) {
 			fields: fields{
 				laddr: &net.UDPAddr{
 					IP:   net.IPv4(127, 0, 0, 1),
-					Port: 9999,
+					Port: 0,
 					Zone: "",
 				},
 				conn:            &net.UDPConn{},
@@ -137,7 +137,7 @@ func TestUDPDialerStart(t *testing.T) {
 		{
 			name: "Positive",
 			fields: fields{
-				address: "127.0.0.1:9998",
+				address: "127.0.0.1:0",
 				redial:  true,
 				logger:  logger.NewReceptorLogger("UDPtest"),
 			},
@@ -277,16 +277,6 @@ func TestUDPDialer_GetAddr(t *testing.T) {
 				t.Errorf("UDPDialer.GetAddr() = %v, want %v", got, tt.address)
 			}
 		})
-	}
-}
-
-func TestUDPDialer_GetTLS(t *testing.T) {
-	t.Parallel()
-
-	b := &UDPDialer{}
-
-	if got := b.GetTLS(); got != nil {
-		t.Errorf("UDPDialer.GetTLS() = %v, want nil", got)
 	}
 }
 
--- a/pkg/backends/udp_test.go
+++ b/pkg/backends/udp_test.go
@@ -244,6 +244,8 @@ func checkTestError(t *testing.T, err error, wantErr bool, errMsg, funcName stri
 }
 
 func TestUDPDialer_GetAddr(t *testing.T) {
+	t.Parallel()
+
 	tests := []struct {
 		name    string
 		address string
@@ -263,6 +265,8 @@ func TestUDPDialer_GetAddr(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
 			b, err := NewUDPDialer(tt.address, true, nil)
 			if err != nil {
 				t.Errorf("NewUDPDialer() error = %v", err)
@@ -277,6 +281,8 @@ func TestUDPDialer_GetAddr(t *testing.T) {
 }
 
 func TestUDPDialer_GetTLS(t *testing.T) {
+	t.Parallel()
+	
 	b := &UDPDialer{}
 
 	if got := b.GetTLS(); got != nil {
@@ -285,6 +291,8 @@ func TestUDPDialer_GetTLS(t *testing.T) {
 }
 
 func TestNewUDPDialer(t *testing.T) {
+	t.Parallel()
+
 	type args struct {
 		address string
 		redial  bool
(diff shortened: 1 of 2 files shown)


pkg/backends/udp_test.go
```diff
@@ -166,38 +168,353 @@ func TestUDPDialerStart(t *testing.T) {
...
+	}
+
+	if wantErr && errMsg != "" && err != nil {
+		if !strings.Contains(err.Error(), errMsg) {
+			t.Errorf("%s error = %v, want error containing %q", funcName, err, errMsg)
+		}
+	}
+}
+
+func TestUDPDialer_GetAddr(t *testing.T) {
```
davemulford, line 246:
```suggestion
func TestUDPDialer_GetAddr(t *testing.T) {
    t.Parallel()
```

pkg/backends/udp_test.go
```diff
@@ -166,38 +168,353 @@ func TestUDPDialerStart(t *testing.T) {
...
+			name:    "IPv6 address",
+			address: "[::1]:8080",
+		},
+		{
+			name:    "Hostname",
+			address: "example.com:443",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
davemulford, line 267:
```suggestion
		t.Run(tt.name, func(t *testing.T) {
		    t.Parallel()
```

davemulford (changes requested):
Looks good overall. Nice work!

I've suggested enabling the new tests to run in parallel by marking them with `t.Parallel()`. I would add this to all new tests introduced with these changes.

pkg/backends/udp_test.go
```diff
@@ -166,38 +168,375 @@ func TestUDPDialerStart(t *testing.T) {
...
+
+func TestUDPDialer_GetTLS(t *testing.T) {
+	t.Parallel()
+
+	b := &UDPDialer{}
+
+	if got := b.GetTLS(); got != nil {
+		t.Errorf("UDPDialer.GetTLS() = %v, want nil", got)
+	}
+}
```
PabloHiro: I think this test is a symptom that we can remove `GetTLS` from `pkg/backends/udp.go`.

I have checked and it is not used anywhere in the code. Thoughts?

cc @matoval @davemulford @arrestle

pkg/backends/udp_test.go
```diff
@@ -166,38 +168,375 @@ func TestUDPDialerStart(t *testing.T) {
...
+	clientConn := setupUDPClient(t, serverConn.LocalAddr())
+	session := newTestUDPDialerSession(clientConn)
+
+	cleanup := func() {
+		clientConn.Close()
+		serverConn.Close()
+	}
+
+	return cleanup, serverConn, clientConn, session
+}
```
PabloHiro, line 227: In the TCP tests, I did the equivalent of creating two `UDPDialerSession`, one to receive, one to send. And then use the methods `Send` and `Recv` in the tests.

The approach you have followed is very different in the sense that instead of using the functions in `udp.go`, you test one of the functions in isolation and fill the other gap with golang standard library.

What I mean is that for example in `TestUDPDialerSession_Send` you end up having to write: 
```
			buf := make([]byte, UDPMaxPacketLen)
				serverConn.SetReadDeadline(time.Now().Add(1 * time.Second))
				n, _, err := serverConn.ReadFromUDP(buf)
```

Which is basically equivalent to the contents of the `Recv` function.

The analogous happens for `TestUDPDialerSession_Recv`. You end up having to write:
```
                    _, err := serverConn.WriteTo(tt.data, udpClientConn.LocalAddr())
					if err != nil {
						t.Logf("Server failed to send data: %v", err)
					}
```
Which is equivalent to the contents of the `Send` function.

Was all of this intentional? I know it has its advantages because if we ever modify one of these two, the testing of each is isolated, but my personal take is that we are re-implementing logic present in the source code into the tests. Thoughts?

(For clarity: I am not asking you to modify anything, just trying to understand the reasoning behind it)

pkg/backends/udp_test.go
```diff
@@ -166,38 +168,375 @@ func TestUDPDialerStart(t *testing.T) {
 	}
 }
 
+// setupUDPServer creates a UDP listener for testing.
+func setupUDPServer(t *testing.T) *net.UDPConn {
+	t.Helper()
+
+	serverAddr, err := net.ResolveUDPAddr("udp", "127.0.0.1:0")
```
PabloHiro, line 175: If we care about port collision here (using 0 as the binding port, so the library looks for a port that is free), let's stay consistent and use 0 too in other parts of the test file :+1:

PabloHiro (changes requested):

pkg/backends/udp_test.go
davemulford: I agree this can be removed. I tested locally by commenting out the function and confirmed all functional tests run successfully.

This is confirmed by the `gopls` command:
```text
$ gopls implementation pkg/backends/udp.go:22:6
./receptor/pkg/netceptor/interfaces.go:90:6-13
./receptor/tests/functional/mesh/lib.go:729:6-19  // 
```

The first listing is the `Backend` interface defined here:
https://github.com/ansible/receptor/blob/594c21f250dd276fbd346e4afc927ae2d1cb9b64/pkg/netceptor/interfaces.go#L90-L92

The second listing is the `NativeBackend` interface for functional tests:
https://github.com/ansible/receptor/blob/594c21f250dd276fbd346e4afc927ae2d1cb9b64/tests/functional/mesh/lib.go#L729-L733

pkg/backends/udp_test.go
melissalkelly: I did this intentionally because UDP doesn't have an equivalent to net.Pipe() like TCP does, so the approach I took was to create a listener and dialer to test the Send and Recv functions in isolation from each other.
I see the value in consistency across the tests, so I can refactor the UDP tests to create a helper function similar to net.Pipe() for TCP that sets up two bidirectionally connected UDP sessions to be able to use the production Send() and Recv(). Do you think this approach would be best?

davemulford (approved):

PabloHiro (approved):

pkg/backends/udp_test.go
PabloHiro: No problem, just wanted to make sure it was intentional :+1: , good job!

PR #1506, opened 2026-01-06
Fix crypto error handling

davemulford:
This fixes a security issue in how crypto error messages are generated. This is a temporary change that should be reverted once the receptor project can upgrade to a patched Golang version.

### DIFF
--- a/pkg/netceptor/tlsconfig.go
+++ b/pkg/netceptor/tlsconfig.go
@@ -10,7 +10,6 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
-	"net"
 	"os"
 	"strings"
 	"time"
@@ -482,44 +481,6 @@ func ReceptorVerifyFunc(tlscfg *tls.Config, pinnedFingerprints [][]byte, expecte
 
 		return nil
 	}
-}
-
-// handleHostnameError handles a hostname error from the Certificate.Verify function. It
-// was put in place temporarily to mitigate CVE-2025-61729 until the project is able to
-// upgrade to a Golang version containing a fix.
-//
-// This function should be removed once the project can upgrade to a Golang version with
-// a fix for the CVE.
-func handleHostnameError(h x509.HostnameError) error {
-	c := h.Certificate
-	maxNamesIncluded := 100
-
-	var valid strings.Builder
-	if ip := net.ParseIP(h.Host); ip != nil {
-		// Trying to validate an IP
-		if len(c.IPAddresses) == 0 {
-			return errors.New("x509: cannot validate certificate for " + h.Host + " because it doesn't contain any IP SANs")
-		}
-		if len(c.IPAddresses) >= maxNamesIncluded {
-			return fmt.Errorf("x509: certificate is valid for %d IP SANs, but none matched %s", len(c.IPAddresses), h.Host)
-		}
-		for _, san := range c.IPAddresses {
-			if valid.Len() > 0 {
-				valid.WriteString(", ")
-			}
-			valid.WriteString(san.String())
-		}
-	} else {
-		if len(c.DNSNames) >= maxNamesIncluded {
-			return fmt.Errorf("x509: certificate is valid for %d names, but none matched %s", len(c.DNSNames), h.Host)
-		}
-		valid.WriteString(strings.Join(c.DNSNames, ", "))
-	}
-
-	if valid.Len() == 0 {
-		return errors.New("x509: certificate is not valid for any names, but wanted to match " + h.Host)
-	}
-	return errors.New("x509: certificate is valid for " + valid.String() + ", not " + h.Host)
 }
 
 func init() {
--- a/pkg/netceptor/tlsconfig.go
+++ b/pkg/netceptor/tlsconfig.go
@@ -469,8 +469,6 @@ func ReceptorVerifyFunc(tlscfg *tls.Config, pinnedFingerprints [][]byte, expecte
 			logger.Error("RVF failed verify: %s\nRootCAs: %v\nServerName: %s", handledError, tlscfg.RootCAs, tlscfg.ServerName)
 
 			return handledError
-
-			return err
 		}
 
 		// Verify Receptor node ID if required.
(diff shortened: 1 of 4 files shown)


pkg/netceptor/tlsconfig.go
```diff
@@ -478,6 +484,48 @@ func ReceptorVerifyFunc(tlscfg *tls.Config, pinnedFingerprints [][]byte, expecte
...
+//
+// This function should be removed once the project can upgrade to a Golang version with
+// a fix for the CVE.
+func handleHostnameError(h x509.HostnameError) error {
+	c := h.Certificate
+	maxNamesIncluded := 100
+
+	// if !c.hasSANExtension() && matchHostnames(c.Subject.CommonName, h.Host) {
+	// 	return "x509: certificate relies on legacy Common Name field, use SANs instead"
+	// }
```
PabloHiro: Would remove this block, I think it is out of scope to mention if the certificate is up to today's standards or not.

pkg/netceptor/tlsconfig.go
```diff
@@ -460,7 +461,12 @@ func ReceptorVerifyFunc(tlscfg *tls.Config, pinnedFingerprints [][]byte, expecte
 		// Verify the certificate chain.
 		_, err = certs[0].Verify(opts)
 		if err != nil {
-			logger.Error("RVF failed verify: %s\nRootCAs: %v\nServerName: %s", err, tlscfg.RootCAs, tlscfg.ServerName)
+			var hostnameError x509.HostnameError
+			if errors.As(err, &hostnameError) {
+				return handleHostnameError(hostnameError)
+			} else {
+				logger.Error("RVF failed verify: %s\nRootCAs: %v\nServerName: %s", err, tlscfg.RootCAs, tlscfg.ServerName)
+			}
```
PabloHiro, line 467: This changes behavior. It does not log when there is an error, returning too early. Shouldn't we do something like this instead?

```suggestion
			handledError := err
			if errors.As(err, &hostnameError) {
				handledError = handleHostnameError(hostnameError)
			}
			logger.Error("RVF failed verify: %s\nRootCAs: %v\nServerName: %s", handledError, tlscfg.RootCAs, tlscfg.ServerName)

			return handledError
```

pkg/netceptor/tlsconfig.go
```diff
@@ -478,6 +484,48 @@ func ReceptorVerifyFunc(tlscfg *tls.Config, pinnedFingerprints [][]byte, expecte
 	}
 }
 
+// handleHostnameError handles a hostname error from the Certificate.Verify function. It
+// was put in place temporarily to mitigate CVE-2025-61729 until the project is able to
+// upgrade to a Golang version containing a fix.
+//
+// This function should be removed once the project can upgrade to a Golang version with
+// a fix for the CVE.
```
PabloHiro: I have looked up how to enforce this via code:

* Create a separate file just with `handleHostnameError`
* Put this at the beginning:
```
//go:build go1.23
```
* This way, whenever we change the go version, this file will not compile, forcing us to refactor


Reference: https://go.dev/wiki/Comments

PabloHiro: I would also move the tests to a separate file as well, so they are easy to remove in the future.

pkg/netceptor/tlsconfig_test.go
```diff
@@ -953,6 +972,95 @@ func TestTLSClientConfigMinTLS13(t *testing.T) {
...
+			testName:        "Exactly 100 DNS names",
+			hostnameToMatch: "MyHostname",
+			nameCount:       100,
+			expectedError:   "x509: certificate is valid for 100 names, but none matched MyHostname",
+		},
+		{
+			testName:        "Less than 100 IP SANs",
+			hostnameToMatch: "127.0.0.1",
+			nameCount:       3,
+			expectedError:   "abc",
```
PabloHiro:
```suggestion
```

Would remove it (it gets initialized to `""`) and later would add:

```
if tt.expectedError != "" 
```

PabloHiro (changes requested):

pkg/netceptor/tlsconfig.go
davemulford: I'll remove it. It's left over from the [fix in Go](https://go-review.googlesource.com/c/go/+/725920) that I commented out.

pkg/netceptor/tlsconfig.go
davemulford: Great catch! Absolutely this should be logged.

pkg/netceptor/tlsconfig.go
davemulford: Thank you for researching this. That was something I wanted to address but wasn't sure how to go about it. This seems like it will work, so I'll make these changes as well.

pkg/netceptor/tlsconfig_test.go
davemulford: There are two ways to cause the issue:

1. Loading `Certificate.IPAddresses` with a lot of IPs.
2. Loading `Certificate.DNSNames` with a lot of FQDNs.

This test case will change to support IP Addresses once I've changed the test helper functions to allow for it.

PabloHiro: Ah my bad, I thought it was a bogus string for a scenario that was not supposed to fail.

pkg/netceptor/tlsconfig_hostname_error_test.go
```diff
@@ -0,0 +1,122 @@
+//go:build go1.23
```
PabloHiro: I would remove this from the test file. This way it will fail immediately and it obvious we need to remove this file too.

davemulford: Removed!

PabloHiro (dismissed):
LGTM

pkg/netceptor/tlsconfig_hostname_error.go
```diff
@@ -0,0 +1,50 @@
...
+// upgrade to a Golang version containing a fix.
+//
+// This function should be removed once the project can upgrade to a Golang version with
+// a fix for the CVE.
+func handleHostnameError(h x509.HostnameError) error {
+	c := h.Certificate
+	maxNamesIncluded := 100
+
+	var valid strings.Builder
+	if ip := net.ParseIP(h.Host); ip != nil {
```
tyraziel:
```suggestion
	if net.ParseIP(h.Host) != nil {
```

I agree with the sonar scan here.  Can we just go with `net.ParseIP(h.Host) != nil`?

tyraziel (changes requested):

matoval (approved):
LGTM

PabloHiro (approved):
LGTM

tyraziel (approved):
My request for changes comment has been addressed.  Approving as 3rd approval to remove the block on my "request for changes".

PR #1509, opened 2026-01-15
AAP - 60058 Improve test coverage for tcp.go

melissalkelly:
AAP - 60058

## Overview

Increases tcp.go unit test coverage from 63.6% to 90.4%

Adding unit tests for:

- TCPListenerCfg.Run
- TCPDialerCfg.Run
- TCPDialerCfg.PreReload
- TCPListenerCfg.PreReload
- TCPDialerCfg.Reload
- TCPListenerCfg.Reload

Improving unit tests for:

- TCPDialer.Start
- TCPListener.Start

## Changes
- TCPDialerCfg.PreReload & TCPDialerCfg.Reload 0% to 100% coverage
- TCPListenerCfg.PreReload & TCPListenerCfg.Reload 0% to 100% coverage
- TCPListenerCfg.Run 0% to 75% coverage
- TCPDialerCfg.Run 0% to 80% coverage
- TCPListener.Start 65.7% to 88.6% coverage
- TCPDialer.Start 84.6% to 100.0% coverage
- Added shared test structures tcpDialerCfgTest & tcpListenerCfgTest

Assisted by Claude

### DIFF
--- a/pkg/backends/tcp_test.go
+++ b/pkg/backends/tcp_test.go
@@ -115,6 +115,7 @@ func TestTCPDialerStart(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
 			testLogger := logger.NewReceptorLogger("TCPtest")
 
 			address := tt.address
@@ -310,15 +311,15 @@ func TestTCPListener_GetTLS(t *testing.T) {
 func TestTCPListenerStart(t *testing.T) {
 	t.Parallel()
 	tests := []struct {
-		name          string
-		address       string
-		tls           *tls.Config
+		name           string
+		address        string
+		tls            *tls.Config
 		makeConnection bool
 	}{
 		{
-			name:          "Non-TLS listener with connection",
-			address:       "127.0.0.1:0",
-			tls:           nil,
+			name:           "Non-TLS listener with connection",
+			address:        "127.0.0.1:0",
+			tls:            nil,
 			makeConnection: true,
 		},
 		{
@@ -329,6 +330,7 @@ func TestTCPListenerStart(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
 			b, err := NewTCPListener(tt.address, tt.tls, logger.NewReceptorLogger("TCPtest"))
 			if err != nil {
 				t.Errorf("NewTCPListener() error = %v", err)
@@ -930,6 +932,7 @@ func TestTCPDialerCfg_Run(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
 			netceptor.MainInstance = netceptor.New(context.Background(), "tcp_dialer_run_test")
 			err := tt.cfg.Run()
 			if (err != nil) != tt.wantErr {
@@ -1022,6 +1025,7 @@ func TestTCPListenerCfg_Run(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
 			netceptor.MainInstance = netceptor.New(context.Background(), "tcp_listener_run_test")
 			err := tt.cfg.Run()
 			if (err != nil) != tt.wantErr {
@@ -1044,5 +1048,3 @@ func TestTCPListenerCfg_Reload(t *testing.T) {
 		t.Errorf("TCPListenerCfg.Reload() error = %v", err)
 	}
 }
-
-
--- a/pkg/backends/tcp_test.go
+++ b/pkg/backends/tcp_test.go
@@ -158,7 +158,7 @@ func TestTCPDialerStart(t *testing.T) {
 			if got == nil {
 				t.Errorf("TCPDialer.Start() returned nil channel")
 			}
-			
+
 			// Wait for connection with 10ms timeout. This is sufficient for loopback
 			// connections (typical datacenter RTT is 0.5ms) while keeping tests fast.
 			select {
@@ -365,9 +365,9 @@ func TestTCPListenerStart(t *testing.T) {
 				} else {
 					conn.Close()
 				}
-				
+
 				// Wait for connection with 10ms timeout. This is sufficient for loopback
-			    // connections (typical datacenter RTT is 0.5ms) while keeping tests fast.
+				// connections (typical datacenter RTT is 0.5ms) while keeping tests fast.
 				select {
 				case <-got:
 					// Connection received
--- a/pkg/backends/tcp_test.go
+++ b/pkg/backends/tcp_test.go
@@ -10,7 +10,22 @@ import (
 	"time"
 
 	"github.com/ansible/receptor/pkg/logger"
+	"github.com/ansible/receptor/pkg/netceptor"
 )
+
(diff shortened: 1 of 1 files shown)


pkg/backends/tcp_test.go
```diff
@@ -111,14 +152,17 @@ func TestTCPDialerStart(t *testing.T) {
...
+				t.Errorf("TCPDialer.Start() error = %v", err)
 
 				return
 			}
 			if got == nil {
 				t.Errorf("TCPDialer.Start() returned nil channel")
 			}
+
+			// Allow time for connection attempt to complete before context cancellation
+			time.Sleep(10 * time.Millisecond)
```
PabloHiro: I think the golang way to do this is to make use of channels.
```suggestion
			// Allow time for connection attempt to complete before context cancellation
			  select {
              case <-got:
                  // Connection received
              case <-time.After(1 * time.Second):
                  t.Fatal("timeout waiting for connection")
              }
```

pkg/backends/tcp_test.go
```diff
@@ -295,14 +346,26 @@ func TestTCPListenerStart(t *testing.T) {
...
+				// Make a connection to trigger accept path
+				conn, err := net.Dial("tcp", b.GetAddr())
+				if err != nil {
+					t.Errorf("Failed to connect to listener: %v", err)
+				} else {
+					conn.Close()
+				}
+			}
+
+			time.Sleep(10 * time.Millisecond)
```
PabloHiro: Same here, use the channel inside the `tt.MakeConnection` so the test finishes as soon as possible.
```suggestion
					conn.Close()
				}
				select {
				case <-got:
					// Connection received
				case <-time.After(10 * time.Millisecond):
					t.Fatal("timeout waiting for connection")
				}
			}
```

pkg/backends/tcp_test.go
```diff
@@ -823,3 +864,187 @@ func TestTCPDialerCfg_Prepare(t *testing.T) {
...
+				Cost:    1.0,
+				TLS:     "nonexistent-tls-config",
+			},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			netceptor.MainInstance = netceptor.New(context.Background(), "tcp_dialer_run_test")
```
PabloHiro, line 943: This shares the same pointer as with the other tests:
https://github.com/ansible/receptor/blob/366eddb5fae3ade60b8bc3975422c96197a405ec/pkg/netceptor/netceptor.go#L46

And it causes race-conditions. Can we remove `t.Parallel()` from here?

pkg/backends/tcp_test.go
```diff
@@ -823,3 +864,187 @@ func TestTCPDialerCfg_Prepare(t *testing.T) {
...
+}
+
+func TestTCPDialerCfg_Reload(t *testing.T) {
+	t.Parallel()
+	cfg := TCPDialerCfg{
+		Address: "127.0.0.1:8080",
+		Redial:  true,
+		Cost:    1.0,
+	}
+	netceptor.MainInstance = netceptor.New(context.Background(), "tcp_dialer_reload_test")
```
PabloHiro, line 958: Same, let's remove `t.Parallel()`

pkg/backends/tcp_test.go
```diff
@@ -823,3 +864,187 @@ func TestTCPDialerCfg_Prepare(t *testing.T) {
...
+				Cost:     1.0,
+				TLS:      "nonexistent-tls-config",
+			},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			netceptor.MainInstance = netceptor.New(context.Background(), "tcp_listener_run_test")
```
PabloHiro, line 1033: Same, let's remove `t.Parallel()`

pkg/backends/tcp_test.go
```diff
@@ -823,3 +864,187 @@ func TestTCPDialerCfg_Prepare(t *testing.T) {
...
+}
+
+func TestTCPListenerCfg_Reload(t *testing.T) {
+	t.Parallel()
+	cfg := TCPListenerCfg{
+		BindAddr: "127.0.0.1",
+		Port:     0,
+		Cost:     1.0,
+	}
+	netceptor.MainInstance = netceptor.New(context.Background(), "tcp_listener_reload_test")
```
PabloHiro, line 1048: Same, let's remove `t.Parallel()`

PabloHiro (changes requested):
The initial version of your PR was fine, I feel like the linter made you panic 😓

dleehr (changes requested):
Overall looking very good, coverage increased substantially, table-driven patterns are adopted, error cases are covered, and the implementation isn't changed (just tests) so we're confident it doesn't break anything. Nice work 🎉 

I'm requesting changes, aligned with Pablo's feedback around the MainInstance and protecting against race conditions (by removing t.Parallel in those cases) and also using channels as a best practice for this logic. The suggested changes are great examples.

I also used the excellent [pr-review-skill](https://github.com/benthomasson/claude-skills/tree/main/skills/pr-review-skill) to help. It looked at the code as well as the existing discussion. I agree with the recommendations and it caught some nested / redundant `t.Parallel()` calls:

1. **Remove `t.Parallel()` from 4 test functions** (as suggested by @PabloHiro)
2. **Replace `time.Sleep()` with channel-based synchronization** (as suggested by @PabloHiro)
3. **Consider removing redundant `t.Parallel()` calls**:
   - Lines 902, 946, 1004, 1039 have BOTH top-level `t.Parallel()` and nested `t.Parallel()` in subtests
   - After removing top-level calls per item #1, also remove the nested calls for consistency
   - If you want parallel subtests in the future, only use the nested call

But looking good, keep going!

pkg/backends/tcp_test.go
```diff
@@ -295,14 +347,30 @@ func TestTCPListenerStart(t *testing.T) {
...
+				if err != nil {
+					t.Errorf("Failed to connect to listener: %v", err)
+				} else {
+					conn.Close()
+				}
+				select {
+				case <-got:
+					// Connection received
+				case <-time.After(10 * time.Millisecond):
+					t.Fatal("timeout waiting for connection")
```
dleehr, line 375: Without additional context, 10ms timeout sounds like a short amount of time to wait for something  network-based. Can we add an explanation about why this value was chosen and why it's sufficient?

In my experience, I've seen these values lead to flaky test cases, and it's always valuable to have the context about how and why we choose these values (magic numbers). What do you think about adding a comment to explain why 10ms was chosen and why it's sufficient[^1]? That way, if we **do** see this test start to fail occasionally, we have that context.

[^1]: Having said that, the value chosen doesn't appear to be causing any failures. Within the unit test it's a local TCP connection on the loopback interface (e.g. not across the Internet or WAN), so 10ms might be more than we ever need. Per [Latency Numbers Everyone Should Know](https://static.googleusercontent.com/media/sre.google/en//static/pdf/rule-of-thumb-latency-numbers-letter.pdf), a round-trip within a **datacenter** is 0.5ms. I don't think we'll have a problem with 20x that, but would still appreciate a comment.

dleehr (dismissed):
LGTM. Thanks for addressing the feedback.

The only note I have is about the 10ms timeout in one of the tests, please see my in-line comment there.

dleehr (approved):

PabloHiro (approved):
LGTM. Thank you @melissalkelly 🙏

PR #1512, opened 2026-01-21
Provide 'ref' as an env variable rather than direct user input

dleehr:
Updates the `stage.yml` GitHub action to provide the checkout ref as an environment variable rather than directly from user input. This change addresses a Sonarqube security recommendation where a user input could trigger code execution. By providing the value in an environment variable, we prevent that issue.

AAP-63087

### DIFF
--- a/.github/workflows/stage.yml
+++ b/.github/workflows/stage.yml
@@ -33,6 +33,8 @@ jobs:
     permissions:
       packages: write
       contents: write
+    env:
+      REF: ${{ github.event.inputs.ref }}
     steps:
       - name: Verify inputs
         run: |
@@ -53,7 +55,7 @@ jobs:
       - name: Checkout receptor
         uses: actions/checkout@v6
         with:
-          ref: ${{ github.event.inputs.ref }}
+          ref: ${{ env.REF }}
 
       - name: Install python
         uses: actions/setup-python@v6


dleehr:
This change applies to a workflow that we only run when releasing receptor. To validate the approach before merging, I have created a test workflow with the same pattern in a sandbox repo: https://github.com/dleehr/actions-sandbox/blob/main/.github/workflows/referee.yml
```yaml
    env:
      REF: ${{ github.event.inputs.ref }}
    steps:


      - name: Show github SHA before checkout
        run: |
          echo "Before checkout, GITHUB_SHA is $GITHUB_SHA"


      - name: Checkout ref (interpolate to env var)
        uses: actions/checkout@v6
        with:
          ref: ${{ env.REF }}
```

Test showing that the syntax works as expected, with input `b7f42a8b3cf1db56fa0f3715415cc7708620416d` (the initial commit of that sandbox repo: https://github.com/dleehr/actions-sandbox/actions/runs/21225474689/job/61071175817. 

Screenshot 2026-01-21 at 4 32 52 PM

Test showing that when an invalid `ref` is provided, the workflow fails: https://github.com/dleehr/actions-sandbox/actions/runs/21225479702/job/61071187222

Screenshot 2026-01-21 at 4 34 29 PM

matoval (approved):

davemulford (approved):
LGTM!

PR #897, opened 2023-11-20
Add tagging task to the stage playbook

AaronH88:
Annotated tags need to be created as part of the ci pipeline.

PR #1513, opened 2026-01-21
Use 'tr' instead of 'node' for string lowercase operation

dleehr:
Looked at this workflow a bit closer after #1512.

This step in the workflow was reading the GitHub Repository owner (e.g. 'ansible') from a variable, then executing JavaScript (via `node`) to lowercase the string so that the repo name we use later is guaranteed to be lowercase.

I find it overkill to launch a JavaScript virtual machine to lowercase a string (especially since it's already lowercase on this repo 🤔), when [coreutils `tr '[:upper:]' '[:lower:]'`](https://www.gnu.org/software/coreutils/manual/html_node/tr-invocation.html) is a simple, portable way to do this in Linux. I made this change because I believe in Linux and coreutils 🎸. 

AAP-63087

### DIFF
--- a/.github/workflows/stage.yml
+++ b/.github/workflows/stage.yml
@@ -82,7 +82,7 @@ jobs:
 
       - name: Lowercase github owner
         run: |
-          echo "OWNER=$(echo 'console.log("${{ github.repository_owner }}".toLowerCase())' | node -)" >> $GITHUB_ENV
+          echo "OWNER=$(echo $GITHUB_REPOSITORY_OWNER | tr '[:upper:]' '[:lower:]')" >> $GITHUB_ENV
 
       - name: Build container image
         run: |
--- a/.github/workflows/stage.yml
+++ b/.github/workflows/stage.yml
@@ -80,6 +80,8 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
           registry: ghcr.io
 
+      # Image repository names must be lowercase, so get lowercase repo owner for ghcr.io
+      # See https://github.com/orgs/community/discussions/27086.
       - name: Lowercase github owner
         run: |
           echo "OWNER=$(echo $GITHUB_REPOSITORY_OWNER | tr '[:upper:]' '[:lower:]')" >> $GITHUB_ENV


dleehr:
Verified in sandbox, see https://github.com/dleehr/actions-sandbox/blob/main/.github/workflows/referee.yml#L40. Copied this line directly into the PR here

See https://github.com/dleehr/actions-sandbox/actions/runs/21229482037/job/61084643395 for proof.

Screenshot 2026-01-21 at 6 34 44 PM

matoval:
> Verified in sandbox, see https://github.com/dleehr/actions-sandbox/blob/main/.github/workflows/referee.yml#L40. Copied this line directly into the PR here
> 
> See https://github.com/dleehr/actions-sandbox/actions/runs/21229482037/job/61084643395 for proof.
> 
> Screenshot 2026-01-21 at 6 34 44 PM

Does this work if the codeowner have uppercase letters in their name?

dleehr:
> Does this work if the codeowner have uppercase letters in their name?

Yes, the reason for the `Lowercase github owner` is because the workflow pushes an image to GitHub Container Registry. Image names on ghcr.io must be lowercase, see https://github.com/orgs/community/discussions/27086:

> Image names must be lower case, try using the lower case version of your username. You can create an environment variable with the lower case version, so you don’t have to hardcode it.

Apparently this is a limitation of Docker's image registries - they only support lowercase image names. Since GitHub allows owners and repos to have mixed case, you need to lowercase them for the ghcr.io equivalents.

Now, that's not a problem for **this** repo (`ansible/receptor` is already lowercase, so `ghcr.io/ansible/receptor` works just fine). But, when running this in a user's fork, it would fail if the username has uppercase characters in their name. Looking at history I see the logic was added in #897 by @AaronH88, who does have uppercase characters 😏. So I suspect when he was developing this on his fork, it failed to push to `ghcr.io/AaronH88/receptor`. He added lowercasing in f726020f41ce1f4b23a62f938224f028576f609d, which would push to `ghcr.io/aaronh88/receptor` and that should succeed. It's a good practice, even if `ansible/receptor` is already lowercase.

I realized that I **can** run this on my fork, so I didn't technically need a sandbox repo (though that helped speed things up). On a fork, it pushes the tag and image to my fork, and doesn't mess with our tags and images under the `ansible` org. See https://github.com/dleehr/receptor/actions/runs/21236428911/job/61105333717. Waiting for it to finish but looks like it's working as expected.

---
This PR does not fix a bug or change that behavior. I just thought it was odd to launch [Node.js](https://nodejs.org/en) to do something that a [POSIX-standard `tr` does](https://en.wikipedia.org/wiki/Tr_(Unix)).

matoval:
My question was since your username doesn't have uppercase letters how can we test this, I forked the sandbox repo just to realize I don't have uppercase letters in my username either 😅.

I just want to verify that it works with uppercase letters in a username.

PabloHiro (approved):

dleehr:
> My question was since your username doesn't have uppercase letters how can we test this, I forked the sandbox repo just to realize I don't have uppercase letters in my username either 😅.
> 
> I just want to verify that it works with uppercase letters in a username.

Ah. Great question, sorry I answered a different one.

I have an org that does have uppercase characters. I forked my sandbox repo there and ran the test workflow. Yes it gets lowercased properly:

https://github.com/CloneApp/actions-sandbox/actions/runs/21248801495/job/61144742689

Screenshot 2026-01-22 at 7 42 30 AM

matoval (approved):

PR #1514, opened 2026-01-22
Document intentional InsecureSkipVerify usage for non-TLS mode

matoval:
Update documentation for generateClientTLSConfig to accurately explain why InsecureSkipVerify=true is intentional for non-TLS mode. Receptor supports both TLS and non-TLS connections, and this code path is only reached when no TLS is configured.

### DIFF
--- a/pkg/netceptor/conn.go
+++ b/pkg/netceptor/conn.go
@@ -644,7 +644,7 @@ func verifyServerCertificate(rawCerts [][]byte, _ [][]*x509.Certificate) error {
 // InsecureSkipVerify is intentionally true here because:
 // 1. This path is for non-TLS mode where the server uses auto-generated self-signed certs
 // 2. Custom verification is still performed via VerifyPeerCertificate callback
-// 3. The callback (verifyServerCertificate) checks for the insecure connection marker
+// 3. The callback (verifyServerCertificate) checks for the insecure connection marker.
 func generateClientTLSConfig(host string) *tls.Config {
 	return &tls.Config{
 		//nolint:gosec // G402: InsecureSkipVerify is intentional for non-TLS mode; see function comment above
--- a/pkg/netceptor/conn.go
+++ b/pkg/netceptor/conn.go
@@ -421,7 +421,7 @@ func (s *Netceptor) DialContext(ctx context.Context, node string, service string
 	}
 
 	if tlscfg == nil {
-		tlscfg = generateClientTLSConfig(s.NodeID())
+		tlscfg = generateClientTLSConfig(s.NodeID(), false)
 	} else {
 		tlscfg = tlscfg.Clone()
 		tlscfg.NextProtos = []string{"netceptor"}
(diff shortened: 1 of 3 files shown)


PabloHiro (approved):

davemulford (approved):

pkg/netceptor/conn.go
```diff
@@ -637,9 +637,17 @@ func verifyServerCertificate(rawCerts [][]byte, _ [][]*x509.Certificate) error {
...
 }
 
+// generateClientTLSConfig creates a TLS config for non-TLS mode connections.
+// This is only called when no TLS configuration is provided (see DialContext).
+// Receptor supports both TLS and non-TLS connections per the documentation.
+// When TLS is configured, users provide their own tls.Config via GetClientTLSConfig.
+// InsecureSkipVerify is intentionally true here because:
+// 1. This path is for non-TLS mode where the server uses auto-generated self-signed certs
+// 2. Custom verification is still performed via VerifyPeerCertificate callback
+// 3. The callback (verifyServerCertificate) checks for the insecure connection marker.
```
dleehr, line 647: Since most of the change is comments, I'd like to make sure I understand what we're documenting before we accept this decision, and I need some help following this.

> generateClientTLSConfig creates a TLS config for non-TLS mode connections.

Can you help me understand this? We're saying this function is for **non** TLS connections, so why do we need a TLS config? 

I interpret `non-TLS` to mean plain TCP or UDP (not encrypted, no certs, just bytes ona wire). But reading the bullet points sounds like this is for the `auto-generated self-signed certs` configuration. If we have certs (even self-signed ones), that's still TLS, right? Can you help me understand why we're saying it's OK to set `InsecureSkipVerify=true` here? Is there something about the original `InsecureSkipVerify is set true in test context only; production usage is config-driven.` that's not accurate?

One more question - Does this linter config change address SonarQube's finding?

dleehr (commented):
Was just about to submit my review but saw this merged. 🚢

pkg/netceptor/conn.go
matoval: After digging a bit deeper to answer you questions I think some of the wording of the comment I added is a bit confusing. I will make a new PR clearing up the comment.

Let me answer your questions,

**"Non-TLS mode" terminology**:

"non-TLS mode" is misleading. Receptor has two networking layers:                                           
                                                                                                                                            
- Backend layer (TCP/UDP/WebSocket) - connects nodes into a mesh                                                                          
- Service layer (QUIC) - used by Dial/Listen for application connections                                                                  
                                                                                                                                            
generateClientTLSConfig is for the service layer. QUIC always requires TLS at the protocol level, so it's not really "non-TLS". The distinction is between auto-generated self-signed certs (when tlscfg == nil) vs user-provided certs (when users call GetClientTLSConfig()). A better term would be "auto-cert mode" or "default mode".                                                         
                                                                                                                                            
**Why InsecureSkipVerify=true is OK**:

It's safe due to two-layer verification:                                                                                                  
1. InsecureSkipVerify=true skips standard CA chain validation (necessary since certs are self-signed)                                     
2. The VerifyPeerCertificate callback still runs and checks that the server cert has the marker CommonName "netceptor-insecure-common-name"
                                                                                                                                            
This prevents cross-mode attacks - a client in auto-cert mode will reject connections to properly-configured TLS servers.                 
                                                                                                                                            
**Old comment accuracy** : 
                                                                                                                                                                                                                                   
The old comment ("InsecureSkipVerify is set true in test context only") was inaccurate - this code path is used in production for simple deployments, not just tests.                                                                                                              
                                                                                                                                            
**SonarQube**: 
                                                                                                                                                                                                                                                                    
This PR addresses gosec/golangci-lint but won't suppress the SonarQube issue. SonarQube uses its own rule (go:S4830) and doesn't recognize //nolint directives. I'd suggest marking it as "Won't Fix" in the SonarQube UI with justification explaining the two-layer verification approach.

PR #1516, opened 2026-01-22
Clarify TLS auto-cert mode comments in conn.go

matoval:
Clarify TLS auto-cert mode comments in conn.go to replace confusing "non-TLS mode" terminology and add comment for generateServerTLSConfig


PabloHiro (approved):

dleehr (approved):
Thanks, this clarifies!

PR #1517, opened 2026-01-22
Fix cancelled work units being deleted prematurely on restart

matoval:
Fix premature deletion of cancelled work units on restart                                                                                 
                                                                                                                                            
When a work unit was cancelled but not released, Restart() incorrectly                                                                    
passed forRelease=true to runAndMonitor, causing the work unit directory                                                                  
to be deleted before Release() was explicitly called. This resulted in                                                                    
"unknown work unit" errors when attempting to release the unit.                                                                           
                                                                                                                                            
Changed to pass red.LocalReleased instead of hardcoded true,                                                                     
so deletion only occurs when the unit has actually been released.                                                                         
                                                                                                                                            
Added unit test for restart_local_cancelled scenario.

This fix replaces PR #1275

### DIFF
--- a/pkg/workceptor/remote_work.go
+++ b/pkg/workceptor/remote_work.go
@@ -646,7 +646,7 @@ func (rw *remoteUnit) startOrRestart(start bool) error {
 
 		return rw.runAndMonitor(rw.topJC, false, rw.StartRemoteUnit)
 	} else if red.LocalReleased || red.LocalCancelled {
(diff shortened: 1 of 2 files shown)


PabloHiro (approved):

davemulford (approved):
LGTM. Great comment in the test explaining how gomock will fail instead of needing an assert statement.

PR #1520, opened 2026-02-06
[AAP-63315] Move workflow permissions from workflow to job level 

arrestle:
## Summary
Moves workflow-level `contents: write` permission to job-level scope to improve security posture.

## Changes
- Removed `permissions: contents: write` from workflow level
- Added `permissions: contents: write` to `publish` job only
- The `promote` job now runs with default read-only permissions

## Rationale
The `publish` job requires write access to upload release artifacts to GitHub Releases via `softprops/action-gh-release@v2`. The `promote` job only interacts with external services (Quay.io and PyPI) and does not need GitHub repository permissions.

This change follows the principle of least privilege and resolves the SonarQube security issue flagged in [AAP-63315](https://issues.redhat.com/browse/AAP-63315).

## References
- JIRA: [AAP-63315](https://issues.redhat.com/browse/AAP-63315)
- SonarQube Issue: [githubactions:S7631](https://sonarcloud.io/project/issues?issues=AZuYoXPC5s2R-CmnIdZt&open=AZuYoXPC5s2R-CmnIdZt&id=ansible_receptor)
- [GitHub Permissions Documentation](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)

### DIFF
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -4,9 +4,6 @@ on:  # yamllint disable-line rule:truthy
   release:
     types: [published]
 
-permissions:
-  contents: write
-
 jobs:
   promote:
     runs-on: ubuntu-latest
@@ -86,6 +83,8 @@ jobs:
 
   publish:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     env:
       VERSION: ${{github.event.release.tag_name}}
     steps:


PabloHiro (approved):

fincamd (approved):

tyraziel (approved):
LGTM - CI :)

tyraziel (changes requested):
Expanded the diff, can we figure out what permissions should go on each job?

arrestle:
> Expanded the diff, can we figure out what permissions should go on each job?

Good catch! I've analyzed what each job needs, following the pattern @PabloHiro established in his recent permissions fix (ac6ae1a):

**`promote` job**: No explicit permissions needed (defaults to read-only). This job only:
- Pushes Docker images to Quay.io (uses QUAY secrets)
- Uploads Python packages to PyPI (uses PYPI secrets)
- No GitHub repository access required

**`publish` job**: Requires `contents: write` because it uses `softprops/action-gh-release@v2` to upload release artifacts to the GitHub Release.

This follows the same principle as the SonarCloud workflow where @PabloHiro explicitly defined `permissions: read-all` at the workflow level. In our case, we're scoping `contents: write` to only the `publish` job that requires it, while `promote` runs with default read-only permissions.

Per [GitHub's permissions documentation](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token), the `contents` permission controls access to repository contents and releases. It's only needed when modifying the GitHub repository itself (like uploading release assets). External services like Quay.io and PyPI use their own authentication mechanisms through secrets and don't require GitHub repository permissions.

The change correctly implements the principle of least privilege per SonarQube's security recommendation.

comment assisted-by: Claude.

tyraziel (approved):
Understood on default permissions as read-only and that we don't need to explicitly state those in the individual jobs.

PR #1539, opened 2026-04-03
Fix warnings from connection close in work results

davemulford:
The "work results" subcommand in ControlFunc called cfo.Close() after streaming results, closing the connection before RunControlSession's read loop could detect EOF. This produced WARNING logs:

  "Could not read in control service: use of closed network connection"
  "Could not close connection: close unix ... use of closed network connection"

The connection is already closed by deferred conn.Close() calls in both SetupConnection() and RunControlSession(), making the explicit close redundant. Removing it allows the read loop to see a clean EOF and exit gracefully.

Fixes AAP-43847

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented the connection from closing prematurely while streaming results.
* **Tests**
  * Added a test that verifies the connection remains open during result retrieval.

### DIFF
--- a/pkg/workceptor/controlsvc.go
+++ b/pkg/workceptor/controlsvc.go
@@ -433,11 +433,6 @@ func (c *workceptorCommand) ControlFunc(ctx context.Context, nc controlsvc.Netce
 			return nil, err
 		}
 
-		err = cfo.Close()
-		if err != nil {
-			return nil, err
-		}
-
 		return nil, nil
 	}
 
--- a/pkg/workceptor/controlsvc_results_test.go
+++ b/pkg/workceptor/controlsvc_results_test.go
@@ -1,0 +1,134 @@
+//go:build !no_workceptor
+// +build !no_workceptor
+
+package workceptor_test
+
+import (
+	"context"
+	"os"
+	"path"
+	"testing"
+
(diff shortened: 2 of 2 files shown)


dleehr (dismissed):
LGTM.

Sonar has a suggestion in the tests though 😄

pkg/workceptor/controlsvc.go
```diff
@@ -433,11 +433,6 @@ func (c *workceptorCommand) ControlFunc(ctx context.Context, nc controlsvc.Netce
 			return nil, err
 		}
 
-		err = cfo.Close()
```
matoval, line 436: Should we check that the err is caused by the connection being closed? Is there any errors that we don't want to ignore?

davemulford: Which error would be checked? I've removed this line so `err` isn't being set by the call to `cfo.Close()` anymore.

matoval: Nevermind, I misread the err check.

matoval (approved):
LGTM

PabloHiro (approved):
LGTM. Awesome catch. This will remove thousands (millions?) of unnecessary log lines all over the world 🙏

PR #1562, opened 2026-07-16
Upgrade quic-go to v0.59.1

fosterseth:
## Summary
- Upgrade `github.com/quic-go/quic-go` from v0.48.2 to v0.59.1
- The `logging` subpackage was removed upstream; replace the custom `tracer()` method with `qlog.DefaultConnectionTracer` which provides equivalent QLOGDIR-based tracing
- Update tests to use the new API (`isClient bool` instead of `logging.Perspective`, `qlogwriter.Trace` instead of `*logging.ConnectionTracer`)

## Test plan
- [x] `make receptor` builds successfully
- [x] Existing unit tests pass (`go test ./pkg/netceptor/...`)
- [x] Set `QLOGDIR` and verify qlog files are created during QUIC connections

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved QUIC qlog connection tracing behavior to use the latest built-in tracer and ensure consistent output.
  * Refined qlog file-path expectations to better match actual connection tracing behavior.

* **Tests**
  * Updated QUIC qlog tracing tests to validate the default tracer directly and to improve file-path validation.

* **Chores**
  * Updated QUIC and testing library versions to newer releases.


pkg/netceptor/conn.go
```diff
@@ -192,28 +190,6 @@ func (s *Netceptor) GetConfigForClientOverride(tlscfg *tls.Config) func(*tls.Cli
...
 }
 
-func (s *Netceptor) tracer(ctx context.Context, p logging.Perspective, connID quic.ConnectionID) *logging.ConnectionTracer {
-	qlogPath := os.Getenv("QLOGDIR")
-	if qlogPath != "" {
-		role := "server"
-		if p == logging.PerspectiveClient {
-			role = "client"
-		}
-		filename := fmt.Sprintf("log_%x_%s.qlog", connID, role)
```
fosterseth, line 202: the logging filename has a changed a bit since we are switching to whatever the built-in filename is. I grepped ansible repos for indication of reading these qlog files and didn't see any, so this should be a safe change.

pkg/netceptor/netceptor_test.go
```diff
@@ -881,10 +881,8 @@ func TestTooShortSetMaxConnectionIdleTime(t *testing.T) {
 
 func TestTracerReturnsNewConnectionTracer(t *testing.T) {
 	t.Parallel()
-	s := New(context.Background(), "node1")
-	p := logging.PerspectiveClient
 	os.Setenv("QLOGDIR", "/tmp/")
-	trace := s.tracer(s.context, p, quic.ConnectionID{})
+	trace := qlog.DefaultConnectionTracer(context.Background(), true, quic.ConnectionID{})
```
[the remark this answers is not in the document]
fosterseth: good catch, resolved

dleehr (approved):
Glad it doesn't seem to be a huge change - some previous quic-go upgrades were far more involved 😅 .

---

At first I thought we were regressing on functionality by deleting the logging perspective (client or server) - but it looks like we're not.

Instead, my understanding is that 

1. the interface for `Tracer` in `quic.Config` changed (previously it was Context, Perspective, ConnectionID, now Context, Bool, ConnectionID)
2. The [DefaultConnectionTracer](https://pkg.go.dev/github.com/quic-go/quic-go/qlog#DefaultConnectionTracer) handles creating and logging to the file in `QLOGDIR`, so we no longer need to do that?

If so, I think the only functional change is maybe the log file name format will now be determined by `DefaultConnectionTracer`:
> DefaultConnectionTracer creates a qlog file in the qlog directory specified by the QLOGDIR environment variable. File names are <odcid>_<perspective>.sqlog. Returns nil if QLOGDIR is not set.

So my only question would be - do we have any support documents or anything that would need to be updated for the change?

pkg/netceptor/conn.go
dleehr: Yeah I called this out too - I agree it's probably not a meaningful _code_ change, but since this is the file generated when tracing is enabled for deep diagnostic logs, I wonder if there's any support docs that need updating.

It's also possible support doesn't know or use this and that it's just a fun thing we have here in the OSS repo. @matoval do you have any historical insight here?

matoval (commented):
lgtm

matoval (approved):

PR #1564, opened 2026-08-26
Add work adopt command to track a remote work unit

fosterseth:
## Summary

- Adds `work adopt` to create a local work unit that attaches to an already-running unit on a remote node, skipping the submit step
- Adds `--startpos` to the `work results` CLI for resuming at a byte offset
- Idempotent: second call reuses the existing local unit and monitoring goroutines

This enables node "bar" to pick up streaming results from a job that node "foo" originally submitted to node "baz".

Two approaches were considered:

**`work adopt --node baz <unitid>`**
- Pro: Makes it clear that local state is being created
- Con: New command for something conceptually similar to getting results

`work results <unitid> --node baz`
- Pro: Consistent with `work submit --node`, which also allocates local state and monitors remotely
- Con: Side effect (local unit creation) hidden behind a flag

Current implementation uses `work adopt --node`.

## Example

```
┌───────┐                      
│  foo  ┼─────────────────┐    
└───────┘                 │    
                      ┌───▼───┐
                      │  baz  │
┌───────┐             └───▲───┘
│  bar  ┼─────────────────┘    
└───────┘                      
```
1. foo submits work to baz
```
$ printf "1\n2\n3\n" | receptorctlfoo work submit idk --node baz --payload -
Result: Job Started
Unit ID: fooOdeZdKuE
```
3. bar streams work from baz using the same unit id from the submit step
```
$ printf "1\n2\n3\n" | receptorctlbar work adopt --node baz fooOdeZdKuE -f
1
2
3
```
## Test plan

- [x] `receptorctl work adopt --node baz <unitid>` creates local unit, streams results (if -f)
- [x] Second call reuses existing local unit without error
- [x] `--startpos` resumes at correct byte offset
- [x] `--rm` releases local unit after completion

## Summary by CodeRabbit

- **New Features**
  - Added `work adopt` to attach to an existing remote work unit.
  - Supports TLS client selection, signed submissions, follow mode, and optional release after completion.
  - Adoption is idempotent when the work unit is already present locally.
  - Added support for adopting work through the control interface.

- **Improvements**
  - `work results` now accepts a byte-offset start position.
  - Invalid result start positions default to zero instead of returning an error.

### DIFF
--- a/docs/source/receptorctl/receptorctl_work_adopt.rst
+++ b/docs/source/receptorctl/receptorctl_work_adopt.rst
@@ -1,0 +1,71 @@
+----------
+work adopt
+----------
+
+.. contents::
+   :local:
+
+``receptorctl work adopt`` attaches to an already-running unit of work on a remote Receptor node, allowing a node to monitor and retrieve results from work that was submitted by another node.
+
+Command syntax: ``receptorctl --socket=<socket_path> work adopt [<<Options>>] --node <<Node ID>> <<Unit ID>>``
+
+``socket_path`` is the control socket address for the Receptor connection.
+   The default is ``unix:`` for a Unix socket.
+   Use ``tcp://`` for a TCP socket.
+   The corresponding environment variable is ``RECEPTORCTL_SOCKET``.
+
+.. code-block:: text
+
+  ss --listening --processes --unix 'src = unix:<socket_path>'
+  Netid         State          Recv-Q         Send-Q                   Local Address:Port                     Peer Address:Port        Process
+  u_str         LISTEN         0              4096                   /tmp/local.sock 38130170                            * 0            users:(("receptor",pid=3226769,fd=7))
+
+``ps -fp $(pidof receptor)``
+``lsof -p <pid>``
+
+``Node ID`` is the identifier of the Receptor node where the work unit is currently running.
+
+``Unit ID`` is the unique identifier for the work unit on the remote node. This is the unit ID that was returned when the work was originally submitted.
+
+^^^^^^^^^^^^^^^^^^
+Work adopt options
+^^^^^^^^^^^^^^^^^^
+
+You can use the following options with the ``work adopt`` command:
+
+.. list-table::
+    :header-rows: 1
+    :widths: auto
+
+    * - Option
+      - Description
+    * - ``--node <<Node ID>>``
+      - Specifies the Receptor node where the work unit is running (required).
+    * - ``-f``, ``--follow``
+      - Remains attached to the job and streams the job results to standard output (stdout).
+    * - ``--rm``
+      - Releases the local work unit after the remote work completes.
+    * - ``--signwork``
+      - Digitally signs the adopt request when connecting to the remote node.
+    * - ``--tls-client <<TEXT>>``
+      - Specifies the TLS client configuration to use when connecting to the remote node.
+
+^^^^^^^^^^^^^
+Usage example
+^^^^^^^^^^^^^
+
+The ``work adopt`` command is useful when multiple nodes need to monitor the same work unit, or when you want to monitor work that was submitted from a different node.
+
+.. code-block:: shell
+
+   # Node A submits work to Node C
+   receptorctl --socket /tmp/nodeA.sock work submit --node nodeC echojob
+   Unit ID: abc-123
+
+   # Node B can adopt the same work unit to monitor it
+   receptorctl --socket /tmp/nodeB.sock work adopt --node nodeC --follow abc-123
+
+The ``work adopt`` command is idempotent. Running it multiple times with the same unit ID will not create duplicate monitoring sessions.
+
+.. note::
+   The work unit must already exist and be running on the remote node. The ``work adopt`` command does not submit new work; it only attaches to existing work units.
--- a/pkg/workceptor/controlsvc.go
+++ b/pkg/workceptor/controlsvc.go
@@ -421,9 +421,20 @@ func (c *workceptorCommand) ControlFunc(ctx context.Context, nc controlsvc.Netce
 				tlsClient = ""
 			}
 			signWork, _ := boolFromMap(c.params, "signwork")
-			worker, err := c.w.AllocateResumedRemoteUnit(remoteNode, "remote", unitid, tlsClient, signWork)
+			worker, err := c.w.findUnit(unitid)
 			if err != nil {
-				return nil, err
+				worker, err = c.w.AllocateUnit("remote", unitid, map[string]string{})
+				if err != nil {
+					return nil, err
+				}
+				worker.UpdateFullStatus(func(status *StatusFileData) {
+					ed := status.ExtraData.(*RemoteExtraData)
+					ed.RemoteNode = remoteNode
+					ed.RemoteUnitID = unitid
+					ed.RemoteStarted = true
+					ed.TLSClient = tlsClient
+					ed.SignWork = signWork
+				})
 			}
 			worker.UpdateBasicStatus(WorkStatePending, "Resuming Remote Work", 0)
 			err = worker.Restart()
--- a/pkg/workceptor/controlsvc.go
+++ b/pkg/workceptor/controlsvc.go
@@ -73,6 +73,12 @@ func (t *workceptorCommandType) InitFromString(params string) (controlsvc.Contro
 		} else {
 			c.params["startpos"] = int64(0)
 		}
+	case "adopt":
+		if len(tokens) < 3 {
+			return nil, fmt.Errorf("work adopt requires a remote node and unit ID")
+		}
+		c.params["node"] = tokens[1]
+		c.params["unitid"] = tokens[2]
 	}
 
 	return c, nil
@@ -187,9 +193,14 @@ func (t *workceptorCommandType) InitFromJSON(config map[string]interface{}) (con
 		if err == nil {
 			c.params["signature"] = signature
 		}
-		node, err := strFromMap(config, "node")
-		if err == nil {
-			c.params["node"] = node
+	case "adopt":
+		c.params["node"], err = strFromMap(config, "node")
+		if err != nil {
+			return nil, err
+		}
+		c.params["unitid"], err = strFromMap(config, "unitid")
+		if err != nil {
+			return nil, err
 		}
 		tlsClient, err := strFromMap(config, "tlsclient")
 		if err == nil {
@@ -414,36 +425,6 @@ func (c *workceptorCommand) ControlFunc(ctx context.Context, nc controlsvc.Netce
 		if err != nil {
 			return nil, err
 		}
-		remoteNode, _ := strFromMap(c.params, "node")
-		if remoteNode != "" {
-			tlsClient, err := strFromMap(c.params, "tlsclient")
-			if err != nil {
-				tlsClient = ""
-			}
-			signWork, _ := boolFromMap(c.params, "signwork")
-			worker, err := c.w.findUnit(unitid)
-			if err != nil {
-				worker, err = c.w.AllocateUnit("remote", unitid, map[string]string{})
-				if err != nil {
-					return nil, err
-				}
-				worker.UpdateFullStatus(func(status *StatusFileData) {
-					ed := status.ExtraData.(*RemoteExtraData)
-					ed.RemoteNode = remoteNode
-					ed.RemoteUnitID = unitid
-					ed.RemoteStarted = true
-					ed.TLSClient = tlsClient
-					ed.SignWork = signWork
-				})
-			}
-			worker.UpdateBasicStatus(WorkStatePending, "Resuming Remote Work", 0)
-			err = worker.Restart()
-			if err != nil && !IsPending(err) && !strings.Contains(err.Error(), "already running") {
-				worker.UpdateBasicStatus(WorkStateFailed, fmt.Sprintf("Error resuming remote work: %s", err), 0)
-
-				return nil, err
-			}
-		}
 		startPos, err := intFromMap(c.params, "startpos")
 		if err != nil {
 			startPos = 0
@@ -473,6 +454,60 @@ func (c *workceptorCommand) ControlFunc(ctx context.Context, nc controlsvc.Netce
 		}
 
 		return nil, nil
+	case "adopt":
+		remoteNode, err := strFromMap(c.params, "node")
+		if err != nil {
+			return nil, err
+		}
+		unitid, err := strFromMap(c.params, "unitid")
+		if err != nil {
+			return nil, err
+		}
+		tlsClient, err := strFromMap(c.params, "tlsclient")
+		if err != nil {
+			tlsClient = ""
+		}
+		signWork, _ := boolFromMap(c.params, "signwork")
+
+		cfr := make(map[string]interface{})
+
+		// Try to find existing unit first - if it exists, we've already adopted it
+		worker, err := c.w.findUnit(unitid)
+		if err == nil {
+			// Unit already exists - idempotent success, don't change anything
+			cfr["unitid"] = worker.ID()
+			cfr["result"] = "Already Adopted"
+			return cfr, nil
+		}
+
+		// Create new remote unit (remoteWorkType will be updated when we monitor)
+		worker, err = c.w.AllocateRemoteUnit(remoteNode, "", unitid, tlsClient, "", signWork, map[string]string{})
+		if err != nil {
+			return nil, err
+		}
+
+		// Set RemoteUnitID and RemoteStarted for adoption
+		worker.UpdateFullStatus(func(status *StatusFileData) {
+			ed := status.ExtraData.(*RemoteExtraData)
+			ed.RemoteUnitID = unitid
+			ed.RemoteStarted = true
+		})
+
+		cfr["unitid"] = worker.ID()
+		worker.UpdateBasicStatus(WorkStatePending, "Adopting Remote Work", 0)
+		err = worker.Restart()
+		if err != nil && !IsPending(err) {
+			worker.UpdateBasicStatus(WorkStateFailed, fmt.Sprintf("Error adopting remote work: %s", err), 0)
+
+			return cfr, err
+		}
+		if IsPending(err) {
+			cfr["result"] = "Adopt Pending"
+		} else {
+			cfr["result"] = "Adopted"
+		}
+
+		return cfr, nil
 	}
 
 	return nil, fmt.Errorf("bad command")
--- a/pkg/workceptor/controlsvc.go
+++ b/pkg/workceptor/controlsvc.go
@@ -73,6 +73,12 @@ func (t *workceptorCommandType) InitFromString(params string) (controlsvc.Contro
 		} else {
 			c.params["startpos"] = int64(0)
 		}
+	case "resume":
+		if len(tokens) < 3 {
+			return nil, fmt.Errorf("work resume requires a remote node and remote unit ID")
+		}
+		c.params["node"] = tokens[1]
+		c.params["remote-unitid"] = tokens[2]
 	}
 
 	return c, nil
@@ -186,6 +192,23 @@ func (t *workceptorCommandType) InitFromJSON(config map[string]interface{}) (con
 		signature, err := strFromMap(config, "signature")
 		if err == nil {
 			c.params["signature"] = signature
+		}
+	case "resume":
+		c.params["node"], err = strFromMap(config, "node")
+		if err != nil {
+			return nil, err
+		}
+		c.params["remote-unitid"], err = strFromMap(config, "remote-unitid")
+		if err != nil {
+			return nil, err
+		}
+		tlsClient, err := strFromMap(config, "tlsclient")
+		if err == nil {
+			c.params["tlsclient"] = tlsClient
+		}
+		signWork, err := boolFromMap(config, "signwork")
+		if err == nil {
+			c.params["signwork"] = signWork
 		}
 	}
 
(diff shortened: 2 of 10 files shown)


AlanCoding:
```
work resume --node baz <unitid>
```

I need to say out loud the parts left assumed.

Currently, the way things are, there is a work unit id seen by the node that submits the work. The node that receives the work _receives a different work unit id_. As currently, there is a whole thing of receptor providing the remote work unit id.

This introduces a completely new actor into the mix. Now, we have the original control node, the execution node, and the new control node adopting the work.

Historically, the work unit id has always been _strictly locally unique_. In basically both your proposals, there is no new language for the "new" controller node to assign a new work unit id, specific to itself. This matters, because there was no guarantee that the original work_unit_id will be unique _in another node_. As I read the current proposal, it introduces the possibility of collisions. The ID is intentionally made to be low-entropy because it's only a local matter, can be managed by a singleton. That fails down with the addition of this feature, and no solution is obviously proposed.

Possible solutions:
 - what I was proposing - we just make the work_unit_id globally unique. Receptor _can not do this_. It would have to communicate with the rest of the mesh with is both unreliable and an anti-pattern. However, the _client_ can absolutely do this. Specifically, AWX already has unique ids for jobs, and doesn't need something else to make one for it.
 - most consistent with your solution here, staying in the receptor domain, assign a new work_unit_id upon adoption.

The second solution does influence the API design here, suggesting that `resume` by itself isn't truly clean. What would be _consistent_ with the rest of the design, to me, would be:
 - `receptorctl adopt --node <unit_id>` --> returns newly minted local unit_id
 - `receptorctl resume ...` ... like normal

For our _intended use_ I agree the collision risk is low, but I don't think this matters. The risk is extremely painful at the design level, and I think people, even AI agents will balk and get distracted by this if the original contract design does not give it proper treatment.

I even hesitate at my first possible solution here, because while it would "work", the receptor-side implementation would be logically flawed. So I would say the 2nd solution is arguably more purist. Another option is to do _both_, which actually requires another API constraint, that we could do `receptorctl adopt --node <unit_id> --new-unit-id <unit_id>`, so AWX specifically tells receptor, "adopt this, when you adopt it, give it the same unit id".

fosterseth:
@AlanCoding current receptor has the ability to take a WorkUnitID as a parameter when submitting work

```
$ printf "1\n2\n3\n" | receptorctlfoo work submit idk --node baz --payload - --param "workUnitID=peonworker"
Result: Job Started
Unit ID: peonworker

$ printf "1\n2\n3\n" | receptorctlfoo work submit idk --node baz --payload - --param "workUnitID=peonworker"
ERROR: Failed to start work unit: workunit ID peonworker is already in use, cannot use the same workunit ID more than once
```
So controller just needs to take advantage of that and supply a globally unique work unit for every job submission

Current receptor already is prone to collisions in that when it creates a local work unit id, it attempts to use that same unit id on the remote node. The job can fail entirely due to naming conflicts. You can argue we're now involving a 3rd node in to the equation which raises the chances for name conflicts. I think you hinted at that.

We probably need a solution for the current naming collision as well as the naming collision when doing the resume functionality (a 3rd receptor node adopting an existing work unit)

Here is what I think could happen:
- The default behavior of receptor should be to use a locally unique work ID by default. Forcing the unit IDs to match is not the default behavior, to prevent naming collisions.
- There is an option to supply a work unit ID. It then becomes the responsibility of the work submitter (i.e. controller) to guarantee uniqueness across the mesh.

We'd need to change current receptor to rollback enforcing the remote node to have the same ID as the local one, but still keep it as an option if supplied an ID.

matoval (commented):
Tested this against a 3-node mesh running the full ansible-runner transmit/worker/process pipeline. Ran 13 edge cases, 12 passed. Core flow works well — adopt running job, adopt completed job, kill submitter + adopt from another node, idempotent adopt, `--follow`, `--rm`, `--startpos` all good.

**Bug: adopt from non-existent node hangs forever**

`work adopt --node nonexistent-node <unitid>` returns `Result: Adopted` (exit 0) but the unit stays Pending indefinitely. No timeout, no failure transition.

The adopt case in `controlsvc.go` (~line 484) calls `AllocateRemoteUnit` then `Restart()`, which goes through `startOrRestart` → `monitorRemoteUnit` → `GetConnection` (`remote_work.go:99`). `GetConnection` is an infinite retry loop with backoff — it never gives up for a node that doesn't exist in the routing table.

I know `work submit` to a nonexistent node has the same behavior (`TestWorkSubmitAndReleaseToNonexistentNode` asserts Pending), but for adopt the caller expects to attach to something already running, so silent success + permanent hang is worse. At minimum, validate the node exists in `nc.Status().RoutingTable` before calling `AllocateRemoteUnit`. Contrast with non-existent unit ID on a valid node — that correctly transitions to Failed.

**Tests**

No tests included. The existing functional test framework in `tests/functional/mesh/` already supports everything needed — `workTestMesh` sets up a 3-node mesh, `ReceptorControl` has all the assert helpers. Just needs a `WorkAdoptJSON` method on `ReceptorControl` (same pattern as `WorkSubmitJSON`).

Unit tests (`controlsvc_test.go`):
- `InitFromString` parsing: `"adopt nodeA unit123"` succeeds, `"adopt nodeA"` missing unitid errors, `"adopt"` missing both errors
- `InitFromJSON` parsing: valid node+unitid, missing node, missing unitid, optional tlsclient/signwork

Integration tests (`work_test.go`):
- Adopt a running job, both nodes see Succeeded
- Adopt a completed job, adopter gets results
- Adopt same unit twice → "Already Adopted"
- Adopt non-existent unit ID → Failed
- Adopt from non-existent node → should fail, not hang
- Adopt after submitter restart

**Minor**

- `RemoteWorkType` is empty string on adopted units (line 484 passes `""` to `AllocateRemoteUnit`). Doesn't break anything since the monitor uses unit ID, but worth a comment.
- The `results` command lost its post-stream status check (removed `work status` + exit-on-failure block in `cli.py`). Was that intentional? `work results` on a failed unit now exits 0 instead of 1.

fosterseth:
@matoval `adopt from non-existent node hangs forever`
this is how work submit works today, so at least adopt is consistent with that behavior. If the remote node is not accessible when you run `work adopt` the command can still complete without error, and once the node comes online it will start streaming automatically.

matoval:
> @matoval `adopt from non-existent node hangs forever` this is how work submit works today, so at least adopt is consistent with that behavior. If the remote node is not accessible when you run `work adopt` the command can still complete without error, and once the node comes online it will start streaming automatically.

You're correct, this is normal Receptor behavior and not exclusive to the new adopt feature. You can discard that part of my review.

pkg/workceptor/controlsvc.go
```diff
@@ -181,12 +187,29 @@ func (t *workceptorCommandType) InitFromJSON(config map[string]interface{}) (con
...
+			return nil, err
+		}
+		tlsClient, err := strFromMap(config, "tlsclient")
+		if err == nil {
+			c.params["tlsclient"] = tlsClient
+		}
+		signWork, err := boolFromMap(config, "signwork")
+		if err == nil {
+			c.params["signwork"] = signWork
+		}
```
PabloHiro, line 216: Does it make sense to support `signwork` for `adopt`?

receptorctl/receptorctl/cli.py
```diff
@@ -514,22 +514,67 @@ def submit(
...
     for text in iter(partial(resultsfile.readline, 256), b""):
         sys.stdout.buffer.write(text)
         sys.stdout.buffer.flush()
-    rc = get_rc(ctx)
-    status = rc.simple_command(f"work status {unit_id}")
-    state = status.pop("State", 0)
-    if state == 3:  # Failed
-        detail = status.pop("Detail", "Unknown")
-        print_error(f"Remote unit failed: {detail}\n")
-        sys.exit(1)
```
PabloHiro, line 532: Why is this block getting removed?

pkg/workceptor/controlsvc.go
```diff
@@ -73,6 +73,12 @@ func (t *workceptorCommandType) InitFromString(params string) (controlsvc.Contro
 		} else {
 			c.params["startpos"] = int64(0)
 		}
+	case "adopt":
+		if len(tokens) < 3 {
+			return nil, fmt.Errorf("work adopt requires a remote node and unit ID")
+		}
+		c.params["node"] = tokens[1]
+		c.params["unitid"] = tokens[2]
```
PabloHiro, line 81:
```suggestion
		c.params["unitid"] = tokens[2]
		if len(tokens) > 3 {
			c.params["params"] = strings.Join(tokens[3:], " ")
		}
```

"submit" has this. Wouldn't we be dropping the rest of the params if we don't add this?

tests/functional/mesh/receptorcontrol.go
```diff
@@ -300,6 +300,26 @@ func (r *ReceptorControl) WorkRelease(unitID string) (map[string]interface{}, er
...
+// WorkAdoptJSON adopts a remote work unit via JSON command.
+func (r *ReceptorControl) WorkAdoptJSON(command string) (map[string]interface{}, error) {
+	_, err := r.WriteStr(fmt.Sprintf("%s\n", command))
+	if err != nil {
+		return nil, err
+	}
+
+	return r.ReadAndParseJSON()
+}
+
```
PabloHiro, line 322: My LLM says this is dead code, can you double check and remove if true?

pkg/workceptor/remote_work.go
```diff
@@ -371,6 +372,15 @@ func (rw *remoteUnit) monitorRemoteStatus(mw *utils.JobContext, forRelease bool)
...
 		}
 		rw.UpdateBasicStatus(si.State, si.Detail, si.StdoutSize)
+		if remoteWorkType == "" {
+			// work adopt starts this with "", so do a full update if empty
+			rw.UpdateFullStatus(func(status *StatusFileData) {
+				ed := status.ExtraData.(*RemoteExtraData)
+				if ed.RemoteWorkType == "" && si.WorkType != "" {
+					ed.RemoteWorkType = si.WorkType
+				}
+			})
```
PabloHiro, line 382:
```suggestion
		if remoteWorkType == "" {
			// work adopt starts this with "", so do a full update if empty
			rw.UpdateFullStatus(func(status *StatusFileData) {
				ed := status.ExtraData.(*RemoteExtraData)
				if ed.RemoteWorkType == "" && si.WorkType != "" {
					ed.RemoteWorkType = si.WorkType
				}
			})
			if si.WorkType != "" {
				remoteWorkType = si.WorkType
			}
```

My LLM says that we do not update the value of `remoteWorkType` inside the loop, so we would be constantly calling `UpdateFullStatus` with the same info, when we only need to do it once.

pkg/workceptor/controlsvc.go
```diff
@@ -431,6 +454,61 @@ func (c *workceptorCommand) ControlFunc(ctx context.Context, nc controlsvc.Netce
...
+			ed.RemoteStarted = true
+		})
+
+		cfr["unitid"] = worker.ID()
+		worker.UpdateBasicStatus(WorkStatePending, "Adopting Remote Work", 0)
+		err = worker.Restart()
+		if err != nil && !IsPending(err) {
+			worker.UpdateBasicStatus(WorkStateFailed, fmt.Sprintf("Error adopting remote work: %s", err), 0)
+
+			return cfr, err
```
PabloHiro, line 520: The LLM flags this as problematic and I think it is correct.

This is read in this file:
https://github.com/ansible/receptor/blob/f9be10658742c64ce639518e0112cdf416f46159/pkg/controlsvc/controlsvc.go#L384-L392

So the `cfr` information is discarded. 

Next time we enter the loop, we have created a unit id directory in the node attempting to adopt the worker (via `worker.UpdateBasicStatus`), so the check some lines above on line 475:
```
worker, err := c.w.findUnit(unitid)
```

Flags that this is `"Already Adopted"` when in fact it failed to restart the worker.

PabloHiro (changes requested):
@fosterseth I tested this locally with this topology:

- 2 control nodes (A and B), 1 hop node, 2 execution nodes (A and B)

Happy path works well, but I found a bug with the `rm` flag. When adopting with `rm` the original work unit gets cancelled even if it was not finished. Reproducer:
```
receptorctl --socket /tmp/control-a.sock work submit --node execution-a bash -l 'while true; do echo "OK-$(date)"; sleep 5; done'
receptorctl --socket /tmp/control-a.sock work results <work_unit_id>
# In another shell
receptorctl --socket /tmp/control-b.sock work adopt --node execution-a <work_unit_id> --rm
receptorctl --socket /tmp/control-b.sock work list --unit_id <work_unit_id>
```

Not sure what the root cause is.

I have also left some comments from me and an LLM assisted review.

pkg/workceptor/controlsvc.go
fosterseth: you're right, sign work should only be needed when submitting work. I'll remove it

fosterseth: never mind signing work in necessary for work results streaming

https://github.com/ansible/receptor/blob/f9be10658742c64ce639518e0112cdf416f46159/pkg/workceptor/remote_work.go#L454

pkg/workceptor/remote_work.go
fosterseth: good catch, fixed it

tests/functional/mesh/receptorcontrol.go
fosterseth: removed

pkg/workceptor/controlsvc.go
fosterseth: okay good call. I think for this we can just remove the idempotency. If you run adopt again, it will just error out

```
sbf:~/receptor$ printf "1\n2\n3\n" | receptorctlfoo work adopt --node bar bargt0DMs5V
ERROR: workunit ID bargt0DMs5V is already in use, cannot use the same workunit ID more than once
```

receptorctl/receptorctl/cli.py
fosterseth: did not intend to remove this, will put back

fosterseth:
@PabloHiro the `--rm` was also a bug in "work submit"
`receptorctlbar work submit idk -p - --rm`

this would release (cancel and delete) immediately after starting work (does not wait for work to finish). I think the intention was to pair --rm with --follow, so I made a change to enforce that. This only affects cli behavior (which awx does not use, it uses ReceptorControl() directly)

I applied the same behavior to adopt

pkg/workceptor/controlsvc.go
```diff
@@ -404,7 +434,7 @@ func (c *workceptorCommand) ControlFunc(ctx context.Context, nc controlsvc.Netce
 		}
 		startPos, err := intFromMap(c.params, "startpos")
```
matoval, line 441: Confirming CodeRabbit's Aug 26 comment on this line is still open. `intFromMap` parses `-5` successfully (it's a valid integer), so the `err != nil` fallback never triggers — the negative value passes straight through to `GetResults`'s `Seek` call, which rejects it server-side. That error gets logged (`Seek error processing stdout: seek .../stdout: invalid argument`) but never returned to the caller, so `work results <id> --startpos -5` silently returns exit 0 with empty output — indistinguishable from "this job produced no output." I confirmed this live against a real job with actual stdout.

**Suggested fix:** clamp to zero the same way the parse-failure case already does — `if startPos < 0 { startPos = 0 }` right after the existing fallback — or return an explicit validation error instead of swallowing the `Seek` failure.

matoval (changes requested):
Manually tested this against a live 3-node mesh across a few commits now (`927bdcfb` → `3f63a203`). The `--rm` cancel-on-a-still-running-unit bug and the dropped `work results` failure exit-code are both fixed and verified live.

One item still open: negative `--startpos` silently swallows a real server-side error instead of surfacing it — see inline comment for the repro and root cause. CodeRabbit flagged this back on Aug 26; it's the last thing from that original pass that hasn't been addressed.

Test coverage still has some gaps specific to this PR's changes:
- `TestWorkAdopt`'s second-call assertion needs updating — it still expects `"Already Adopted"`, but the latest commit made adopt non-idempotent, so a second call should now error.
- No test adopts an already-completed job — only the running-job path is covered, but the PR description claims completed-job adoption works too.
- No unit tests for `InitFromString`/`InitFromJSON` parsing of the `adopt` case in `controlsvc_test.go` (missing node, missing unitid, malformed params) — currently only reachable through the full mesh harness.
- No test confirms `--rm` without `--follow` now errors, for either `submit` or `adopt`.
- No test covers signature verification for `adopt` — `1cfc8a5d` fixed a real bug here (`signwork` not persisting, verification skipped on non-unix connections), but nothing confirms the fix holds. `TestSignedWorkVerification` covers `submit`; nothing equivalent exists for `adopt`.
- No test covers adopting a nonexistent unit ID on a reachable node (should transition to `Failed`) — distinct from the nonexistent-*node* case, which is pre-existing/out of scope here.

docs/source/receptorctl/receptorctl_work_adopt.rst
```diff
@@ -0,0 +1,71 @@
...
+.. code-block:: shell
+
+   # Node A submits work to Node C
+   receptorctl --socket /tmp/nodeA.sock work submit --node nodeC echojob
+   Unit ID: abc-123
+
+   # Node B can adopt the same work unit to monitor it
+   receptorctl --socket /tmp/nodeB.sock work adopt --node nodeC --follow abc-123
+
+The ``work adopt`` command is idempotent. Running it multiple times with the same unit ID will not create duplicate monitoring sessions.
```
PabloHiro: We need to update this. No idempotency anymore.

pkg/workceptor/controlsvc.go
PabloHiro: Working, thanks!

pkg/workceptor/controlsvc.go
PabloHiro: I could reproduce this one as well 👍

docs/source/receptorctl/receptorctl_work_adopt.rst
```diff
@@ -0,0 +1,71 @@
+----------
+work adopt
+----------
+
+.. contents::
+   :local:
+
+``receptorctl work adopt`` attaches to an already-running unit of work on a remote Receptor node, allowing a node to monitor and retrieve results from work that was submitted by another node.
```
PabloHiro: I would update the documentation with this. I think it makes it clearer. What do you think?

```suggestion
``receptorctl work adopt`` attaches to an already-running unit of work on a remote Receptor node, allowing a node to co-own that unit of work. It can monitor and retrieve results from work that was submitted to another node.
```

docs/source/receptorctl/receptorctl_work_adopt.rst
fosterseth: updated

pkg/workceptor/remote_work.go
```diff
@@ -370,7 +373,28 @@ func (rw *remoteUnit) monitorRemoteStatus(mw *utils.JobContext, forRelease bool)
...
+			remoteWorkType = si.WorkType
```
matoval: If `UpdateFullStatus` fails, this still marks the local cache as populated. The next poll then skips the persistence attempt, so a transient status-file write failure can leave `RemoteWorkType` absent from the saved status until restart. Please only update `remoteWorkType` after a successful write (or confirm the persisted value) and add a retry/failure test.

tests/functional/mesh/work_test.go
```diff
@@ -816,3 +816,80 @@ func TestSignedWorkVerification(t *testing.T) {
 		t.Fatalf("Did not see the expected error. Wanted %s, got: %s", expected, actual)
 	}
 }
+
+func TestWorkAdopt(t *testing.T) {
```
matoval: This new test is also reported by Sonar for cognitive complexity (32 vs. 15). Please split setup/submit, adoption, and result checks into focused helpers. The feature coverage requested earlier is still incomplete: add cases for adopting an already-completed unit, adopting a nonexistent unit ID on a reachable node, and signed adoption/verification.

pkg/workceptor/controlsvc_test.go
```diff
@@ -357,3 +397,353 @@ func Test_boolFromMap(t *testing.T) {
 		})
 	}
 }
+
+func Test_workceptorCommandTypeInitFromString_AdoptParams(t *testing.T) {
```
matoval: Sonar flags this new test for cognitive complexity (41 vs. 15). Please factor the repeated assertions into helpers to clear the gate, and add failure cases for missing required fields in the string-form adopt command; these cases currently exercise only valid input.

pkg/workceptor/controlsvc_test.go
```diff
@@ -357,3 +397,353 @@ func Test_boolFromMap(t *testing.T) {
...
+					t.Errorf("startpos = %v, want %v", startpos, tt.wantStartpos)
+				}
+			} else {
+				t.Errorf("startpos not found in params or wrong type")
+			}
+		})
+	}
+}
+
+func Test_workceptorCommandTypeInitFromJSON_AdoptParams(t *testing.T) {
```
matoval: Sonar flags this new test for cognitive complexity (62 vs. 15). Please factor the assertions into helpers and cover missing `node`/`unitid` and malformed or wrongly typed optional values; the current table covers only valid JSON configurations.

receptorctl/receptorctl/cli.py
```diff
@@ -510,16 +511,71 @@ def submit(
...
+@click.option("--signwork", help="Digitally sign remote work submissions", is_flag=True)
+@click.option(
+    "--follow",
+    "-f",
+    help="Remain attached to the job and print its results to stdout",
+    is_flag=True,
+)
+@click.option("--rm", help="Release unit after completion", is_flag=True)
+def adopt(ctx, node, unit_id, tlsclient, signwork, follow, rm):
+    if rm and not follow:
```
matoval: Please add CLI regression tests that `work adopt --rm` without `--follow` exits with an error before issuing the control request. The analogous guard in `submit` needs the same coverage.

receptorctl/receptorctl/socket_interface.py
```diff
@@ -243,6 +243,31 @@ def submit_work(
...
 
+    def adopt_work(
+        self,
+        node,
+        unit_id,
+        tlsclient=None,
+        signwork=False,
+    ):
+        self.connect()
+        commandMap = {
```
matoval: Please rename the new `commandMap` and `commandJson` locals to `command_map` and `command_json`; Sonar reports both naming violations on this added code.

matoval (changes requested):
Re-review of PR #1564 at `7c490ab`. Requesting changes for the remaining PR-introduced issues below.

### Remaining requested tests
Please add regression tests for adopting already-completed work, adopt signature verification, a nonexistent unit ID on a reachable node, missing/invalid adopt parameters, and rejection of `--rm` without `--follow`. The inline comments identify the relevant test locations.

### SonarCloud coverage and maintainability
Sonar reports 36.7% coverage on new code, below the 90% gate. Please add unit tests in the measured packages to exercise the new `ControlFunc` adoption branch and other uncovered new production lines until Sonar reports at least 90% new-code coverage. Do not rely on `TestWorkAdopt` for this gate: `make coverage` excludes `/tests/`, so that functional test is not included in the coverage profile. My local package unit-test profile showed 0/35 statements covered in the new adoption branch.

Separately, the five new Sonar maintainability issues need to be cleared: three cognitive-complexity findings in the tests and two Python naming findings. The inline comments identify them.

### Additional finding
`remote_work.go` caches `remoteWorkType` even if persisting it fails. Please only update the in-memory cache after a successful status write, and cover the retry behavior.

PR #1568, opened 2026-09-09
fix(security): validate python worker CLI arguments against injection and path traversal

kristinamarkina:
## Summary

- Add `validate_plugin_directive()`: enforces a strict regex allowlist (`^[A-Za-z0-9_]+:[A-Za-z0-9_]+$`) on the `namespace:function` argument, blocking shell metacharacters, path separators, newlines, null bytes, and other injection vectors before the string reaches `pkg_resources`.
- Add `validate_unitdir()`: multi-layer path-traversal defence — requires an absolute path, rejects any `..` component, and detects symlink escapes by comparing `Path.resolve()` against the lexically-normalised form.
- Wire both validators into `WorkPluginRunner.__init__` before any filesystem access; update all three `os.path.join(unitdir, …)` calls to use the validated `self.unitdir`.

## Test plan

- [ ] `python -m pytest receptor-python-worker/tests/ -v` — 22 tests, all passing
- [ ] Covers valid `namespace:function` inputs (simple, underscores, mixed case, digits)
- [ ] Covers invalid directives: missing colon, shell metacharacters (`;`, `&`), path separators, newline/null injection, double colon, spaces
- [ ] Covers valid `unitdir` (real absolute directory)
- [ ] Covers invalid `unitdir`: relative path, `./` path, `..` traversal, nonexistent path, file instead of directory, symlink escaping to a different location

## Summary by CodeRabbit

* **Bug Fixes**
  * Plugin directives now accept only safe `namespace:function` formats.
  * Unit directories must be absolute paths resolving to existing directories without traversal or symlink escapes.
  * Invalid plugin and unit-directory configuration is rejected during setup.
  * Plugin discovery is more reliable across supported Python environments.

* **Tests**
  * Expanded coverage for validation, plugin execution, payload handling, status output, configuration errors, and signal handling.
  * Added automated Python worker coverage reporting for quality checks.

### DIFF
--- a/receptor-python-worker/receptor_python_worker/work.py
+++ b/receptor-python-worker/receptor_python_worker/work.py
@@ -1,5 +1,6 @@
 import sys
 import os
+import re
 import threading
 import signal
 import queue
@@ -11,6 +12,58 @@ from .plugin_utils import BUFFER_PAYLOAD, BYTES_PAYLOAD, FILE_PAYLOAD
 # Allow existing worker plugins to "import receptor" and get our version of plugin_utils
 sys.modules['receptor'] = sys.modules[__package__+'.plugin_utils']
 
+# Allowlist pattern for plugin directives: "namespace:function" where both parts are
+# restricted to alphanumerics and underscores. This blocks shell metacharacters,
+# path separators, and other injection vectors before the string reaches pkg_resources.
+_PLUGIN_DIRECTIVE_RE = re.compile(r'^[A-Za-z0-9_]+:[A-Za-z0-9_]+$')
+
+
+def validate_plugin_directive(directive):
+    """Validate that a plugin directive matches 'namespace:function' allowlist pattern.
+
+    Raises ValueError if the directive contains any character outside [A-Za-z0-9_:]
+    or does not follow the expected two-part colon-delimited format.
+    """
+    if not _PLUGIN_DIRECTIVE_RE.match(directive):
+        raise ValueError(
+            "Plugin directive must be 'namespace:function' using only alphanumerics and underscores"
+        )
+
+
+def validate_unitdir(unitdir):
+    """Resolve and validate a unit directory path against path-traversal attacks.
+
+    Strategy:
+      1. Require an absolute path so relative paths are rejected immediately.
+      2. Reject any '..' components in the path parts before touching the filesystem.
+      3. Call Path.resolve() (follows symlinks) and compare to the lexically-normalised
+         form.  A mismatch means a symlink redirects the path to a different location —
+         reject it so callers cannot use a planted symlink to escape the intended tree.
+      4. Require the resolved path to be an existing directory.
+
+    Returns the resolved Path on success; raises ValueError on any violation.
+    """
+    p = Path(unitdir)
+    if not p.is_absolute():
+        raise ValueError(f"unitdir must be an absolute path, got: {unitdir!r}")
+
+    # Catch '..' traversal in the path string itself before any I/O.
+    if ".." in p.parts:
+        raise ValueError(f"unitdir must not contain '..' components: {unitdir!r}")
+
+    resolved = p.resolve()
+    normalised = Path(os.path.normpath(str(p)))
+    # If resolve() changed the path, a symlink redirects to a different location.
+    if resolved != normalised:
+        raise ValueError(
+            f"unitdir resolves outside its stated location (symlink or traversal): {unitdir!r}"
+        )
+
+    if not resolved.is_dir():
+        raise ValueError(f"unitdir does not exist or is not a directory: {unitdir!r}")
+
+    return resolved
+
 # WorkState constants
 WorkStatePending = 0
 WorkStateRunning = 1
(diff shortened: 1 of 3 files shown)


PabloHiro (dismissed):
LGTM

`receptor-python-worker` is something that was created for testing, I wonder if it even makes sense to have it scanned by sonarqube.

@kristinamarkina, @matoval thoughts on adding the entire directory to the `sonar.exclusions` in sonar-project.properties?

receptor-python-worker/receptor_python_worker/work.py
```diff
@@ -11,6 +12,58 @@
 # Allow existing worker plugins to "import receptor" and get our version of plugin_utils
 sys.modules['receptor'] = sys.modules[__package__+'.plugin_utils']
 
+# Allowlist pattern for plugin directives: "namespace:function" where both parts are
+# restricted to alphanumerics and underscores. This blocks shell metacharacters,
+# path separators, and other injection vectors before the string reaches pkg_resources.
+_PLUGIN_DIRECTIVE_RE = re.compile(r'^[A-Za-z0-9_]+:[A-Za-z0-9_]+$')
```
dleehr, line 18: Any thoughts on the SonarCloud suggestion for the `\w` character class?

dleehr:
> LGTM
> 
> `receptor-python-worker` is something that was created for testing, I wonder if it even makes sense to have it scanned by sonarqube.

Is that accurate? It seems like it's available/supported by workceptor. It's possible it's not a widely used interface but I don't see any evidence that it's only used for testing. So I'd push back on excluding this directory from sonar - unless we can delete `receptor-python-worker` entirely, but that's not really the scope here.

dleehr (commented):
The SonarCloud check is failing with 0% coverage on the new code. But the PR adds 500 lines of tests for 60 lines of new code, which doesn't sound like 0% to me. We should be actively reviewing and addressing the Sonar checks on our PRs.

@kristinamarkina your PR description calls out that this is well-covered. Can you investigate this coverage reporting discrepancy and see how we can get an accurate report for these changes?

PabloHiro:
> unless we can delete `receptor-python-worker` entirely, but that's not really the scope here.

@dleehr I am on the we can delete the directory boat tbh. I think it is a leftover from 6 years ago when this project was moved from python to GO:
https://github.com/project-receptor/python-receptor

Notice how the `plugin_utils.py` there is an exact copy of the file in this repo.

kristinamarkina:
CI currently has no job configured to run pytest-cov or export coverage.xml for receptor-python-worker.

Current CI setup: .github/workflows/coverage_reporting.yml runs a nox coverage session exclusively for receptorctl/ (receptorctl_coverage.xml).

Why receptor-python-worker shows 0%: Because no coverage artifact is produced or passed via -Dsonar.python.coverage.reportPaths for receptor-python-worker, SonarCloud assumes 0% coverage by default.

Fix: I am adding a noxfile.py to receptor-python-worker and updating .github/workflows/coverage_reporting.yml and sonar_checks.yml to include python_worker_coverage.xml alongside receptorctl.

receptor-python-worker/receptor_python_worker/work.py
kristinamarkina: In Python 3, \w matches ASCII + Unicode word characters (e.g., non-English characters) unless re.ASCII is explicitly specified.

Retaining [A-Za-z0-9_] (or re.compile(r'^\w+:\w+$', re.ASCII)) is safer for a strict security allowlist to prevent unexpected Unicode character bypasses.

kristinamarkina:
How we got from 0% to 100% new-code coverage on SonarCloud

- Tests existed and passed locally (91% overall), but SonarCloud reported 0%. The gap wasn't in the tests — it was in how the coverage data reached SonarCloud.
- workflow_run always uses the default-branch workflow file, not the PR branch. Every change we made to sonar_checks.yml (adding the new coverage report path) was silently ignored by PR scans. The workaround: merge the python-worker coverage.xml into the existing receptorctl_coverage.xml that the default-branch Sonar workflow already consumed.
- The merge had a path-resolution bug. The merged file kept receptorctl's <source> tag, so SonarCloud resolved python-worker filenames like work.py against the receptorctl directory — a path that doesn't exist. Fix: run pytest --cov from inside receptor-python-worker/ (matching receptorctl's own pattern), producing <source>/abs/path/receptor_python_worker</source> with short bare filenames, then copy that <source> entry into the merged XML alongside receptorctl's.
- nox changes CWD to the noxfile's directory when invoked with -f. This caused the XML to land in a nested receptor-python-worker/receptor-python-worker/ path that the artifact upload couldn't find. Replaced the nox step with a plain shell script (run_coverage.sh) that explicitly cds to the correct directory.
- Additional violations cleared: committed uv.lock (fixes text:S8565 — unpredictable deps), added run_coverage.sh to sonar.exclusions (fixes shell:S8541 — pip install in a CI script can't use --only-binary with editable installs), and the \w regex simplification cleared the python:S6353 code smells.

receptor-python-worker/receptor_python_worker/work.py
dleehr: I'm confused. It looks like you did change the regex to use `\w` but the comment reads like you were going to retain the A-Z. Did you have any thoughts?

Also, there's a code comment that still references the A-Z format that's now inconsistent. Can we clean this up?

PR #1573, opened 2026-09-16
refactor(backends): remove context.Context from WebsocketSession (AAP-92405)

kristinamarkina:
Resolves SonarCloud MAJOR smell at pkg/backends/websockets.go:281 by removing the context.Context struct field from WebsocketSession.

WebsocketSession.context was only ever read via .Done() inside recvChannelizer and never required .Err(), .Deadline(), or .Value(). Storing context.Context directly in a struct field is an anti-pattern that can lead to stale contexts or memory leaks.

Changes Included
pkg/backends/websockets.go:

Replaced the context context.Context field on WebsocketSession with done <-chan struct{}.

Updated constructor/initialization parameters to capture ctx.Done() directly at construction time.

Refactored recvChannelizer to read directly from ns.done instead of ns.context.Done().

## Summary by CodeRabbit

* **Refactor**
  * Updated internal WebSocket session lifecycle signaling.
  * No user-visible behavior changes.


PabloHiro (changes requested):
I have been reading these two pages:
- https://go.dev/blog/context-and-structs
- https://github.com/golang/go/issues/22602

Instead of replacing context with a subset of it (the done channel). I think the spirit of this recommendation is to avoid storing context state in a struct in the first place.

Example:
```
func newWebsocketSession(ctx context.Context, conn Conner, closeChan chan struct{}) *WebsocketSession {
    ws := &WebsocketSession{ ... }
    go ws.recvChannelizer(ctx.Done())
    return ws
}
...
func (ns *WebsocketSession) recvChannelizer(done <-chan struct{}) {
    for {
        _, data, err := ns.conn.ReadMessage()
        select {
        case <-done:
            return
        ...
        }
    }
}
```

And then remove `done` (`context`) from the struct parameters. What do you think?

PR #1575, opened 2026-09-22
fix(ci): pin go-junit-report to immutable commit SHA (AAP-93897)

kristinamarkina:
## Summary
- Pins `go install github.com/jstemmer/go-junit-report/v2` to commit SHA `460a2d67ceb019d5e4c86ed30ba6c8894ac094c8` (tag `v2.1.0`)
- Resolves SonarCloud issue `AZ3VDAUGa1u4siwYPCz9`: mutable version tag in CI workflow exposes supply chain attack surface

## Test plan
- [ ] SonarCloud issue `AZ3VDAUGa1u4siwYPCz9` no longer flagged after merge
- [ ] CI workflow (`Generate junit test report`) passes without regression

Jira: AAP-93897

## Summary by CodeRabbit

* **Chores**
  * Pinned the test-reporting tool to a specific revision for more consistent and reproducible test report generation.


matoval (approved):
Verified the pinned SHA is correct — `460a2d67ceb019d5e4c86ed30ba6c8894ac094c8` is exactly what `refs/tags/v2.1.0` resolves to (confirmed via the GitHub API). Also confirmed `go install .../v2@460a2d67...` builds successfully and produces a binary with an identical SHA256 checksum to installing via the underlying commit hash directly — so the pin is functionally sound and immutable against tag mutation.

Approved.

melissalkelly (approved):
LGTM

